PR #3, opened 2022-01-20
Update run_setup.py

Lance0724:
U can use STM32CUBEIDE's PATH Variables(like: WORKSPACE_LOC, PROJECT_LOC ....) in the .project file
After first replace you don't really need a replace anymore

### DIFF
--- a/run_setup.py
+++ b/run_setup.py
@@ -42,8 +42,13 @@ def call_replace(filename):
     F.write(content)
     F.close()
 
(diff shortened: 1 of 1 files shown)


Lance0724:
solved the hard code in .project file

olliw42:
many thx, sir
I was thinking that such variables should exist but could not identify them, very useful
I will merge this PR to get your attribution right and then separate it out into a tool for maintenance, but I fist in fact have to remind me how to handle others PR LOL
thx again

PR #5, opened 2022-02-16
move clear event and restart the rtc

Imen-D:
Hi Olliw, this is me jinchuuriki on rc groups
As we enter the rx function, I forgot to re start the rtc timer, so interrupt based on rtc timer (time out) can't fire anymore.

### DIFF
--- a/mLRS/Common/sx-drivers/sx126x.cpp
+++ b/mLRS/Common/sx-drivers/sx126x.cpp
@@ -103,6 +103,7 @@ uint8_t out_buf[SX126X_SPI_BUF_SIZE];
     SpiTransfer(out_buf, data, len);
     SpiDeselect();
     // no delay according to semtech driver
+    ClearRxEvent();
 }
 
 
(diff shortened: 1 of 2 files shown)


olliw42:
Hey, great you made it to here !!!

I hear what you are saying, but from scratch it doesn't look like the correct approach to me. Shouldn't this be done in the ISR itself?

I mean, we may receive a frame from someone else, so the ISR would be triggered because of a RxDone, but since the syncword doesn't match the main loop would never be passed, and the timer never be restarted, but it should.

I thus would think that it should be called in the ISR, and the simplest approach could be to just call it within ClearIrqStatus().

If you agree with this, and can confirm it does work, I would suggest this change.

olliw42:
argh, it probably would be somewhat more efficient if in the ISR we call it inside the if (irq_status & SX12xx_IRQ_RX_DONE) { } block.

So, another approach could be that all drivers should get this function, but would be empty for the sx127x and sx1280 drivers.

We also could introduce a GetAndClearIrqStatus(SX12xx_IRQ_ALL) functions which returns the irqstatus but also does the ClearIrqStatus(SX12xx_IRQ_ALL); and for the sx126x also the ClearRxEvent. From an API side might be the cleanest.

what would you think?

Imen-D:
Very great Olliw,
I have prefer for 2nd option more cleanest sure

olliw42:
so, we opt for a new function GetAndClearIrqStatus(SX12xx_IRQ_ALL)

could you add it to the sx126x driver and do the change in the isrs in the rx and tx main codes?

pl confirm explicitely that this works for you (if it does LOL)

I will then deal with adding this function to the sx127x and sx1289 drivers (the latter requires a submodule update, so a tiny bit more complicated)

Imen-D:
Ok Olliw, I will modify the driver and open a new PR

olliw42:
fantastic

(usually one would add to the PR, and do a squash, but doing as you do is perfectly fine either :))

PR #6, opened 2022-02-16
introduce GetAndClearIrq and clear event on sx1262, also fix the calc_sx_power

Imen-D:
Sx1262 setTxParameter (power) use actual value in dbm. with set PA config to max 22dbm we have range -9 to 22 dbm
and value we pass on setTxParameter is actual value (-9 ... 22)

### DIFF
--- a/mLRS/Common/sx-drivers/sx126x.cpp
+++ b/mLRS/Common/sx-drivers/sx126x.cpp
@@ -103,6 +103,7 @@ uint8_t out_buf[SX126X_SPI_BUF_SIZE];
     SpiTransfer(out_buf, data, len);
     SpiDeselect();
     // no delay according to semtech driver
+    ClearRxEvent();
 }
 
 
@@ -300,17 +301,6 @@ uint8_t buf[3];
     buf[2] = (uint8_t) (tmo_periodbase & 0x0000FF);
     WriteCommand(SX126X_CMD_SET_RX, buf, 3);
 
-    // 15.3 Implicit Header Mode Timeout Behavior   p. 103
-    // 15.3.1 Description
-    // When receiving LoRa® packets in Rx mode with Timeout active, and no header (Implicit Mode), the timer responsible for
-    // generating the Timeout (based on the RTC timer) is not stopped on RxDone event. Therefore, it may trigger an unexpected
-    // timeout in any subsequent mode where the RTC isn’t re-invoked, and therefore reset and re-programmed.
-    if (_header_type == SX126X_LORA_HEADER_IMPLICIT) {
-        WriteRegister(SX126X_REG_RTC_STOP, 0x00);
-        buf[0] = ReadRegister(SX126X_REG_RTC_EVENT);
-        buf[0] |= 0x02;
-        WriteRegister(SX126X_REG_RTC_EVENT, buf[0]);
-    }
     SetDelay(100); // semtech driver says 100 us
 }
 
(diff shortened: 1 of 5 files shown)


olliw42:
this looks great

can you pl explicitely confrim that it does work now for you as you expect it to work? yes/no

there is tiny this formatting thing which I don't like to 100%, but no need for you to do anything, I will do this as follow up

----
concerning the power, 

I first want to state that this will have to be reworked globally, it needs to become more "complicated" to allow also handling boards like the R9M and so on. I however like to discuss it briefly, so I understand how the situation is with the sx1262: 

I thought over this, and had the same considerations as you, but came to a different conclusion. As much as I understand:

setTxParameter() is supposed to take the sx' 'internal' values for setting the power, which in this case are SX126X_POWER_ENUM. the variable holding this value is sx_power

these values however are, as it seems to me, and as you seem to confirm, values in units of dBm

the conversion function calc_sx_power() is supposed to take the real power in units in dBm, let's call it power_dBm

therefore, if there is no amplifier, the relation is

sx_power = power_dBm

an amplifier would reduce the setting for the sx, and therefore sx_power = power_dBm - power_of_PA_dbm, but the PA power is exactly POWER_GAIN_DBM.

so, with the explanation you are giving I come again to the result

sx_power = power - POWER_GAIN_DBM.

Would you agree with this argumentation, and conclusion?

The define POWER_SX126X_MAX_DBM is supposed to give the sx value which is maximally allowed. In your case you should have POWER_GAIN_DBM = 0 and POWER_SX126X_MAX_DBM = SX126X_POWER_22_DBM. Therfore your formula would give

sx_power = 10 - 22 - 0 = -12

which if the above is correct would be incorrect, since for achieving 10 dBm one should set sx_power = 10.

Correct, right?

Imen-D:
I use module from Ebyte E22-900M. They claimed radio have output max is 30dbm, let says this radio have ouput max 30dbm, so POWER_GAIN_DBM is actually power gain PA which is 30-22 = 8 right ?.
This is make sense then with the formula. so I think the formula is right, then I can change the POWER_GAIN_DBM on my hal diy board.
Btw I can confirm the telemetry is working, not yet to the RC control, will post some video later. but I think if we have telemetry working, then we will be happy 
thanks

olliw42:
> Btw I can confirm the telemetry is working, not yet to the RC control, will post some video later. but I think if we have telemetry working, then we will be happy

great, so I'm going to merge this. THX A LOT !!

> I use module from Ebyte E22-900M. They claimed radio have output max is 30dbm, let says this radio have ouput max 30dbm, so POWER_GAIN_DBM is actually power gain PA which is 30-22 = 8 right ?. This is make sense then with the formula. so I think the formula is right, then I can change the POWER_GAIN_DBM on my hal diy board.

hmmmm

in that case you should have 

POWER_GAIN_DBM = 8
POWER_SX126X_MAX_DBM = SX126X_POWER_22_DBM (= 22)

so, for the above example of 10 dBm, your formula would give 

sx_power = 10 - 22 - 8 = -20 !!!!!!!!!!!!!!!!!!!

while my formula would give 

sx_power = 10 - 8 = +2

I continue to think your formula is incorrect

BTW, the gain of the amplifier most likely won't be the 8 dBm estimated before, but more like 14 dBm or so (you need to look up the datasheet). Moreobver, there might be a maximum input power alloed to the PA.

Imen-D:
Hi Olliw, sorry for my mistake, I think your formula is right, I just wrong define my POWER_GAIN_DBM.
So the right value might be 8, before is 30. solve now

olliw42:
ok, great
so, I will change this too :)

thx for this, I look forward to see this working

once it does, I will merge to the main branch :)

PR #7, opened 2022-02-18
Fix channel shifting

Imen-D:


Imen-D:
I just PR on fix channel shifting only, why update sx12xx-lib is include :D

Imen-D:
ok, sure

PR #10, opened 2022-02-27
Add send mavlink radio status on rx side 

Imen-D:
Hi olliw, with this additional commit, we will have a big different in downlink rate and packet lost. I got working on 57600 baud, just have packet lost when first connect to mission planner as FC send all parameter to mission planner. After all parameter had downloaded, I didn't have packet lost anymore. For this radio send status working we need a large buffer size on serial rx 8k working for me now

### DIFF
--- a/mLRS/Common/common.h
+++ b/mLRS/Common/common.h
@@ -355,7 +355,7 @@ class tSerialPort : public tSerialBase
     bool available(void) { return uartb_rx_available(); }
     char getc(void) { return uartb_getc(); }
     void flush(void) { uartb_rx_flush(); uartb_tx_flush(); }
-    uint16_t bytesAvailable(void) { return uartb_rx_bytes_available();}
+    uint16_t bytesAvailable(void) { return uartb_rx_bytesavailable();}
 #endif
 };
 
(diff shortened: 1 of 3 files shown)


olliw42:
many thx, much appreciated
I will not merge soon however, since (1) this way of injecting it, which mirrors what is there for tx, I had added only for experimental reasons and is not how it eventually should be done, and because (2) it is only "half" of the story, as you noted yourself, we especially want it work also for parameter upload and so on. 
:)

Imen-D:
Right, I just mirror on tx side. As you said this is only experimental, but effectively hold the flight control to send to much data stream rate to the rx side, of course this might not how should be, but at least this is working now. 

Thanks for your work Olliw

olliw42:
ok, you made me changing my mind :D
I wanted to do this later, but it doesn't harm, so I merge.
MANY THX!

olliw42:
@Imen-D 

I have made some tidying changes. Could you please pull the latest main and test if it is still working fine ?

I mainly moved the function txBufSpace to the serial base class, and I also took away the >> 3.

THX!

PR #20, opened 2022-07-31
Don't use back-slash in #includes to allow builds on non-windows systems

brad112358:
I must be the first one building this on Linux.


olliw42:
oh, thx!
would you want to add instructions on how to compile, to e.g. the README in the start folder?

brad112358:
I use Debian Linux and only had to download and install the Debian version of the IDE and a few dependencies for it and for your python scripts.  Unfortunately, I didn't keep notes.  I think most developers would have no trouble, but I suppose it would be good to add some hints.  Perhaps I'll install it on another system and work up another pull request later.
Thanks for the quick merge!

olliw42:
@brad112358 
thx
don't worry

btw, I have made an attempt at injecting RC_CHANNELS in this branch here: https://github.com/olliw42/mLRS/tree/dev-sendrcchannels
I haven't tested it at all, but it may serve you as a starting point

PR #21, opened 2022-08-02
rx, inject send rc_channels, 230400 baudrate

brad112358:
Creating this pull request so we can discuss this here:


mLRS/Common/setup.h
```diff
@@ -407,6 +409,7 @@ void setup_configure(void)
     case SERIAL_BAUDRATE_38400: Config.SerialBaudrate = 38400; break;
     case SERIAL_BAUDRATE_57600: Config.SerialBaudrate = 57600; break;
     case SERIAL_BAUDRATE_115200: Config.SerialBaudrate = 115200; break;
+    case SERIAL_BAUDRATE_230400: Config.SerialBaudrate = 23040; break;
```
brad112358, line 412: Is this a typo?  Should it be 230400?

mLRS/CommonRx/mavlink_interface_rx.h
```diff
@@ -183,4 +243,41 @@ if(txbuf>90) dbg.puts("-20 "); else dbg.puts("+-0 ");*/
 }
 
 
+void MavlinkBase::generate_rc_channels_override(void)
+{
+    fmav_msg_rc_channels_override_pack(
+        &msg_serial_out,
+        RADIO_STATUS_SYSTEM_ID, MAV_COMP_ID_TELEMETRY_RADIO, // sysid, SiK uses 51, 68
```
brad112358, line 250: The FC autopilot will probably only accept the override message if the system ID matches, so we should use 255 here which is what the ground control system is using by default. 68 is still good for the component ID.

olliw42: you are correct, good point
the correct value would be AP's MY_GCS parameter, but for the moment 255 should indeed do it
https://github.com/ArduPilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L3461-L3463

mLRS/Common/setup.h
olliw42: oh yeah LOL , sure 230400
I'm btw not sure it's really usefull to go beyond 115200, it's probably more for testing at this point in time

PR #23, opened 2022-09-12
Allow for 32 bit TIM2

brad112358:
The cause of the large bursts of lost frames on the R9 Tx side turns out to be a problem on the Rx side.  But it's not a difference with the radio chip.  It's a difference with the STM32 controller used.  The code in clock.h assumed that TIM2 has a 16 bit wide counter, but on the STM32L4, TIM2 is 32 bits wide.  So, when Reset() is called, after CNT is larger than 2^16, the timer never fires and the receiver looses sync when the next frame is lost.

Note, I was unable to test this code on other receivers so this still needs testing.  I'm guessing that passing 0xFFFFFFFF in tim_init_up is ok even for 16 bit timers and that it will be truncated, but I haven't actually checked to be sure..  This is probably also not the only way to fix this bug.

I still see many more lost frames on the Tx side than on the Rx side.  Do you see this with other transmitters?

### DIFF
--- a/mLRS/CommonRx/clock.h
+++ b/mLRS/CommonRx/clock.h
@@ -43,7 +43,7 @@ class ClockBase
     void init_isr_off(void);
     void enable_isr(void);
 
-    uint32_t tim_10us(void);
+    uint16_t tim_10us(void);
 };
 
 
@@ -93,7 +93,7 @@ void ClockBase::enable_isr(void)
 }
 
 
-uint32_t ClockBase::tim_10us(void)
+uint16_t ClockBase::tim_10us(void)
 {
     return CLOCK_TIMx->CNT;
 }
--- a/mLRS/CommonRx/clock.h
+++ b/mLRS/CommonRx/clock.h
@@ -62,9 +62,9 @@ void ClockBase::Reset(void)
     if (!CLOCK_PERIOD_10US) while (1) {}
 
     __disable_irq();
-    CLOCK_TIMx->CNT = 0;
-    CLOCK_TIMx->CCR1 = CLOCK_PERIOD_10US;
-    CLOCK_TIMx->CCR3 = CLOCK_SHIFT_10US;
+    uint32_t CNT = CLOCK_TIMx->CNT;
+    CLOCK_TIMx->CCR1 = CNT + CLOCK_PERIOD_10US;
+    CLOCK_TIMx->CCR3 = CNT + CLOCK_SHIFT_10US;
     LL_TIM_ClearFlag_CC1(CLOCK_TIMx); // important to do
     LL_TIM_ClearFlag_CC3(CLOCK_TIMx);
     __enable_irq();
--- a/mLRS/CommonRx/clock.h
+++ b/mLRS/CommonRx/clock.h
@@ -13,6 +13,7 @@
 // we use a 10us time base, so that overrun is 655 ms
 // 65 ms would only be 3 packets
 // I have tested it with 1us time base, and it also works fine, but hey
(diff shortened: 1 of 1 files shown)


brad112358:
As I said, I have only tested it with the controller that has the 32 bit timer.  The main problem with the original version of Reset is the type of CNT which can only store 16 bits, and of course, this causes restricts maximum value of the two ->CCR registers to be much smaller than the full range which ->CNT will be become.  Thus, ->CNT will quickly be larger than the ->CCR registers and the interrupt won't trigger for eons until ->CNT wraps around.

I'll try a different approach, but I still won't be able to fully test it.

brad112358:
I pushed a change which avoid resetting the counter in Reset().  In this case, we are depending on the assignments to CLOCK_TIMx->CCR1to do truncation to 16 bits.  But, this needs to be tested with a 16 bit counter.  I think the STM32L4xxxx also has 16 bit timers so I may be able to mostly test this myself by switching to a different timer.

olliw42:
this looks great

tim_10us() should ot be expanded to u32 but remain at u6, otherwise it would fail for 16bit timers

if you can test thoroghly on boards with 32 bit tim2, this would be great. I will do too. I can test it on 16bit tim2 boards.

brad112358:
> tim_10us() should ot be expanded to u32 but remain at u6, otherwise it would fail for 16bit timers

I'm not sure about this.  Why would it fail?  If CLOCK_TIMx->CNT is 16 bits, the return value will just be promoted to 32 bit with the high order bits 0.  But, perhaps it is more consistent if tim_10us only returns the low order 16 bits as you say.  The code must expect it to wrap anyway in case we only have a 16 bit counter.

I tried to test with TIM15 which on theSTM32L4xxxx is a 16 bit general purpose timer, but something must be wrong in the HAL layer because the compiler tells me TIM15 isn't defined.  I'll try to track it down.

brad112358:
> it won't wrap correctly and subtraction can give wrong results for 16 bit timers
OK.   Better always 16 bits then.

olliw42:
I have running it with the two changes (u16->u32 and 0xFFFFFFFF) on my R9 system ... while I had plenty of bursts before I don't had any in the last 25-30 minutes ... it indeed seems the bursts are GONE !!!! you are genius

olliw42:
as much as I can tell it also works great on my E22 system ... for which TIM2 is also a 32 bit timer however !!!!

it is strange that I couldn't see such burst here at all .... is there maybe a difference in how the timers exactly work ?

anyway, I will test on my E28 system, which uses F103 ...

olliw42:
so, I cannot notice any issue on my SX1280 system, which uses F103 ... so, I think this really has nailed the issue.

I'm going to merge and close.

MANY THX again, Brad. This was bug was not easy to spot, excellent work.

olliw42:
dammed, I need first to merge LOL

brad112358:
Thanks Olli.  Yes, it was a subtle one.  Finding it was more luck, intuition and code review than anything else.  Especially since I was at first focused on the sx side rather than the stm32 side.  I think the R9 is much more usable now, but I'm still a bit concerned about the number of lost frames I'm seeing on the Tx.  Do you still see more frame loss on the R9M than with other Tx?

brad112358:
> it is strange that I couldn't see such burst here at all .... is there maybe a difference in how the timers exactly work in L4's and G4's?
I don't see a significant difference in TIM2 between these two product lines.  Does your E22 test platform use diversity?  If no frames are lost on the Rx, you would not notice this issue I think.  That was part of the reason I started looking at timing on the Rx side.  I suspect you might reproduce it at greater distance with low power.

PR #28, opened 2022-10-16
Update CLI.md

geofrancis:


olliw42:
thx
I see the intention but won't go with it
it e.g. would mean additional maintenance burden (and I really don't like burden be inflated LOL)
also I think personal cheat sheets are, well, personal :)

PR #32, opened 2022-10-26
Frsky R9mx as tx

vrquaeler:
Hi @olliw42,
As mentioned in the **rc-groups** thread I was interested in a "sub-2-gramm" Sik-radio replacement. 
Here is a (preliminary) proposal of a "minimal TX" based on R9MX hardware. 
AFAIK it was one of the first " desired configurations" I found in the "design decisions". 

[https://github.com/olliw42/mLRS/blob/main/mLRS/Common/design_decissions.h#L19-L27]()

Due to hardware limitations it doesn't meet all the requirements. Nonetheless the new very light weight "Sik-Radio-replacements" will find their usage cases. 

All mod's are based on your idea's I found in the code, like the dual use of the serial and CLI-port. 
Next a modified version of the double usage of the LED pins on "tx-R9M" with the DIP switches. 
The Bind button is (temporarily) used to switch between Serial and CLI.

The usage is simple: 
Powering up the R9MX without holding the button (default) will boot in "Serial" mode.
To enter the CLI mode press the "Bind button" while powering up. The device run's in CLI mode. RX and TX can be configured as usual via CLI commands. 

Hopefully I haven't broken anything. Its my very first "CubeIDE" project.


olliw42:
@vrquaeler 
many thx for this, much appreciated!

overall it looks quite good, well done. I will have to very carefully check the project files etc., and it might be that I rather delete them and generate from scratch again, but that's not your concern.

otherwise, I like the idea of how to use the button, only needs tiny cleaning I guess

please remove the TX_R9MM defines in device_conf.h and hal.h. I will not support this board for tx (I rather would remove the R9MM alltogether LOL). The number of targets is really large already (and I guess at some point we will have to find a better way of handling this). If you want it you may do so in your fork.

the main thing I am missing is an IN port. I understand you think in terms of "SiK radio replacement", yet I don't like that it is limited to this. I am not sure from top of my head if the hardware allows an IN port or not, but I would kindly ask you to carefully check if it does, and if it does to add it.
There are  Pin3 SPort/FPort PA5 ,  Pin4 Inv SPort   PB11,  Pin5 SBusOut  PA2 inverted, and Pin6 SBusIn. Maybe one is suitable for IN.

vrquaeler:
Thank's for the remarks. 
I will close it for the newer PR.

PR #33, opened 2022-10-27
R9 mx as tx 2nd try

vrquaeler:
Hi @olliw42,

Here you will find a cleaned version of the first PR.

PS: For the time being the R9MM is my favorite target as "Sik-radio-replacement". It's even smaller and lighter. And I have a few laying around here back from the days you got one free with any Frsky-order.


vrquaeler:
Replaces 
[https://github.com/olliw42/mLRS/pull/32](url)

olliw42:
@vrquaeler 
thx

first, you do not need to do a new PR each time you want to do changes. Simply do the changes in your fork, commit them in your fork, and then push them to the repo.

concerning tx R9MM, I hear what you are saying, but I stick to that I will not add it to main 
(I'm sure once you will have done a firmware update the 5th time you will start to hate R9MM)
(you hadn't pushed this target anyhow, so it was dead code anyways :))

more comments you find inline.

there is a change also in rx-R9MM_l433cb/.project, which should not have occured as this target is not subject here, pl undo this modification, at least for the sake of this PR

great job :) many thx

olliw42:
pl do not forget to check for the possibility of an IN port :)

vrquaeler:
Hi Olli, 
I totally agree the R9MM is NOT the most preferred DEV-board. The application in a weight constrained vehicle (sub- 250 gramms) is the one and only benefit.

I try to figure out the optional "IN port" thing. 
From my understanding a real  **serial_rx** is needed. 

I have no useful test setup here yet.  I could give "_UART_USE_UART3_" a try. Then the "**Pin4 Inv SPort PB11**" will be the new **IN pin**.
First I have to figure out a simple way to check and test this.

olliw42:
> From my understanding a real **serial_rx** is needed.

yes, a pheripheral uart rx

> I have no useful test setup here yet

hold on, you are saying this code here you have NOT tested to work on an actual system?

vrquaeler:
No worries. The last change is the cleanup only and tested on real hardware. 

The code change for the _"in port"_ **WILL** be tested on real hardware too before you will see it. Until now I am in progress of evaluating the possibilities. I just wanted to keep you informed about my thoughts.

What I found until now:
- The pin **PB11** seems to be the one and only which is accessible and supports the needed functionality. 
- With the use case "Sik radio replacement" in mind there is no need to have a _in port_.
- Years ago I was fiddling around with different diy transmitter modules. If i remember correctly I can't connect a MC to the module bay pins without level converting. I have to find a setup first. Preferably some thing which doesn't burn the R9MX.

olliw42:
ah, indeed, PB11 is uart3 rx ... maybe it's as simple as that

vrquaeler:
Thanks for info about the allowed signal level on the L433cb. 

Here is a updated PR. The PR is tested on real hardware. It runs smooth as silk with "Inverted SBUS" on the "IN-port".
I have some doubt's if the L433cb can do "usart inversion". With the available hardware I was not able to make it run properly. But this could be related to the signal leveling too.

I made a 3 hour endurance test and the R9MX went through without **ANY** packet loss. 
In the same test conditions with R9M (red tx module) I got always some "packet loss". 

The setup:
- 
- Plane in the garden ca. 7 meters away.
- TX next to PC in the living room. 
- One wall in between.
- Both tx and rx at lowest power settings.
- In both cases the plane was equipped with (the same) R9MM. 


Please could you review the latest changes? 

One question is left. I had to add `#define USE_IN ` to make it work.
The line can be found here:
[#define USE_IN                      // hal.h#L191 should do it](https://github.com/vrquaeler/mLRS/blob/f9d0e2e3ac2726ae9dd77aeff3d5e6ebeabbbfc5/mLRS/Common/hal/tx-hal-R9MX-868-l433cb.h#L30)
From my understanding this line should not be necessary but I could not wrap my head around it.

olliw42:
many thx, fantastic

you have however not adressed the comments I have made to some code line, pl adress them

USE_IN should not be here, delete it

I guess I will have to recreate teh project, and if it is just becasue I have decided that since now on every new target must come with a detailed description of how it is created ... it's getting quite confusing now ... but I won't want to burrden you with that no

olliw42:
@vrquaeler 
ok, new plan, I will merge this now, since I want to move forward with some stuff myself and this likely would create merge conflicts for this PR here and you would have to rebase, which I am not sure you are confortable with. Since I will do the changes as indicated in the above, but won't find time to test (I only will test for compiling), I will ask you to test, once the new code is in main. Hope this makes sense to you.
:)

vrquaeler:
@olliw42 
Thank you very much for all the positive comments, the right suggestions and the needed patience. 
This is the biggest PR I have ever done. :-)

For sure I will do the testing. :+1:

olliw42:
@vrquaeler 
don't worry, I well remember what seruious issues I had with getting along with git & github

I've done the stuff now, I feel a bit unsure it's all ok since I found it more difficult to sync with what I had done in the meantime, so hopefully I haven't mixed things up.

would be fantastic if you could give as much of the R9 stuff testing as you can. Concerning the tx R9MX, I guess there is a questionmark as regards the in port, you may find now that you have to use e.g. sbus inv instead of sbus. If so it's easy to correct for in a "bug" commit, I just didn't know which way around it is.

many thx for having contributed this.

vrquaeler:
@olliw42 
Thank's for fixing my left over "residues" in `device_conf.h`. Was already preparing a bug-fix. My brain seems to be way slower than yours. 

I will give it a test build. 

Regarding the "SBUS-inv" on the "IN-port" of the tx-R9MX. Here I had major issues to get it work. The one and only way to make it work was "SBUS-inv" on IN-port.
The "DEVICE_HAS_IN" of the R9MX seems to be a regular _usart_ without possibilities for inversion. A regular _usart_ (without hardware inversion) should lead to "SBUS inv" only. I will check this.

PS: Nice instructions for new targets too. I had to try some (wrong) permutations of the steps by my own. Which did not work out very well. ;-)

vrquaeler:
One more:
I have to do the belonging documentation for the tx-R9MX too.

olliw42:
> The one and only way to make it work was "SBUS-inv" on IN-port.

hm, crsf did not work? 
crsf and sbus-inv are both normal uart levels, and hence if one works also the other should work

the L433CB used in teh R9MX has the capability to invert RX, so it should be able to do both

> PS: Nice instructions for new targets too. I had to try some (wrong) permutations of the steps by my own. Which did not work out very well. ;-)

as indicated, I will henceforth require every target to come with a most detailed description of its creation

> I have to do the belonging documentation for the tx-R9MX too.

this would be most welcome
it is a bit of an issue however too, since it's not clear where to actually place it
the current approach to add it to the git repo does have serious limitations, one of it is that it spoils the commit structure with wiki updates ...
maybe one should create a github wiki repo

vrquaeler:
have not tried "crsf" input. 
At the menu item: `Tx_In_Mode` I have only the choice of : `0 = sbus` or `1 = sbus inv` .

If i remember correctly, somewhere in the code the BAUD rate was set to 100000.
I was assuming this gives SBUS/SBUS-inv only.
Found it:
https://github.com/vrquaeler/mLRS/blob/R9MX_as_TX_2nd/mLRS/Common/hal/tx-hal-R9MX-868-l433cb.h#L68

> maybe one should create a github wiki repo

Seems to work for Ardupilot ...
For the mLRS-hardware you have a separate github-repo too.

Have to test your new version first...

olliw42:
> At the menu item: Tx_In_Mode I have only the choice of : 0 = sbus or 1 = sbus inv .

argh, my fault, this here is a tx not a rx LOL

if you can get both working, and if "sbus" works for then you expect sbus and "sbus inv" works for then you expect sbus inv, it's all good :)

vrquaeler:
@olliw42 
Short update:
`sbus` and `sbus inv` are working properly. :+1: :+1:

PR #35, opened 2022-11-02
tx-R9MX docu

vrquaeler:
Hi @olliw42 

Here the promised docs for tx-R9MX. It's for advanced users only.


olliw42:
@vrquaeler 
great thx
there are some few spelling errors, and I guess I would want some few changes
so what I'm doing is to merge it (and close it), and then pr my changes

on a longer run we need to find a way to separate the docu from the code

many thx again, much appreciated

PR #36, opened 2022-11-14
Don't assume the current directory is in the execution path

brad112358:
This should fix the trouble some people have with run_setup.py on Linux
I have not checked that this still works on Windows, but I think it should.

### DIFF
--- a/run_setup.py
+++ b/run_setup.py
@@ -35,7 +35,7 @@ def generate_mavlink_c_library():
     print(' run fmav_generate_c_library.py')
     print('----------------------------------------')
     os.chdir(os.path.join(mLRSdirectory,'Common','mavlink'))
-    os.system('fmav_generate_c_library.py')
+    os.system('./fmav_generate_c_library.py')
     print('# DONE #')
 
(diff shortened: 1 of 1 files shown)


olliw42:
else one might need something like 
```
    if os.name == 'nt': # the os is Windows
        os.system('fmav_generate_c_library.py')
    else:
        os.system('./fmav_generate_c_library.py')
```

I guess

brad112358:
> @brad112358 many thanks! it unfortunately is not working on Win also... would os.system(os.path.join('.','fmav_generate_c_library.py')) work on linux?
Crap.  I had read that Python on windows did the translation from / to \  But I guess not in this case.
Yes, I think that would work.  I'll test it...

brad112358:
Yes, your os.path.join suggestion works on Linux.   I'm sure it works on Windows too (and not just because you probably already tested it).  :)

olliw42:
yes, I did test the suggested change with join ... great to hear it also works on linux
so, will merge
THANKS!!!

PR #37, opened 2022-12-12
Add calls to implement Automatic Frequency Correction.  This seems to…

brad112358:
… eliminate occasional lost frames on the R9M!

I have not even compiled this for targets other than R9M and R9MX.  Please let me know if I broke anything and if this eliminates lost frames on the R9M as it seems to do for me.

### DIFF
--- a/mLRS/CommonRx/mlrs-rx.cpp
+++ b/mLRS/CommonRx/mlrs-rx.cpp
@@ -679,6 +679,7 @@ IF_ANTENNA1(
           irq_status = 0;
           bool do_clock_reset = (link_rx2_status == RX_STATUS_NONE);
           link_rx1_status = do_receive(ANTENNA_1, do_clock_reset);
+	  if (link_rx1_status == RX_STATUS_VALID) sx.HandleAFC();
           DBG_MAIN_SLIM(dbg.puts("!");)
         }
       }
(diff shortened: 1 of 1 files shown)


olliw42:
@brad112358 
MANY THX!!
looks simple enough, but I wonder there this should be actually be placed. Would it be possible to place it into the sx127x driver itself? At least from a code arcithecture point of view this would be the place where I would think it shoudl go (i.e., as low-level as possible)

olliw42:
we may also have to scan the code, I think I added a weak precusor for allowing afc, could be removed (was it in fhss?)

brad112358:
And in the other PR comments, you seem to be wanting to move more code into the upper layers.  :-)  
I put this here because I thought it might be important to use FEI info only from correctly received messages with correct sync word and CRC.  I guess this is just me not trusting the chip completely but I don't want noise or frames from another transmitter to cause us to change the frequency.

olliw42:
> I put this here because I thought it might be important to use FEI info only from correctly received messages with correct sync word and CRC

ah, ok
checking the syncword is probably a crucial idea, so one stays with the transmitter, but crc might mean we don't do much AFC when the connection is weak ... not sure if that is relevant, the freq correction probably doesn't change much with time

olliw42:
@brad112358 
MANY THX!
This will be a huge and signifciant improvement for the R9 users.

PR #38, opened 2023-01-18
Fix ftp message rewriting

brad112358:
I discovered some problems when testing with Mavproxy and Mission Planner.
Without these changes, Mavproxy gets stuck trying to download parameters via ftp.

Use strncmp to allow for options at the end of the URL string.  Both Mission planner and Mavproxy request "@PARAM/param.pck?-withdefaults=1"

Mavproxy uses target component of 0 (all components).

Need to recalculate CRC or autopilot will just discard the message.

This has been tested with both Ardupilot and PX4 and Mavproxy, Mission Planner, and QGroundControl.

### DIFF
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -180,15 +180,17 @@ void MavlinkBase::putc(char c)
         // this will make ArduPilot to response with a NACK:FileNotFound
         // which will make the GCS to quickly jump to normal parameter upload
         if (msg_serial_out.msgid == FASTMAVLINK_MSG_ID_FILE_TRANSFER_PROTOCOL) {
+            uint8_t target_component = msg_serial_out.payload[2];
             uint8_t opcode = msg_serial_out.payload[6];
             char* url = (char*)(msg_serial_out.payload + 15);
-            if (opcode == MAVFTP_OPCODE_OpenFileRO) {
+            if (((target_component == MAV_COMP_ID_AUTOPILOT1) || (target_component == MAV_COMP_ID_ALL)) && (opcode == MAVFTP_OPCODE_OpenFileRO)) {
 	        if (!strncmp(url, "@PARAM/param.pck", 16)) {
                     // now fake it to "@xARAM/xaram.xck"
                     url[1] = url[7] = url[13] = 'x';
 		    // Need to recalculate CRC
 		    uint16_t crc = fmav_crc_calculate(&(buf_link_in[1]), FASTMAVLINK_HEADER_V2_LEN - 1);
 		    fmav_crc_accumulate_buf(&crc, msg_serial_out.payload, msg_serial_out.len);
+		    //uint16_t crc = fmav_crc_calculate(&(buf_link_in[1]), FASTMAVLINK_HEADER_V2_LEN - 1 + msg_serial_out.len);
 		    fmav_crc_accumulate(&crc, msg_serial_out.crc_extra);
 		    msg_serial_out.checksum = crc;
                 }
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -180,14 +180,17 @@ void MavlinkBase::putc(char c)
         // this will make ArduPilot to response with a NACK:FileNotFound
         // which will make the GCS to quickly jump to normal parameter upload
         if (msg_serial_out.msgid == FASTMAVLINK_MSG_ID_FILE_TRANSFER_PROTOCOL) {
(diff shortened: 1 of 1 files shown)


olliw42:
THX

interesting

which MP version are you talking about? 1.3.77 or 1.3.79?
I ask because I found this to work before for me for both MP and QGC, but also have found that MP has made some little but important changes from 1.3.77 to 1.3.79. Just want to understand the situation.

mLRS/CommonRx/mavlink_interface_rx.h
```diff
@@ -180,14 +180,17 @@ void MavlinkBase::putc(char c)
...
-            if ((target_component == MAV_COMP_ID_AUTOPILOT1) && (opcode == MAVFTP_OPCODE_OpenFileRO)) {
-                if (!strcmp(url, "@PARAM/param.pck")) {
+            if (opcode == MAVFTP_OPCODE_OpenFileRO) {
+	        if (!strncmp(url, "@PARAM/param.pck", 16)) {
                     // now fake it to "@xARAM/xaram.xck"
                     url[1] = url[7] = url[13] = 'x';
-                    //return; // for MissionPlanner it doesn't matter what we do, but for QGC?
+		    // Need to recalculate CRC
+		    uint16_t crc = fmav_crc_calculate(&(buf_link_in[1]), FASTMAVLINK_HEADER_V2_LEN - 1);
+		    fmav_crc_accumulate_buf(&crc, msg_serial_out.payload, msg_serial_out.len);
```
olliw42, line 191: I need to think more about this, but my first impulse:
- can't line 190&191 be done together by giving the correct length?
- more seious: is line 191 correct? If I'm not mistaken, msg_ser_out.len is the length of the complete frame as it has to be send over the serial so that using it here the calculation actually goes way beyond the frame?

olliw42 (changes requested):

olliw42:
> I upgraded to 1.3.79 before I started testing. 

thx, makes sense. I started using 1.3.79 only recently, since it also gave issues with STorM32, which I sorted out just a week ago or so. 

this are generally good changes, I'm glad you noticed.

olliw42:
is mavproxy using a spscific sysid, or can this vary largely depending on the user?

I feel uncomfortable to just always do this, i.e. not check target at all, and like to be more mavlink-ish. One often have to deal with that ArduPilot often isn't, but at least I'd like to attempt.

mLRS/CommonRx/mavlink_interface_rx.h
brad112358: Yes, I think 190 and 191 could be combined.  Do you prefer that?  I was following examples in the library.
No, I checked the code in fmav_frame_buf_to_msg().  len is the payload length.  Also, I got the NAK with this code, and not without it, so the calculated checksum must be correct.

brad112358:
> I feel uncomfortable to just always do this, i.e. not check target at all, and like to be more mavlink-ish. One often have to deal with that ArduPilot often isn't, but at least I'd like to attempt.
I guess we can check what targets the all the gcs are using, but squashing only param.pck seems specific enough to me since there is a fallback for parameter download.
To be honest, I'm uncomfortable with this entire rewriting idea, but it makes parameter download work without fixing PX4 and Ardupilot which I think both should be paying attention to txbuf.  Perhaps this should all be conditional on a setting so the messages can be left unmolested in case the autopilots are fixed?

olliw42:
I agree that this is a totally bloody hack, and we shouldn't have to do it. But that's real world. I'm not so happy with a user setting, noone will understand. I guess if the autopilots/gcses ever get that right (which they won't, if you have followed a discussion I triggered recently), I rather would change the mLRS code and release that.

so, let's add a check ((target_component == 0) || (target_component == autopilot_id))

mLRS/CommonRx/mavlink_interface_rx.h
olliw42: I was mistaken, haven't done a crc calc function but a function which was reconstructing the extra crc from an unknown message. 
So, I suggest to change line 190&191 to `uint16_t crc = fmav_crc_calculate(&(buf_link_in[1]), FASTMAVLINK_HEADER_V2_LEN - 1 + msg_serial_out.len);`. If you are ok with this If not, leave it as it is.

brad112358: I just realized we can't combine 190 and 191 in this way, because we have modified the payload in msg_serial_out without changing it in buf_link_in.  So I'll leave this part of the code as is.

olliw42: ah, sure, good catch!
maybe you could add a comment // use xxxx because yyy or so :)

PR #39, opened 2023-01-18
Replace commented out broken code with reason

brad112358:
Sorry.  I didn't mean to leave this comment in.  Replace it with a more meaningful comment.

### DIFF
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -190,7 +190,7 @@ void MavlinkBase::putc(char c)
 		    // Need to recalculate CRC
 		    uint16_t crc = fmav_crc_calculate(&(buf_link_in[1]), FASTMAVLINK_HEADER_V2_LEN - 1);
 		    fmav_crc_accumulate_buf(&crc, msg_serial_out.payload, msg_serial_out.len);
-		    //uint16_t crc = fmav_crc_calculate(&(buf_link_in[1]), FASTMAVLINK_HEADER_V2_LEN - 1 + msg_serial_out.len);
+		    // Can't combine above 2 lines because we didn't modify the payload of buf_link_in
 		    fmav_crc_accumulate(&crc, msg_serial_out.crc_extra);
 		    msg_serial_out.checksum = crc;
                 }


olliw42:
I have seen that but was okay since I'll push some changes in a minute
so I close without merge as it would screw up my branch I'm preparing :)
THX anyway!!!

brad112358:
OK.  Thanks!

brad112358:
BTW, I do have another PR I am preparing which speeds up parameter download and makes it more reliable.

olliw42:
interesting
I am actually also working on such a thing using a newly introduced mavlink message to pack the params
not yet finished, but I'm afraid we may overlap and your changes may collide
the idea is to collect param messages on the rx side and compress them into a PARAM_VALUE_ARRAY message and send this to the tx instead 8screws up seq no, but hey...). should increase it by a factor of >3.
This https://github.com/mavlink/mavlink/issues/1918 is related.
concerning reliability I think anything should go ontop of this resend thing done, still had no time to work it out, sadly (I'm lately very busy with STorM32)

brad112358:
This is very interesting.  I haven't yet taken the time required to read the entire discussion you referenced.  I will need to look at it again to understand why your proposal is better than making FTP work correctly on low bandwidth links.  I wasn't quite ready, but I think I'll issue a PR for the code I have now just so you can see what I've been working on and to give us a place to discuss it.

olliw42:
yes, pl, PR it, I'm curious
I just wanted to indicate that I might maybe not immediately jump onto it, due to this alternative approach :).
The suggestion I've made didn't fly and won't come to live. Hence my effort now within mLRS. Fundamentally it is not "better" than FTP in terms of download speed, the issue is concept-wise and number of protocols  wise ... and I mean, all GCSes and ArduPilot still struggle to get it right, even after several years... 
In ArduPilot v4.3.3 ftp should work better now, iff (if and only if) you set the baudrate to a correspondingly low value, which we (I) don't want at all. Still no txbuf handling yet. Furthermore, ArduPilot started now to send even metadata (this default thing which yopur PR corrected for), which makes the download unneccessarily large = slower (it is really sad that they again screw things up, and don't use the comp_info mechanism, which is exactly for that purpose)(they just never will learn it, and I by now have the theory, well supported by evidence, that they play ill by purpose, to just not do what mavlink comes up with, their brains just seem to stop working properly on mavlink matter)

PR #41, opened 2023-01-29
Doc & Readme Update

jlpoltrack:
Removed Docs folder w/ Datasheet PDFs.  Added section to Readme on how to output .Hex in STM32CubeIDE.


olliw42:
@jlpoltrack 
many thx, sir
btw, are you Tracer21?

concenring the requested changes
- why should one remove the docs? I actually like them a lot, and I suspect others may do too ...
- why is it so important to create .hex files?

jlpoltrack:
Yes, that's me.  Notes back:

- Was planning to move the Datasheet PDFs to the Docs Repo so that can all be in one place.
- CubeIDE doesn't do this by default (at least for me), so I thought this would be helpful for new folks following the step by step instructions.  I added Hex since that is what you share in the /firmware folder.

olliw42:
ah, ok, I see the arguments

docs: I'm not fond of moving the docs to the doc repo. Language-wise it sure sounds silly, how can the docs not fit into the doc repo, but the doc repo is in fact not supposed to be cloned or downloaded but was a quicky for a webpage or wiki thing. It should just be for reading, so to say.
 
hex files: That's an interesting argument indeed. I guess I produce hex for the firmware folder only because they are text based and hence somewhat more git-ish, but not because they would be of any specific need or offer any advantage.  Not sure how to go along. The simplest would be to henceforth share the .bin files instead. Another approach would be to add a shorter sentence with regards to vcreatinbg hex in themain repo and a link to some more detailed text in the docs repo. (I'm trying to not bloat up the main README with too much info which is not soo important). What do you say?

jlpoltrack:
Fair enough.  Before I update this PR - do you have preferred spot in the Docs repo for the CubeIDE / Build notes?

olliw42:
@jlpoltrack 
not really, anything "aesthetically pleasing" is fine for me
I've added some chapter headers, preparing a spot, but feel free to suggest change if you feel so

jlpoltrack:
Closing.

Issue #42, opened 2023-02-12
Bug: rx Grove E5.


vrquaeler:
Hi Olli, 

I made some tests with Seeed hardware and found the Grove E5 module doesn't seem to work. 
The serial communication seems not working.

A small mod made it working. I will create a PR for a fix.

olliw42:
resolved & merged. THX!!!

PR #43, opened 2023-02-12
bug fix: E5Grove, swap UART

vrquaeler:
Hi Olli, 

Here is the bug fix for the Grove E5 module as receiver. This should fix issue #42. 
The Serial port has pointing to the wrong UART.

PS: I am the only one experimenting with this?

### DIFF
--- a/mLRS/Common/hal/rx-hal-wioe5-grove-wle5jc.h
+++ b/mLRS/Common/hal/rx-hal-wioe5-grove-wle5jc.h
@@ -48,7 +48,7 @@
 #define UARTB_USE_RX
 #define UARTB_RXBUFSIZE           RX_SERIAL_RXBUFSIZE
 
-#define UARTC_USE_UART2 // debug // also UART2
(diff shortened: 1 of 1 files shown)


olliw42:
MANY thx, sir!

> I am the only one experimenting with this?

I have one module and used it when I did the code, but I really haven't used it since then. I'm sure it was working, but I would not rule out that I screwed things up when I prepared the merge. 
So, currently, you indeed might be the only one :)

I will double check ASAP.

olliw42:
you are absolutely correct, is on PB6/7

you need to do the change also in line 51:  `#define UARTC_USE_UART1_REMAPPED // debug // also UART2`

vrquaeler:
Hi Olli, 

Thanks for the confirmation. 
Please feel free to update the PR. You are right, the debug port (Line 51) should be set to UART1, too.  
 I would like to update the the PR, but unfortunately I have screwed my "git" (again)....

olliw42:
MANY thx, sir, This is ver ymuch appreciated :)

PR #44, opened 2023-02-19
915 FCC Tweaks

jlpoltrack:
- Added 2 channels to 915 FCC frequency band (915.5, 926.9) bringing total to 40 channels (same as ELRS).
- Enabled 915 FCC | 31Hz to use 18 channels (same as 2.4 | 31Hz).

Tested on Wio E5 Tx+Rx.

### DIFF
--- a/mLRS/Common/common_conf.h
+++ b/mLRS/Common/common_conf.h
@@ -103,12 +103,12 @@
 #define MODE_31HZ_SEND_FRAME_TMO        15 // just needs to be larger than toa, not critical
 #define MODE_19HZ_SEND_FRAME_TMO        25 // just needs to be larger than toa, not critical
 
-#define FHSS_NUM_BAND_868_MHZ           	6 // it's a very narrow band
-#define FHSS_NUM_BAND_915_MHZ_FCC       	12 // was 24, but a cycle takes then 1.3 sec! would need long disconnect
-#define FHSS_NUM_BAND_915_MHZ_FCC_31HZ_MODE 18 // to match 2P4_31HZ
-#define FHSS_NUM_BAND_2P4_GHZ_19HZ_MODE 	12 // was 24, but a cycle takes then 1.3 sec! would need long disconnect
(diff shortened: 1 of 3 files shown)


olliw42:
@jlpoltrack 
interesting

the 915.5 must have been an omission

concerning the 926.9, I am not sure how the ends were actually determined. Just did some google and everywhere I found the fcc range 902 - 928 MHz, so it's not clear to me why the ends didn't had been extended to these limits (even not in ELRS)...
this is going to kill backwards compatibility, so I think we should reinvestigate what the actual range could/should be, to have a final proper range forever ...

concerning 18 vs 12 I see your rational, but I think the original idea was that it should be smaller since the available channels is smaller than in 2.4 GHz. Did you see any practical advantages of using 18 instead of 12? What would be the downsides of 18? What the downsides of 12?

olliw42:
@jlpoltrack 
I did think a bit more about this
since this is a breaking change I would like to suggest a 2-step procedure, i.e., to separate this PR into two PRs
- in PR#1 you could put in the changes which allow to handle different fhss no for the FCC. Here I would like you to rather change it though such that there is a 19HZ MODE definition and a "universal" definition, instead of vice versa as it is now (the idea is to make it consistent with how it is for the 2.4 GHz case). In this PR however the fhss no would for both modes be still 12
- in PR#2 you could add the additional frequencies (please wo comments, and blocked in blocks of 10), and the 18 instead of 12

This would allow us to merge PR#1 immediately, but merge PR#2 later, namely at a point there more breaking changes would have piled up (and the questions in the previous post would be settled). The idea is  to kind of "bundle" more breaking changes into one firmware, which I think would make it easier for users. (I'm thinking about some "breaking" changes in the parameter section).

jlpoltrack:
Sounds good.  I will close this PR and submit the first PR shortly with just the definition update.  I will take a couple of days to look at the regulations before submitting the second PR.

PR #46, opened 2023-02-21
915 FCC Update

jlpoltrack:
- Redid the 915 FCC channel list to have 43 channels which is the most that can fit with 0.6MHz spacing.
- Updated the number of hop channels to 25 for both 19Hz and 31Hz which is required per the regulations since mLRS uses the 500KHz LoRa bandwidth range:
image
https://www.ecfr.gov/current/title-47/chapter-I/subchapter-A/part-15/subpart-C/subject-group-ECFR2f2e5828339709e/section-15.247#p-15.247(a)(1)(i)

Also of note, 25 hop channels are limited to 250mW | 24dBm.  For >24dBm and <= 30dBm then 50 hop channels are needed:
image
https://www.ecfr.gov/current/title-47/chapter-I/subchapter-A/part-15/subpart-C/subject-group-ECFR2f2e5828339709e/section-15.247#p-15.247(b)(2)


olliw42:
@jlpoltrack 
many thx for this.
This is a massive change and I must admit I am not in a position to determine/decide if that is good or bad. It puts lots of questionmarks into my face but it's outside my range of competences to answer any.

I am worried about 25 channels at 19 Hz, as this is a fhss cycle time of 1.3 secs, and this determines the connection time (reconnection time in case of connection loss). In addition comes the time to see sufficiently many packets, which rasies the (re)connection time to up to 2 secs - and if there is lots of interference it can become much longer.
I realize that then I did these times I choose them kind of "aggressive" (also the time for disconnecting) and I kind of came to the conclusion over this last year that maybe one should be less aggressive, i.e., e.g. rather choose 1 sec for disconnecting, and make reconnection easier, but 2 secs for reconnecting I find quite long.

Note that mLRS works quite differently than ELRS concerning connection/reconnection (if I understand ELRS correctly). I believe that mLRS is superior in relevant regards, but mLRS' scheme has the downside that the fhss cycle time affects the connection/reconnection time.

I'm also a bit worried that - maybe - there was a good reason for ELRS to choose the channels as they did. I mean, we probably should start from assuming that they had a good reason, shouldn't we. So, I wonder.

In some sense I do not care, since I am in the EU, and this doesn't affect me, and happy to go with anyrthing, but I would want to avoid that in few months from now we again have to change it. That is, I would feel much more confident if these points and decissions would be based on a broader consensus.

I have no idea however how such a broader consenus could be reach. I hope you understand my hesitation.

A "safe" way to go would be to simply add another choice to the frequency option, like "915 Mhz FCC2". But ultimately, this needs some resolution.

jlpoltrack:
Certainly understand your hesitation and consider this a a discussion.

On the channel list, I view this as a small upgrade without downsides.  Would provide 5 more channels than mLRS today.  Could also eliminate the first and last channel for more buffer to the limits and still provide 3 more channels.

I also spent some time researching ELRS and found that originally ELRS only had 20 channels on Australia (where ELRS started) 915 bands (915 to 928MHz).

https://github.com/ExpressLRS/ExpressLRS/blob/bb62fb6eca85c786eb5aa31c7f4404b06e666d34/src/PlatformIO/LoRa_SX127x_Lib/src/FHSS.h#L27

My guess is that to expand to 915 FCC it was chosen to have 20 channels on the lower part that played nice with the existing list, this comment implies that there wasn't a particular reason.  This is also the channel list used today.

https://github.com/ExpressLRS/ExpressLRS/blob/1.0.x-maintenance/src/lib/FHSS/FHSS.h#L101

On the channel hops, maybe this could be a user configurable option instead of adding a frequency option?

jlpoltrack:
Sounds good, I will try out 19Hz | 25 hop channels and provide some notes on how it performs.  Does it make sense to spend time on 31Hz as well?  Sounds like it is potentially less of a concern there.

olliw42:
@jlpoltrack 
you may have seen that I pushed a correction to main, which corrects the "missing break" bug (also discussed in the 433/70 PR). This changes behavior for 915 FCC, and thus is a breaking change!!
This means, since for 915 FCC we just broke backwards compatibility anyhow, we also can consider adding the changes you've suggested in here.

There is probably little reason not to use the larger frequency range you've suggested.
As mentioned, I'm concerned however about the much larger hopping number. 
Did you had some chance to test?

Anyway, the main pruprose of this post is to juist mention that 915 FCC got broken and that this PR can thus get some renewed attention.

I guess, ideally we would ask "known" 915 FCC users about their opinions too ...

jlpoltrack:
@olliw42 
I haven't been able to do any testing outside of checking that the link works in the yard so can't provide any real notes on performance with 19 Hz / 25 hop channels.  Do you know if Brad is using 915?  I recall he was using Frsky hardware so he would be affected by this change.

olliw42:
@jlpoltrack 
you do not need to merge the main branch if there is no merge conflict :)

jlpoltrack:
I guess this create notifications for you - I do this so I can build new firmware from my fork with the latest 'main' code.  If there's a better way - let me know :)

olliw42:
ah, ok. There is probably a better way, but I wouldn't know from scratch LOL. So, just keep on. :)

jlpoltrack:
Flew a bunch on this update today and didn't see any bad behavior between link loss and link reacquisition on low power.  Link was reacquired at reasonable distances after loss.  Sensitivity was a bit poorer than before but can't see how this update would affect that (different airframe as well, so not 1:1 comparison).

olliw42:
great

it seems you have not updated to the latest master, in commit 7b164b7c6a8478c9dc83a103322fe771355481d2 I had changed the connect/disconnect time from 0.75 sec to 1.25 sec, I think this should help this FFC change too ... maybe you also have a chance to test this

I guess we want to also ask the person who was initially adding the FCC
seems it has happen in fd2402d177c867b3ed51cb252d280eb3dfd51f87
so would be again @Imen-D :)

@Imen-D, could we get feedback and possibly critical review from you on this PR?
It changes the 915 MHz FCC hopping quite a bit, in two ways: 1. more frequencies, 2. hopping is changed to 25 channels. Intention is to make it legal in the US.
In 19 Hz mode 25 channels mean that the hopping sequence takes 1.3 secs ... which did concern me a bit. Otherwise I don't have any stakes in this matter, so like you guys to decide :)

brad112358:
Yes, I use FCC.  I'm not currently very familiar with these details so I can't comment on the code changes, but it sounds important and I'm happy to help test it.  I'll merge it before I make my next flight and let you know if I see any trouble.

I'm curious what the "missing break" bug was.  I couldn't find it in the PR discussion.

I'm also unsure without investigation what the longer hop cycle time has to do with the disconnect time. 

Thanks.

jlpoltrack:
@brad112358 
Here's the 'missing break' fix that Olli found (discovered during 433/70 work): https://github.com/olliw42/mLRS/commit/89bfbf37f4f4296edd4e184625343257ad554c32

brad112358:
Thanks @jlpoltrack

olliw42:
> what the longer hop cycle time has to do with the disconnect time

as indicated I'm probably just irrationally hypocritical on this

fundamentally you are correct, the two are unrelated.
The idea circulating in my mind is that every frequency should have had a chance to send a packet, which would suggest that the disconnect time should be longer than the fhss cycle time. But I agree that there is no basic principle which would require that.
In fact, one could argue that the fhss cycle time is just existant for the type of fhss choosen here, while in e.g. fhss schemes with true random hops there would be no such thing, and the whole argument thus blunder.

I guess, all I'm saying is that it should have gotten enough testing by enough FCC users, so that we can call it stable and hopefully won't have to change it again anytime soon :)

olliw42:
@jlpoltrack @brad112358 @Imen-D 
any news on this. Is this experimenlly approved? Or are we still waiting for better weather to come :D

jlpoltrack:
No news to share, weather is stopping me from doing more testing.

Somewhat related, I was rereading some RCGroups threads and it turns out that certain Matek flight controllers generate noise on 433 and 915 which reduces sensitivity.  My 433 and 915 tests were with a Matek F405 Wing which may help to explain why I saw poor performance (especially with 433).  

@olliw42 Could you share which FC you use?  I recall you stating that you typically don't see much benefit on 868.

olliw42:
I do use two MatekH743, a "wing" type in my copter, and a "slim" type one on the bench, both of the first generations (do have them since a while)
however, I do not fly that far with my copter that I would reach connection loss
my range testes all are/were "bench" tests (not really on the bench but with bench test setups LOL)
I know about these stories, and there are similar stories with regards to that the Jumper T16/RadioMaster Tx16s radios emit lots of noise in these bands, but there are also other stories ...

brad112358:
I did find time to test this today with R9M and R9MX at their lowest power settings.  I merged this PR with Olli's main branch and did a walk test to confirm expected behavior, including reconnect after signal loss.  I now see 25 channels in the MAVLink for OpenTX telemetry debug screen.  Then, I flew it over a wooded area to the same range where I usually loose the connection because of the trees (about 600m), perhaps just a little farther than usual, before it reported loss of telemetry.  I was connected to QGC on my phone via BT and MBridge.  

The only new observation is that the MAVLink for OpenTX telemetry script was displaying the big box that says "is initializing" the entire time.  I'm sure this has nothing to do with this PR, but I haven't investigated to find the problem yet.

Looks good to me!

olliw42:
@brad112358 
fantastic. MANY thx for the tests and report.

@jlpoltrack @Imen-D 
should we merge?

jlpoltrack:
@olliw42  No concerns from me.

Imen-D:
@olliw42 , I'm in

olliw42:
MANY THX folks, and especially @jlpoltrack for working this out.

There will be a small follow up PR to remove the double defines of 25, but nfc.

PR #47, opened 2023-02-24
Trivial fixes for txbuf problems in Ardupilot and PX4 modes

brad112358:
Fix stuck in burst/normal pattern for Ardupilot;
The current code never considers bandwidth (always changes to burst mode) even for normal streams if stream rate is too high.
This causes lost messages and uncontrolled stream rate at 19Hz with the default settings in QGroundControl.  I was also able to reproduced the problem with Mission planner by setting the Telemetry rate in Config/Planner to Attitude=10, Position=10, all others = 2.
There are other ways to fix this;  Note that both versions of my original PR avoided this behavior in different ways.

Fix off by 1 error for PX4;   txbuf should be 50, not 51 to keep stream rates the same.

### DIFF
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -264,6 +264,7 @@ bool MavlinkBase::handle_txbuf_ardupilot(uint32_t tnow_ms)
     uint8_t txbuf_state_last = txbuf_state; // to track changes in txbuf_state
 
     if ((tnow_ms - radio_status_tlast_ms) >= 1000) {
+        txbuf_state_last = txbuf_state = TXBUF_STATE_NORMAL;
         radio_status_tlast_ms = tnow_ms;
         inject_radio_status = true;
     } else if ((txbuf_state == TXBUF_STATE_NORMAL) && (serial.bytes_available() > RX_SERIAL_RXBUFSIZE/2)) {
@@ -399,7 +400,7 @@ bool MavlinkBase::handle_txbuf_px4(uint32_t tnow_ms)
     } else if (rate_percentage < 80) {
         txbuf = 92;                       // ArduPilot: 91-95  -> -20 ms,    PX4: 51-100 -> *1.025
     } else {
-        txbuf = 51;                       // ArduPilot: 50-90  -> no change, PX4: 35-50  -> no change
+        txbuf = 50;                       // ArduPilot: 50-90  -> no change, PX4: 35-50  -> no change
     }
 
     if (txbuf_state == TXBUF_STATE_BURST) txbuf = 33; // just enough to stop parameter flow
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -400,7 +400,7 @@ bool MavlinkBase::handle_txbuf_px4(uint32_t tnow_ms)
     } else if (rate_percentage < 80) {
         txbuf = 92;                       // ArduPilot: 91-95  -> -20 ms,    PX4: 51-100 -> *1.025
     } else {
-        txbuf = 50;                       // ArduPilot: 50-90  -> no change, PX4: 35-50  -> no change
+        txbuf = 51;                       // ArduPilot: 50-90  -> no change, PX4: 35-50  -> no change
     }
 
     if (txbuf_state == TXBUF_STATE_BURST) txbuf = 33; // just enough to stop parameter flow


olliw42:
@brad112358 
many thx for this

but I'm honestly puzzled
I understand the problem you are refering to, and understand that your fix would easy it within some range, but it also comes at a cost for what all of it is intended to improve in the first place. :) It's a bit like one step forward followed by a step backwards.
I think I mentioned repeatedly the real and potential issues with non-param data and the dicothomy which is logically impossible to resolve just based on flow and filling (which made me suggest to e.g. spy for param), simply too many intrinsic contradictions.
This here seems to cut out what was identified as so important before, and ergo has also aspects which worsens. Also it can't fundamentally "fix" the too-large-rate issue since if one would cranck up the incomming rate it eventually will always break, so it "just" extends it to somehat larger configured data rate. I am frankly not sure I see how your code part did avoid that issue, in my experience it is also vulnerable.
I continue to argue that the whole mechanism needs revision, and not just repeated patchwork.

I think it should be just clear that one cannot squeeze 10000 B/s through that link and that therefore one shouldn't attempt to set the system up for what one could call obviously too much. I think the large majority of users will be very happy with that idea, and in that sense I would call what you try to "solve" a user error. :D :)
I think all we could ask for is that if the system works ok in steady state it can handle well the additional stream from param, mission, mavftp. And this it seems to do, within some range. :) 
Have you seen Tracer21's comments in the thread? At 19 Hz and 38400 even param mavftp works well (with AP4.4).

concerning the 50/51, yo ... could you do it as a separate PR, so I could merge that part?

thx again, I truely appreciate your efforts, and I hope that my hesitances do not stop you keeping on :)

brad112358:
I really didn't expect this much push back as it seems obvious to me that the existing code is broken.

As I said, without some change, the defaults in QGC cause very bad behavior because the default rates are just a little two high.  They have not been cranked up to unreasonable values.  If the user needs to carefully tune the rates, what is the point of setting txbuf based on bandwidth use in the first place?

You mention a cost.  Yes, in my testing, this change did slightly slow down parameter download because sometimes, the 1 second timer expires before buffered data drops back down to  FRAME_RX_PAYLOAD_LEN*3 bytes.  I believe this happens when Ardupilot sends several low rate streams at the same time (PX4 does not have this behavior)  This was the motivation for waiting up to 2 seconds in my code.  Reducing the difference between the high (1024) and low (FRAME_RX_PAYLOAD_LEN*3) water marks slightly might also make it work well enough without the 2 second wait.

The problem being addressed here happens when the data arrival rate of normal streams is greater than about 1024 bytes/sec more than the radio bandwidth in the last second.  Because of the bursty nature of Ardupilot, this can happen even when the average data arrival rate is perhaps only 50% or so more than we can handle which also means that simply switching from 50Hz to 30Hz or 19Hz mode without recon figuring stream rates might well trigger bad behavior.

In the original code I proposed in PR 40, this condition would cause txbuf=33 to be sent, and if that didn't slow things down enough, it would fairly quickly cause txbuf_overflow to be set to 2 and the code you found so confusing at line 283-286 would set txbuf=0 and slow down normal streams as quickly as possible.

In the final version of PR40, the recognition of this condition works differently and is a little slower, but it still works.  At line 145 when a second elapses without the buffer recovering to less than 1024 bytes, we will set txbuf_state to normal which will cause the normal bandwidth calculation to send txbuf=0.

Some of the following might be too obvious, but to ensure we are on the same page, let me sumarize what I believe are the most important ideas here:

1) We want to minimize latency cause by buffering too much data

2) As a practical matter, we want reliable and fast operation even with the bad assumptions MP and QGC and other GC software makes with respect to message loss.

3) Because of 2, we want to avoid buffer overflow and the resulting message loss

4) We want to utilize as much of our limited available bandwidth as we can, given the above constraints

5) I think it's useful to consider 4 types of message flows;
   a) "normal" periodic info update messages which can be configured to be sent at various rates.  If we loose one, the main impact is delayed information.
   b) regulated bulk data transfer messages which delay sending when a small txbuf is reported, currently just parameter download
   c) unregulated bulk data transfer messages like ftp which currently ignore txbuf.  Other than changing the autopilot, there is nothing we can do except do without these or try to make it work by careful settings of baud rates, etc to be low enough to make it work.
   d) Other messages such as forwarded messages from other components such as cameras and gimbals.  As you have said many times, new mechanisms are required to manage these.

6) Both Ardupilot and PX4 significantly reduce the rate of type a) flows when type b) is in progress which helps avoid overflow.  This also makes it easier to detect the difference between a) and b) especially for Ardupilot because txbuf=51 or more will completely stop b), allowing the buffer to drain.  If it doesn't drain in a reasonable time (1-2 seconds), we should assume b) is not the problem and fall back to our normal operation and reduce the flow rate of a) with txbuf=0.  But the current mLRS code for Ardupilot fails to do this.

7) I agree that it is difficult to guarantee optimal bandwidth use for a) and b), but I believe it should be possible to have good reliable behavior with a wide range of configured flow rates without resorting to explicitly checking message types.

8) The cleaver strategy you developed of sending periodic txbuf updates based on bandwidth goals rather than actual buffer levels works very well for a), but it doesn't react fast enough for b).  Simply detecting impending buffer overflow and sending an appropriate txbuf update immediately helps avoid buffer overflow and high latency but can cause other problems when trying to optimize link utilization.  The problem is that this clever strategy is never allowed to work in the current code if the configured stream rates are even 50% too high.

9) Allowing the buffer to overflow during parameter download can cause parameter download to completely fail because of 2) so we really do need more than the original periodic txbuf update.

10) I think we should prioritize solving any problems which might impact common use cases to avoid problems for new users of mLRS.

11) Your current txbuf code fails to avoid buffer overflow for type a) messages in a common use case which is the default settings for QGC and is also a possible problem when switching to a lower bandwith mode in mLRS after a user has optimized for 50Hz mode.  I really think it needs to be fixed in some way.

I think we can do better than what you have now for Ardupilot and I think this simple change help in the short term.  I will continue to experiment in the PX4 section to see if I can find a better more general approach.

olliw42:
hey Brad
I need to go with more time through some of your points, but some can be dropped immediately since we either agree or they are outdated

1: sure, but in fact no: this ONLY can/shall be of relevance in the steady state operation, but this is automatically realized if the steady state is controlled to not overflow the buffers. So, while we sure want that, it's really a no-op.

2: That's illusionary. All we can do is 3., but if the link has a packet loss it's lost. So, it'S a givena and a no-op.

3 & 4: yes. The order of importance is however 3, 3, 3, 4

5: exactly. Good sum up.

8: Long outdated. We have discussed and settled all that.

9: Corollary of 3.

10: The problem is that of the few current (known) users, most do not have a serious problem with how it is. For MissionPlanner and ArduPilot it works fine for them. It's largely one user who has serious concerns. I guess I need to check QGC and ArduPilot however myself. 

11: I agree that it would be nice to have a fix which prevents damage, but I do NOT agree that the simple fix suggested here is appropriate, since it really impeeds what one was trying to achieve. Maybe one should set a counter to ensure recovery. 

So, we are at 6 & 7, and 11.

What I generally don't like about the discussion is that it is presented in a way that "your method" would work flawlessly and would solved all issues. But this is simply not my experience. If this is what I would have found I would have jumped ship immediately. Both options break when I just do what I was my default doing all years before and now with 50 Hz. I find this discussion based on a false presumption tedious. Some facts seem not to be acceptable facts.

Also, ANY user can use your method and choose it!! So, it's a lot of discussion for an option which is already available.

I'm sorry it comes across as push back, that's not my intention. I just do have an opinion here.

brad112358:
Perhaps I should wait for you to read my comment more carefully, but you seem to be missing something...

>1: sure, but in fact no: this ONLY can/shall be of relevance in the steady state operation, but this is automatically realized if the steady state is controlled to not overflow the buffers. So, while we sure want that, it's really a no-op.

But it is NOT always automatically realized with your current code!  If the configured stream rates are just a little too high The buffers do overflow in the steady state and you never send anything except txbuf=51 which won't slow them down!  Have you missed this somehow? If so, I'm sorry I was not clear.  I hope you read more than the numbered points.

>2: That's illusionary. All we can do is 3., but if the link has a packet loss it's lost. So, it's a no-op.

Illusinonary?  It's a goal.   I guess in some way all goals are Illusianary.

>3 & 4: yes. The order of importance is however 3, 3, 3, 4

Yes, of course!  And the change I proposed negatively impacts 4 slightly and fixes a serious problem with 3.

>8: Long outdated. We have discussed and settled all that.

It's not outdated, it's still a problem we haven't perfectly solved and I still plan to work on it.  Sorry if it's obvious, I just wanted to restate the situation because you don't seem to see the problem here and I'm not sure what you've missed.

>9: Corollary of 3.
Again, sorry for stating the obvious.  I did warn that I might.

>11: I agree that it would be nice to have a fix which prevents damage, but I do NOT agree that the simple fix suggested here is appropriate, since it really impeeds what one was trying to achieve. Maybe one should set a counter to ensure recovery.

As I said, this is just a simple quick fix for a serious problem.  And, as I said, there are other ways to fix the problem.  What would you count?

>So, we are at 6 & 7, and 11.

>What I generally don't like about the discussion is that it is presented in a way that "your method" would work flawlessly and would solve all issues. But this is simply not my experience. If this is what I would find I would have jump ship immediately. Both options break when I just do what was my default doing all years before and now with 50 Hz. I find this discussion based on a false presumption tedious. Some facts seem not to be acceptable facts.

What in particular have I said that gave you the impression I think anything works flawlessly?  You've suggested I have such vanity previously as well and I'd like to know what I've said to suggest that.  In this PR I said there are other ways to solve the problem and the very last thing I said was that the PX4 section also needs improvement.

Also, I don't understand what you mean by some facts not being acceptable.  Hopefully, we can all accept facts as facts.

>Also, ANY user can use your method and choose it!! So, it's a lot of discussion for an option which is already available.

True, and I very much appreciate that!  But which method is default?  I'd also like the default method to work for Ardupilot and the default settings of QGC (and resonable configurations of Mission Planner, of course).

olliw42:
@brad112358 
sorry for having not responded further, I was busy with other mLRS stuff (433 MHz mainly :))

I also did another attempt at ardu_1, it's in this branch https://github.com/olliw42/mLRS/tree/dev-ardu_2

At first it may look very similar to your attempt, but it is substantially different in some details. It actually follows a simple logic, to attempt distingusihing high flow from params, which could be phrased as "first try to cut off PARAMs, and if this is not scucessfull slow down". I found it to be quite effective in preventing buf overflow. Even with stream rates like Att 10, Pos 10 Mode 4, RC 4, Sens 4 (or even higher, 19 Hz of course).
I didn't observed such a case, but one could/should add a txbuf_state = TXBUF_STATE_NORMAL; in the >1000 case.
What it is not yet so good at is to quickly speed up the stream in normal mode.

Curious to hear your thoughts & experience with ardu2. (maybe in another thread).

brad112358:
Hi @olliw42 

I'm back from my break and though I haven't yet tested your new code in handle_txbuf_ardupilot(), I did look it over and I have some initial thoughts I'll share here.

> I didn't observe such a case, but one maybe should add a txbuf_state = TXBUF_STATE_NORMAL; in the >1000 case

I'm not sure what you are referring to here.  Does this comment still apply or was it for an older revision than what is currently in main?

Overall, I think your approach should work very well.  It has the significant advantage of speeding up convergence when the initial rates are way too high or when a burst of forwarded traffic threatens to overflow the buffer.

I do have a few suggestions for potential improvement:

The rate limit is good in general, but I think the block for serial.bytes_available() > 1024 in the TXBUF_STATE_NORMAL case should be moved outside of the if ((tnow_ms - radio_status_tlast_ms) >= 100) rate limit block;  If the buffer is filling up beyond 1024 because of parameter transfer, it is best to stop it immediately rather than potentially wait up to 100ms.  This is most important at high baud rates, but no reason to wait even at lower baud rates.  For example at 57600 baud, a 100ms delay in detecting > 1024 might mean we actually have 1024+576 bytes in the buffer by the time we send txbuf=50 (less if it's parameter traffic which is limited to 1/3 now?).  At 230400 baud, with the latest ardupilot development branch, we might have 1024+768 bytes in the buffer before we send txbuf=50 and with the current released branch, I suspect we might overflow our buffer.

I see that you have raised lower threashold level to fall back to normal to 384.  I'm not sure where this number comes from, but I suspect this change will eliminate the need for the 2 second delay I have in handle_txbuf_px4().  I was planning to experiment with lowering the upper threashold from 1024 to about 800 for the px4 code which I think would have a similar effect.

Because of Ardupilot's bursty traffic for the low rate streams (I think you noted 500-600 byte bursts), I suggest bumping the threashold for entering TXBUF_STATE_BURST_HIGH from 1024 to about 1536.  This should help avoid sending an extra txbuf=0 when these bursts come in during parameter download.  Depending on timing, the current threashold might cause unneeded slowing of normal streams when these bursts happen in TXBUF_STATE_BURST state.  On the other hand, slowing down normal streams more than needed only causes a slight delay in resuming normal rates later which I don't see as a major problem, but you did mention it as a concern earlier.   

In case this isn't clear, let me give an example.  Suppose a burst of 550 bytes of normal stream messages comes in shortly after we have stopped parameter flow because we reached the 1024 byte buffer level.  Then, 100ms later 82 or 164 bytes will have been sent over the radio link and the buffer will still contain at least 1024+550-164 or 1410 bytes and we will send txbuf=0.  Then, 100ms after that, the buffer will contain at least 1410-164=1246 bytes and we send txbuf=0 again.  Then we have at least 1246-164=1082 bytes and we send txbuf=0 a third time.  By raising the threshold for TXBUF_STATE_BURST_HIGH to 1536, we would avoid significantly reducing the rate of normal streams when we didn't need too.  But, when the rate of normal streams is way too high, we will still enter TXBUF_STATE_BURST_HIGH and quickly reduce the rate of normal streams.

What do you think of these ideas?

I'll try to find time soon to test the existing code and I may also try my suggested improvements.

olliw42:
@brad112358 

thx for the detailed response.

I sure agree that it might be tuned better in some aspects. What I would consider most important however is that the "catastrophic" losses should not happen anymore (or very very rarely)(which was the purpose of this PR).

few comments

230400 : I'm not nearly as interested in this as you seem, and in fact I would not consider this a use case worth considering as a relevant use case. I rather would consider removing this option in fact. Remind you, I added this to help you in some early stage, but these days are long gone. It's maybe interesting as a stress test, but that's it. 115200 is really plenty, and sooooo much better than 9600 which other systems suggest.

100 ms: I understand your point, also since I pondered over this myself. However, I didn't liked your suggestion since it can then happen that a stroke aof many messages are injected, and this to me doesn't make sense. So one would need some state machine to prvent that. Also, in none of my practical tests I have seen the theoretical case you describe to occur. I think this is because ArduPilot throttles (now) params and mavftp to 1/3 of the baudrate (you have seen my PR corecting this for params?). i considered 50 ms, but found 100 ms to be good enough. But this number can be tuned if evidence suggests.

1536: I don't like this and didn't liked before, hence I went down to 1024. I find the overhead of just 512 bytes too small. In simple words, I think teh best apporach to prevent buffer overflow is to pretend that the buffer is much smaller than it actually is. Note that this also helps with the 230400&100ms thing, which just 512 instead of 1024 overhead it's getting more difficult on these fronts too. Yes, it will more often jump to high burst than in an optimal scenario woudl be needed, but it makes it safe and robust.

384: pulled out of my hair. I considered the 184 it was before just way too small, we are happy with 1024 (or in your case ven 1536), so there is really no need to let it drop to that low filling. But this number can be tuned if evidence suggests.

Btw, the highburst/txbuf = 0 happens surprisingly rarely, at least for me, so optimizing to save one or two is not very substantial, aslo because if it is not needed it is corrected for 100 ms later.

But let's see your tests. As the saying goes: the proof is in the taste of the pudding LOL


>>  I didn't observe such a case, but one maybe should add a txbuf_state = TXBUF_STATE_NORMAL; in the >1000 case

> I'm not sure what you are referring to here. Does this comment still apply or was it for an older revision than what is currently in main?

it's in refernce to main

brad112358:
>However, I didn't liked your suggestion since it can then happen that a stroke of many messages are injected, and this to me doesn't make sense.  So one would need some state machine to prevent that

Yes, that would be bad, bu iit can't happen because you already have a state machine and the first time it happens, you change states so it can't repeat for at least 100ms.

brad112358:
I'm interested in higher baud rates for 2 reasons.  It's a good stress test to avoid the "help" Ardupilot tries to provide in limiting rates to 1/3 and to reduce transmission latency.

> Most importantly, in none of my practical tests I have seen the theoretical case you describe to occur

It won't occur often, so you may not have seen it yet.  This is because the bursts happen infrequently during parameter download.  Perhaps based on this it's not worth avoiding?

olliw42:
> > However, I didn't liked your suggestion since it can then happen that a stroke of many messages are injected, and this to me doesn't make sense.  So one would need some state machine to prevent that
> 
> Yes, that would be bad, but it can't happen because you already have a state machine and the first time it happens, you change states so it can't repeat for at least 100ms (actually longer because the buffer has to drain first)

I guess I then don't unerstand the change you want to make :)

PR #49, opened 2023-02-24
rx-E5-mini LED swap similar to tx-E5-mini

vrquaeler:
Hi Olli, 

I fell in love with your _new_ idea of the LED assembly for the tx-E5-mini. 
Mainly for debugging purposes I have to flash a "tx-E5-mini" with rx-firmware. The presence of a green LED or a real bind-button can help a lot (The rx-Gove-E5 is very limited). 
To be consistent (and interchangeable for debugging) I want to do the same trick here. 

Hopefully this don't break other existing assemblies. 
Let me know what you think.

Regards, Sascha

### DIFF
--- a/mLRS/Common/hal/rx-hal-wioe5-mini-wle5jc.h
+++ b/mLRS/Common/hal/rx-hal-wioe5-mini-wle5jc.h
@@ -6,6 +6,7 @@
 //*******************************************************
 // hal
 //*******************************************************
+// 24.Feb.2023: LED pin changed! pin for artificial GND (PA0)! consistent with tx-wioe5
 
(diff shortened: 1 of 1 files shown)


vrquaeler:
@olliw42
If I remember correctly, the user was **Tracer21** and he did not care about the green LED.

olliw42:
@jlpoltrack 
vrqualer suggests to modify the target for using a WioE5 mini dev as rx module, and to move the pin for the green LED like it was moved for the tx target. Ok with you?

jlpoltrack:
All good with me.

olliw42:
many thx guys
so I'll merge

PR #51, opened 2023-02-27
433 MHz Initial Work

jlpoltrack:
Don't expect this to get merged, just wanted feedback on if I am missing anything to enable 433 MHz support.


olliw42:
very cool

two quick thoughts

1. the 433 frequencies I had added were for EU. These here seem to be for US. I would not want the EU frequencies be removed/erased/replaced. So, the new US ones should be added in addition. Would it make sense to call them 433MHZ_FCC, like for 915?

2. It seems to me that this PR has also the modified 915_FCC stuff in it. If so, then please separate this out; the PR should only have the changes required for 433 in comparison to master.

> I am missing anything to enable 433 MHz support.

there at least needs to be a corresponding change in setup_tx.h, https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/setup_tx.h#L66. Also here, https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/hal.h#L294-L298. There might be more places I don't realize now. If you search for FREQUENCY_BAND_ and similar you should catch them all.

olliw42:
@jlpoltrack 
there are actually probably a number of changes needed, e.g., also in the lua script
if you want, you could PR the 433 MHz stuff for as much as you see it, and I could then work out the rest

olliw42:
ah, that's why there could be so many 433 channels LOL

what licence would this be?
I mean, if it has a somewhat unique shorthand, one could add it as 433MHZ_XYZYZY, and just not make it available via the settings.

jlpoltrack:
It's the US Amateur Radio (HAM) technician license.  Maybe 433_HAM for a name?  There are also geo restrictions, so 430 to 450MHz  is a 'safer' choice even though 420 to 450 MHz is allowed.  1 Watt is max power for radio control.

FYI - this is what Dragon Link requires to use.

https://www.ecfr.gov/current/title-47/chapter-I/subchapter-D/part-97/subpart-C/section-97.215

https://www.arrl.org/files/file/Tech%20Band%20Chart/US%20Amateur%20Radio%20Technician%20Privileges.pdf

One specific note on 5 frequencies - okay to expand this bitmask?  Currently only 4 bits: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/setup.h#L32

olliw42:
so let's try FHSS_NUM_BAND_70_CM_HAM, and see how it goes down LOL will raise many eyebrows I guess

PR #52, opened 2023-02-27
433 initial work

jlpoltrack:
Work for EU 433 & US 70 CM HAM frequency bands.


mLRS/Common/common_conf.h
```diff
@@ -103,6 +103,10 @@
 #define MODE_31HZ_SEND_FRAME_TMO        15 // just needs to be larger than toa, not critical
 #define MODE_19HZ_SEND_FRAME_TMO        25 // just needs to be larger than toa, not critical
 
+#define FHSS_NUM_BAND_433_MHZ               3 // there are only 3
```
olliw42, line 106: pl swap with _19HZ_MODE, so we have equal sequence everywhere

ALSO: can be only 2 (one is needed for bind, and by current design bind cannot be hopped on)(I consider using one of the "bad" frequencies for bind, so that we can have 3 for hopping, but that's futur LOL)

mLRS/Common/fhss.h
```diff
@@ -56,8 +56,53 @@ const uint32_t fhss_freq_list_433[] = {
     SX12XX_FREQ_MHZ_TO_REG(433.480),
 };
 
-const uint8_t fhss_bind_channel_list_433[] = {
-    0, // just pick some
+#ifdef FHSS_HAS_CONFIG__70_CM_HAM
```
olliw42, line 59: one '_' to many

mLRS/Common/fhss.h
```diff
@@ -56,8 +56,53 @@ const uint32_t fhss_freq_list_433[] = {
...
+// USA Ham Bands w/ Tech License
+// https://www.arrl.org/files/file/Tech%20Band%20Chart/US%20Amateur%20Radio%20Technician%20Privileges.pdf
+// 33 Channels w/ 0.6 MHz spacing
+
+const uint32_t fhss_freq_list_70_CM_HAM[] = {
```
olliw42, line 64: here it should be all lower case, _70_cm_ham

mLRS/Common/fhss.h
```diff
@@ -280,7 +326,21 @@ const tFhssConfig fhss_config[] = {
...
-    { Xhkahsdkhasd }, // to force an error
+    {
+        .freq_list = fhss_freq_list_433,
+        .freq_list_len = (uint8_t)(sizeof(fhss_freq_list_433) / sizeof(uint32_t)),
+        .bind_channel_list = fhss_bind_channel_list_433,
+        .bind_channel_list_len = (uint8_t)(sizeof(fhss_bind_channel_list_433) / sizeof(uint8_t))
+    },
+#else
+    { .freq_list = nullptr },
+#ifdef FHSS_HAS_CONFIG_BAND__70_CM_HAM
```
olliw42, line 337: dito, '_'

mLRS/CommonTx/setup_tx.h
```diff
@@ -51,7 +51,7 @@
 #define SETUP_OPT_RX_SERIAL_BAUDRATE  "9600,19200,38400,57600,115200,230400"
 #define SETUP_OPT_SERIAL_LINK_MODE    "transp.,mavlink"
 
-#define SETUP_OPT_RF_BAND_LONGSTR     "2.4 GHz,915 MHz FCC,868 MHz" // this is used e.g. in cli
+#define SETUP_OPT_RF_BAND_LONGSTR     "2.4 GHz,915 MHz FCC,868 MHz, 433 MHz, 70 CM HAM" // this is used e.g. in cli
```
olliw42, line 54: no blanks before/after ','
70 cm HAM

the problem here is that only 44 char are allowed max ... I guess we need to shorten to 433 and 70 ... in the lua they are faked anyway, and I guess one need to do also in the CLI
so, keep it for teh moment, but you will see me doing something

mLRS/Common/setup.h
```diff
@@ -30,13 +30,17 @@ void setup_configure_metadata(void)
...
 #elif defined FREQUENCY_BAND_915_MHZ_FCC
-    SetupMetaData.FrequencyBand_allowed_mask = 0b0010; // only 915 MHz FCC, not editable
+    SetupMetaData.FrequencyBand_allowed_mask = 0b00010; // only 915 MHz FCC, not editable
 #elif defined FREQUENCY_BAND_868_MHZ
-    SetupMetaData.FrequencyBand_allowed_mask = 0b0100; // only 868 MHz, not editable
+    SetupMetaData.FrequencyBand_allowed_mask = 0b00100; // only 868 MHz, not editable
+#elif defined FREQUENCY_BAND_433_MHZ
+    SetupMetaData.FrequencyBand_allowed_mask = 0b01000; // only 433 MHz, not editable
+#elif defined FREQUENCY_BAND_70_CM_HAM
+    SetupMetaData.FrequencyBand_allowed_mask = 0b10000; // only 70_CM HAM, not editable
```
olliw42, line 43: only 70 cm HAM

mLRS/Common/fhss.h
```diff
@@ -389,7 +450,8 @@ class FhssBase
         case FHSS_CONFIG_2P4_GHZ: return SETUP_FREQUENCY_BAND_2P4_GHZ;
         case FHSS_CONFIG_915_MHZ_FCC: return SETUP_FREQUENCY_BAND_915_MHZ_FCC;
         case FHSS_CONFIG_868_MHZ: return SETUP_FREQUENCY_BAND_868_MHZ;
-        // case FHSS_CONFIG_433_MHZ: return SETUP_FREQUENCY_BAND_433_MHZ; // is not yet existing !!
+        case FHSS_CONFIG_433_MHZ: return SETUP_FREQUENCY_BAND_433_MHZ; // exists
+        case FHSS_CONFIG_433_MHZ: return SETUP_FREQUENCY_BAND_70_CM_HAM; // exists
```
olliw42, line 454: probably not :D

lua/mLRS.lua
```diff
@@ -431,6 +431,8 @@ local freq_band_list = {}
 freq_band_list[0] = "2.4 GHz"
 freq_band_list[1] = "915 MHz FCC"
 freq_band_list[2] = "868 MHz"
+freq_band_list[3] = "433 MHz"
+freq_band_list[4] = "70 CM HAM"
```
olliw42, line 435: 70 cm HAM

olliw42:
@jlpoltrack 
if I want to test this, I think I can mis-use my E22-900 modules, right, or would you see a reason it would not work (not talking about suboptimal RF performance with wrong antenna and RF backends)

mLRS/Common/fhss.h
```diff
@@ -56,8 +56,53 @@ const uint32_t fhss_freq_list_433[] = {
     SX12XX_FREQ_MHZ_TO_REG(433.480),
 };
```
olliw42, line 58: #endif if missing

mLRS/Common/common_conf.h
jlpoltrack: Order swapped, updated 2 channels only.

mLRS/Common/fhss.h
jlpoltrack: Grr, fixed.

mLRS/Common/fhss.h
jlpoltrack: Done.

mLRS/Common/fhss.h
jlpoltrack: Fixed.

mLRS/Common/setup.h
```diff
@@ -30,13 +30,17 @@ void setup_configure_metadata(void)
 
     //-- FrequencyBand: "2.4,915 FCC,868"
```
olliw42, line 31: adapt this too

mLRS/CommonTx/setup_tx.h
jlpoltrack: Updated to '2.4 GHz,915 MHz FCC,868 MHz,433,70'

mLRS/Common/setup.h
jlpoltrack: Comment updated.

mLRS/Common/fhss.h
jlpoltrack: :)

lua/mLRS.lua
jlpoltrack: String updated.

mLRS/Common/fhss.h
jlpoltrack: Good catch, fixed.

mLRS/Common/setup.h
jlpoltrack: Comment updated.

jlpoltrack:
Re:
if I want to test this, I think I can mis-use my E22-900 modules, right, or would you see a reason it would not work (not talking about suboptimal RF performance with wrong antenna and RF backends)

As far as I can see the SX1268 which E22-400 uses doesn't support lower frequencies (so it is probably cheaper):

SX 1268 | 410 to 810 MHz:

image

SX1262 | 150 to 960 MHz:

image

jlpoltrack:
> btw, if you want to put down a reference/cite, just add a '>' in front, like '> kasfhashfdkuh' gives

Good tip :)

olliw42:
pl go through mLRS/Common/fhss.h again, it still has a number of 70_CM_HAM vs 70_cm_ham issues
(and it is probably not compiling as is)

in fhss.h, at around line 32, you probably want to add something handling 70

jlpoltrack:
Upper / lower fixes, definition added on 32, ifndef fix.  Does compile for me with new freqs enabled: 
image

mLRS/Common/setup.h
```diff
@@ -28,15 +28,19 @@ void setup_configure_metadata(void)
...
 #elif defined FREQUENCY_BAND_915_MHZ_FCC && defined FREQUENCY_BAND_868_MHZ
-    SetupMetaData.FrequencyBand_allowed_mask = 0b0110; // only 915 FCC, 868
+    SetupMetaData.FrequencyBand_allowed_mask = 0b00110; // only 915 FCC, 868
 #elif defined FREQUENCY_BAND_915_MHZ_FCC
-    SetupMetaData.FrequencyBand_allowed_mask = 0b0010; // only 915 MHz FCC, not editable
+    SetupMetaData.FrequencyBand_allowed_mask = 0b00010; // only 915 MHz FCC, not editable
 #elif defined FREQUENCY_BAND_868_MHZ
-    SetupMetaData.FrequencyBand_allowed_mask = 0b0100; // only 868 MHz, not editable
+    SetupMetaData.FrequencyBand_allowed_mask = 0b00100; // only 868 MHz, not editable
+#elif defined FREQUENCY_BAND_433_MHZ
```
olliw42, line 40: do we want the option to have both 433 and 70?
could it not even make sense to allow the option of having all four, 868,915,433,70?
e.g., on diy modules one may want to have that, so that one doesn't need a new firmware for a different E22
keep it as is, I'm just curious about your thoughts

mLRS/Common/setup.h
```diff
@@ -398,6 +406,20 @@ void setup_configure(void)
         break;
     default:
         while (1) {} // must not happen, should have been resolved in setup_sanitize()
+    case SETUP_FREQUENCY_BAND_433_MHZ:
```
olliw42, line 409: not so nice, this all should come before the final default:
pl swap

olliw42:
@jlpoltrack 
with the suggested changes, I tried it on my system, but unfortunately it crashes it when I select e.g. 70 ... I guess the 900 modules don't like being used at 400 ...

so, we can't do as I first suggested, i.e., that I PR and then add on top what it needs else ... we need to rely on you to test this PR

mLRS/Common/fhss.h
```diff
@@ -298,7 +367,8 @@ class FhssBase
         case SETUP_FREQUENCY_BAND_2P4_GHZ: config_i = FHSS_CONFIG_2P4_GHZ; break;
         case SETUP_FREQUENCY_BAND_915_MHZ_FCC: config_i = FHSS_CONFIG_915_MHZ_FCC; break;
         case SETUP_FREQUENCY_BAND_868_MHZ: config_i = FHSS_CONFIG_868_MHZ; break;
-        // case SETUP_FREQUENCY_BAND_433_MHZ: config_i = FHSS_CONFIG_433_MHZ; break; // is not yet existing !!
+        case SETUP_FREQUENCY_BAND_433_MHZ: config_i = FHSS_CONFIG_433_MHZ; break; // exists
```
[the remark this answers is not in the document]
jlpoltrack: Are you saying this should be:
```suggestion
        case SETUP_FREQUENCY_433_MHZ: config_i = FHSS_CONFIG_433_MHZ; break;
```
I'm a bit confused as the other lines have band in them and in setup_types.h I have:
 'SETUP_FREQUENCY_BAND_433_MHZ,' 

Will look at the others tomorrow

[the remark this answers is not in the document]
olliw42: oh, I am sorery, you are right ... I made the comment for the wrong line
I close and redo

mLRS/Common/fhss.h
```diff
@@ -280,7 +334,22 @@ const tFhssConfig fhss_config[] = {
...
+    {
+        .freq_list = fhss_freq_list_433,
+        .freq_list_len = (uint8_t)(sizeof(fhss_freq_list_433) / sizeof(uint32_t)),
+        .bind_channel_list = fhss_bind_channel_list_433,
+        .bind_channel_list_len = (uint8_t)(sizeof(fhss_bind_channel_list_433) / sizeof(uint8_t))
+    },
+#else
+    { .freq_list = nullptr },
+#endif
+#ifdef FHSS_HAS_CONFIG_BAND_70_CM_HAM
```
olliw42, line 346: dito

mLRS/Common/fhss.h
```diff
@@ -280,7 +334,22 @@ const tFhssConfig fhss_config[] = {
     { .freq_list = nullptr },
 #endif
 #ifdef FHSS_HAS_CONFIG_BAND_433_MHZ
```
olliw42, line 336: no `_BAND_` here

olliw42:
@jlpoltrack 
have you "allow changes by maintainers" enabled?
I tried to push changes to your pr, but it complains that I don't have the access rights

mLRS/Common/fhss.h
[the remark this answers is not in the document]
jlpoltrack: No problem, I will mark this one as resolved.

mLRS/CommonTx/setup_tx.h
```diff
@@ -51,7 +51,7 @@
 #define SETUP_OPT_RX_SERIAL_BAUDRATE  "9600,19200,38400,57600,115200,230400"
 #define SETUP_OPT_SERIAL_LINK_MODE    "transp.,mavlink"
 
-#define SETUP_OPT_RF_BAND_LONGSTR     "2.4 GHz,915 MHz FCC,868 MHz" // this is used e.g. in cli
+#define SETUP_OPT_RF_BAND_LONGSTR     "2.4 GHz,915 MHz FCC,868 MHz,433,70" // this is used e.g. in cli
```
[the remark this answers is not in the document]
jlpoltrack: Made these two updates.

mLRS/Common/fhss.h
```diff
@@ -57,7 +60,57 @@ const uint32_t fhss_freq_list_433[] = {
...
+    SX12XX_FREQ_MHZ_TO_REG(446.6),
+    SX12XX_FREQ_MHZ_TO_REG(447.2),
+    SX12XX_FREQ_MHZ_TO_REG(447.8),
+
+    SX12XX_FREQ_MHZ_TO_REG(448.4),
+    SX12XX_FREQ_MHZ_TO_REG(449.0),
+    SX12XX_FREQ_MHZ_TO_REG(449.6),
+};
+
+
```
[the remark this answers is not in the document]
jlpoltrack: Line deleted.

mLRS/Common/fhss.h
```diff
@@ -389,7 +459,8 @@ class FhssBase
         case FHSS_CONFIG_2P4_GHZ: return SETUP_FREQUENCY_BAND_2P4_GHZ;
         case FHSS_CONFIG_915_MHZ_FCC: return SETUP_FREQUENCY_BAND_915_MHZ_FCC;
         case FHSS_CONFIG_868_MHZ: return SETUP_FREQUENCY_BAND_868_MHZ;
-        // case FHSS_CONFIG_433_MHZ: return SETUP_FREQUENCY_BAND_433_MHZ; // is not yet existing !!
+        case FHSS_CONFIG_433_MHZ: return SETUP_FREQUENCY_BAND_433_MHZ; // exists?
```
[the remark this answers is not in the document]
jlpoltrack: Comment removed.

mLRS/Common/setup.h
```diff
@@ -28,15 +28,19 @@ void setup_configure_metadata(void)
 {
     SetupMetaData = {0};
 
-    //-- FrequencyBand: "2.4,915 FCC,868"
+    //-- FrequencyBand: "2.4,915 FCC,868,433,70 cm"
```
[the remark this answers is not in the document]
jlpoltrack: ' cm' removed.

mLRS/Common/setup.h
jlpoltrack: Not a bad idea to have both 433 & 70.  Have added this.

mLRS/Common/setup.h
jlpoltrack: Updated.

mLRS/Common/fhss.h
jlpoltrack: Now 'FHSS_HAS_CONFIG_70_CM_HAM'

mLRS/Common/fhss.h
jlpoltrack: Now 'FHSS_HAS_CONFIG_433_MHZ'

geofrancis:
Could you expand this to include 458.5 to 459.5 so its legal in the UK? 

https://www.largemodelassociation.com/resources/legal-model-frequencies-in-uk/

olliw42:
>  Looks like it (FYI, I haven't changed this setting):

something mist be bad on my side ... now I cannot even pull anymore without a access denied error ... git/github sometimes is simply really a pain ... I guess I need to do it stupidly and download LOL ... grrrr

jlpoltrack:
@geofrancis 
>Could you expand this to include 458.5 to 459.5 so its legal in the UK?

Can look at after getting 433 and 70 cm working on E22.  Quick note is that 1 MHz is awfully small (especially as mLRS uses 500 KHz bandwidth).  2 channels with 1 for bind and 1 for connection?

olliw42:
@jlpoltrack 
my fault, I should have more specific
you need to move the files to the subfolders where they belong to
massive sorry for the inconvennience

jlpoltrack:
Alright, I did this (and put the files in the correct folders - all good)

> this is super nasty, I'm sorry, but as said I can't ven pull anymore

Did your SSH expire?

> as regards why 400/70 didn't work for me. One for sure will need to do some code changes in the driver in order to activate the LF instead of the HF part.

I think this is SX127x specific?  Continuing to look at SX126x datasheet.

olliw42:
great!!!

I think I can't see anything anymore. I had this tested in as much as I could see the options in the cli, see them in the lua, and change them thorugh the lua (to get the modules crashing).

not sure how to proceed from here. As is it is probably not functional, since it may need some changes in the sx driver, or where ever else. 
I could merge, based on that it won't do anything fior the existing targets. On the other hand, merging code which hasn't been demonstrated to work isn't the textbook approach.
What would be your suggestion?

olliw42:
> Did your SSH expire?

I'm not using SSH ... SSH is way above my paygrade ;)

> I think this is SX127x specific? Continuing to look at SX126x datasheet.

you are probably be correct. Just pulled something out of my head. Good thing is you have some hardware to test :)

jlpoltrack:
How about this:

- Could you provide some hints on how to create a Tx target for G441 + E22?  I think it makes sense to see if I can get the tx working without crashing before doing anything else.  Crashing means CLI / LUA becomes unresponsive?
- If that works, I need to build another copy of the hardware (have parts just need to solder) to test out rx side

olliw42:
it's also a g441kb, like the rx ones? 8 MHz ceramic crystal likes these, right?

olliw42:
what I could do is to make tx target of the rx-diy-e22-g441kb, as I have that hardware and so could even test
from here it should be asy to dapt the hal, I would think
(I actually wonder if I should do a g431kb ... I believe 441 or 431 should not matter - unless that extra pheripheral is needed - would make it mroe general/generic)

jlpoltrack:
Tx target works for me. 

> (I actually wonder if I should do a g431kb ... I believe 441 or 431 should not matter - unless that extra pheripheral is needed - would make it mroe general/generic)

Do you mean G431 code could be used on G441?  If so, that makes sense (mLRS doesn't use the hardware AES).  I guess at some point there will be hardware standardization...

jlpoltrack:
I'm looking at other LoRa libraries and one thing that I see done is 'Image Calibration' Section 9.2.1.  For SX1262 it is default for 902-928 MHz (maybe 868 is close enough but 430 too far away)?

olliw42:
@jlpoltrack 
you should find a tx E22 G431KB target now in main
I in fact didn't had such a board with E22 but only the E28 variant of it, which is hardware wise however equal except of E28 vs E22, and it works even though it has a g441kb soldered in
should hopefully serve you as template for your board

olliw42:
it does crash also if you don't do any sx stuff?
I too had the "same" sympthom, not even leds, so I wonder, may it gets stuck somewhere else (you may have noticed that there are a number of while(1){}, each can be a reason), and not in the sx section
I guess I would first try to establish that the leds can blink, and that the dbg uart works, once you have a dbg uart, you can add puts and see there it crashes

jlpoltrack:
I can blink the LEDs with other code setup to use the clock config from the project, so I think that at least works.

Any steps for enabling debug other than:

Comment out this block: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/tx-hal-diy-e22-g431kb.h#L19
Set UARTC to UARTF: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/tx-hal-diy-e22-g431kb.h#L50

I guess it gets stuck early on, even if I comment out both sx lines in mlrs-tx.cpp I still get no LEDs. https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/mlrs-tx.cpp#L146

jlpoltrack:
Well the hardware works :) but I don't have any leads as to why the 433 code doesn't work.  I was rude and put the 70 cm HAM freqs into the 868 define for the current master (9ca32bd) and setup the device to only use 868:

433 mLRS

olliw42:
yeah, it seems we have a simple bug somewehere ... I came to this concuion in the meantime too ... will look at it

jlpoltrack:
Really appreciate all your help on this, I will continue to poke around and let you know if I find anything.  Well not exactly 1 Watt - maybe I need to get an attenuator for the RF meter (or complain to eByte):

IMG_1202

jlpoltrack:
Is there some threshold for minimum number of channels elsewhere?

- 3 channels of 433 I get stuck - red LED only  
- 6 channels of 433 I get stuck - red LED only
- 10 channels (same as 868) works okay

olliw42:
the accuracy of this meter is just 1dBm or so, and it's easy to get insertion losses of 0.5-1dB ... a 29 would have been nice though LOL

jlpoltrack:
Built up another board for receiver and was able to bind!  Couldn't figure out how to get the out working so just disabled.  Close and yet far away :)  Haven't measured power consumption but looks like the gain of the amplifier is something like 10-12db so you can run really low power while testing.

image

mLRS/Common/setup.h
```diff
@@ -396,6 +411,20 @@ void setup_configure(void)
     case SETUP_FREQUENCY_BAND_868_MHZ:
```
olliw42, line 411: BUG BUG BUG
`break;` is missing here !!!!

(this is also in master, and I wonder how 915 MHZ FCC ever could work !!)

mLRS/Common/setup.h
```diff
@@ -396,6 +411,20 @@ void setup_configure(void)
...
         Config.FhssNum = FHSS_NUM_BAND_868_MHZ;
         break;
+    case SETUP_FREQUENCY_BAND_433_MHZ:
+        switch (Config.Mode) {
+        case MODE_31HZ: Config.FhssNum = FHSS_NUM_BAND_433_MHZ; break;
+        case MODE_19HZ: Config.FhssNum = FHSS_NUM_BAND_433_MHZ_19HZ_MODE; break;
+        default:
+            while (1) {} // must not happen, should have been resolved in setup_sanitize()
+        }
+    case SETUP_FREQUENCY_BAND_70_CM_HAM:
```
olliw42, line 421: BUG BUG BUG
`break;` is missing here !!!!

mLRS/Common/setup.h
```diff
@@ -396,6 +411,20 @@ void setup_configure(void)
...
+        switch (Config.Mode) {
+        case MODE_31HZ: Config.FhssNum = FHSS_NUM_BAND_70_CM_HAM; break;
+        case MODE_19HZ: Config.FhssNum = FHSS_NUM_BAND_70_CM_HAM_19HZ_MODE; break;
+        default:
+            while (1) {} // must not happen, should have been resolved in setup_sanitize()
+        }
     default:
```
olliw42, line 428: BUG BUG BUG
`break;` is missing here !!!!

olliw42:
oh man ... 1.5 h later ...
see the comments

my units are now working, supposedly at 70 cm ...

mLRS/Common/setup.h
jlpoltrack: Isn't break for 868 just on the next line? 

image

jlpoltrack: Oh I see before this - my bad.

mLRS/Common/setup.h
jlpoltrack: Fixed

mLRS/Common/setup.h
jlpoltrack: Fixed

jlpoltrack:
Many, many thanks.  :)

Did you run into the issue with 433 EU Bands hanging?

Last one one from me - I want to have out and only use  2 uarts, so in the HAL I have (all other UART stuff commented out):

image

image

'uart_setprotocol' not declared? 
image

If I change to UART2 it compiles?
image

olliw42:
> Did you run into the issue with 433 EU Bands hanging?

I haven't tried 433 yet, only 70 cm

> Last one one from me 

maybe before doing it fancy, could you pl make a simple test case, and confirm that it now works with 70 cm, and that you do get a good connection, can use the lua, etc ??

olliw42:
> 'uart_setprotocol' not declared?

yes, that's not yet declared for LPUART1 
see https://github.com/olliw42/stm32ll-lib/blob/main/src/stdstm32-uart.h#L577-L578

I hadn't the time yet ... I'm just a human :)

olliw42:
can you pl also rebase onto main, I pushed few things

jlpoltrack:
Resolved conflicts (new to me) let me know if I missed anything.

mLRS/Common/fhss.h
```diff
@@ -29,7 +29,10 @@
 #ifdef FREQUENCY_BAND_433_MHZ
 #define FHSS_HAS_CONFIG_433_MHZ
 #endif
-#ifdef FREQUENCY_BAND_868_MHZ
+#ifdef FREQUENCY_BAND_70_CM_HAM
+#define FHSS_HAS_CONFIG_70_CM_HAM
+#endif
+#if defined FREQUENCY_BAND_868_MHZ
```
olliw42, line 35: here it went a bit wrong, should have stayed` #ifdef FRE...`

mLRS/Common/hal/hal.h
```diff
@@ -300,8 +300,9 @@ Note: Some "high-level" features are set for each device in the device_conf.h fi
 
 #if !defined FREQUENCY_BAND_2P4_GHZ && \
     !defined FREQUENCY_BAND_915_MHZ_FCC && !defined FREQUENCY_BAND_868_MHZ && \
-    !defined FREQUENCY_BAND_433_MHZ
-  #error At least one frequency band must be defined !
+    !defined FREQUENCY_BAND_433_MHZ && !defined FREQUENCY_BAND_70_CM_HAM
+  #error Correct frequency band must be defined !
+
```
olliw42, line 305: here it went a bit wrong
shoudl have stayed `#error At least one frequency band must be defined !`
and also an empty line got added

mLRS/Common/fhss.h
jlpoltrack: Fixed

mLRS/Common/hal/hal.h
jlpoltrack: Fixed

jlpoltrack:
For testing - can you give my some hints on getting away with just 2 UARTs (one with JR Pin 5) due to my hardware bug?
 
- If I try to just have UART + UARTC w/ define JR Pin 5 only, compiler complains about UARTB
- If I try to just have UART + UARTB w/ define JR Pin 5 only, compiler complains about UARTC

olliw42:
concerning the 433 MHz issue:
in fhss.cpp line 73, https://github.com/olliw42/mLRS/blob/main/mLRS/Common/fhss.cpp#L73, you need to write

`if ((config_i == FHSS_CONFIG_2P4_GHZ) && (k > 0)) {`

this is a breaking change as it will affect the hopping sequences of all ferquencies except of 2.4GHz. For the moemnt you could also write `if ((config_i != FHSS_CONFIG_433_MHZ) && (k > 0)) {`, but I iwll change it to teh above line (as it really does not make sense to keep gaps if there aren't many channels).

olliw42:
not sure I understand your 2 uarts issue
2 uarts should be fine, right? one for serial, one for jrpin5 or out
in principle you only would need one uart...

you MUST not fool with UART or UARTB or UARTC, all you can change is the part after the USE, like UARTB_USE_UART1, UARTB_USE_UART2, UARTB_USE_UART3, etc

for tx the associated has to strictly be as written:
// UARTB = serial port
// UARTC = COM (CLI)
// UARTD = serial2 BT/ESP port
// UART  = JR bay pin5
// UARTE = in port, SBus or whatever

olliw42:
what is the hardware bug again?
just looked into your schematics, and I can't really see a bug ...
is it that you think uart2 cannot be on PA15?
the tx-hal-diy-e22-g431kb.h which was added for you should give you the idea of how to do this

jlpoltrack:
Let me provide some more details, I'm a little further but having issues with CRSF / JR5 using this branch.

- Want to use UART1 PA9,PA10 for JR5 w/ R-D network
- Want to use UART2_REMAPPED2 PB3, PA15 for Serial (PA15 is wrong but can be changed as you note, I guess this is a G4 feature)
- Want to use LPUART1 PA2, PA3 for CLI

Only define is for JR5, in the HAL I've got:
- UARTB > UART_USE_UART2_REMAPPED2, with the RX_IO line (PB3, PA15)
- UARTC > LPUART1 (PA2, PA3)
- UART > UART1 (PA9,PA10)
- UARTE & UARTF are there (I assume they are ignored)

The CLI works over LPUART1 and I can change and store settings.  I've made sure that I've got CRSF for 'tx ch source' but can't seem to get CRSF working.  R-D network is in place as per the schematic.

HAL attached.  [tx-hal-diy-e22-g431kb.zip](https://github.com/olliw42/mLRS/files/10866919/tx-hal-diy-e22-g431kb.zip)

olliw42:
strange
if you use a differnt uart (UART2, LPUART1) for jrpin5, then it works? I.e., is this uart1 specific, or generally doesn't work?

olliw42:
@geofrancis 
since we are getting quite close for this to be merged, let me come back to your suggestion of  458.5 to 459.5
as @jlpoltrack said, that's quite tricky for mLRS, since it uses 500kHz bandwidth, and reserves at least one channel for binding, so that it would leave only one frequency.
technically, there is no difficulty with that and no reason not to do that, but you wouldn't have any hopping at all

at this point I wonder how this works for other systems you may use in this frequency band. Maybe it would be better discussed in the rcgropups thread, or as a new issue in this repo.

jlpoltrack:
@olliw42 
Thanks for your review of my HAL, too many copy and pastes :)  Currently have CRSF + Lua working,  can bind and change settings on both Tx and Rx.  Today I will:

- See if I can get CRSF on a FC and verify MAVLink stream 
- Fine tune the power config (really just for my setup I guess)

olliw42:
FANTASTIC

ok, so I think this means that we can merge this PR. We do have now two people having it working in this range (you and me). What you indicate as missing are all not related to 433/70. 

agree?

geofrancis:
> @geofrancis since we are getting quite close for this to be merged, let me come back to your suggestion of 458.5 to 459.5 as @jlpoltrack said, that's a bit tricky for mLRS, since it uses 500kHz bandwidth, and reserves at least one channel for binding, so that it would leave only one frequency. technically, there is no difficulty with that and no reason not to do that, but you wouldn't have any hopping at all
> 
> at this point I wonder how this works for other systems you may use in this frequency band. Maybe it would be better discussed in the rcgropups thread, or as a new issue in this repo.
> 
> I also wonder: isthis actually reasonable? I mean, isn't it that in the UK you also can use the 433 MHz band? (https://en.wikipedia.org/wiki/LPD433)

The lack of hopping isnt really a problem, but if its more work than just increasing the frequency range then dont worry about it. its not essential to anything im doing

jlpoltrack:
@geofrancis 
I like subs, so can definitely look at doing this in the future :)  Related, would there be any feedback on the current hardware?  Size, format, etc?  I can say that I really like this USB DFU for firmware.

@olliw42 
While this is still open, can confirm that I got CRSF on the FC working.

olliw42:
> While this is still open, can confirm that I got CRSF & MAVLink on the FC working.

GREAT. So, you have it all working now, right? And in the process you also "learned" that you don't had a hardware bug/design error, right?

many thx for pushing this forward. 433 was in the books for a while but needed someone with that hardware & application. Now it has become reality, to the benefit of everyone.

I will merge ASAP.

jlpoltrack:
The only thing that I need to do is see if I can get the MAVLink stream from the TX side (I just checked that all of the passthrough sensors showed up).

>  And in the process you also "learned" that you don't had a hardware bug/design error, right?
I didn't realize that pins could be remapped / split like that but I'm glad that you knew. :)

olliw42:
merging now

MANY THX again!!!

I'm also closing. If anyone feels it should be ropened, pl do so :)

PR #53, opened 2023-02-28
WiFi Channel Parameter

jlpoltrack:
Add a define for WiFi channel, set to 13 which should overlap least with mLRS freq list and home WiFi.

Tested on my ESP32.

### DIFF
--- a/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
+++ b/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
@@ -73,7 +73,7 @@ int baudrate = 57600;
 
 // WiFi channel
 // comment out for default setting
-#define WIFI_CHANNEL  13 // Use 13 (2461-2483 MHz) since it has least overlap with mLRS 2.4 freqs, default is 1
+int WIFI_CHANNEL = 13; // Use 13 (2461-2483 MHz) since it has least overlap with 2.4 freqs
 
 
 // serial port usage (only effective for a generic board)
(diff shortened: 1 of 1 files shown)


olliw42:
very good idea
I actually had the very same idea and did some play with it this morning, just didn't added it

would you mind to not do it as a #define but as int wifi_channel = 13; ?

jlpoltrack:
Okay to use 'uint8_t' data type?

olliw42:
good question. Pl use int.
Somehow Arduino seems to uses this more often, and the function softAP() uses int.

jlpoltrack:
Okay :)

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -71,6 +71,10 @@ int baudrate = 57600;
 // comment out for default setting
 #define WIFI_POWER  WIFI_POWER_2dBm // WIFI_POWER_MINUS_1dBm is the lowest possible, WIFI_POWER_19_5dBm is the max
 
+// WiFi channel
+// comment out for default setting
+int WIFI_CHANNEL = 13; // Use 13 (2461-2483 MHz) since it has least overlap with 2.4 freqs
```
olliw42, line 76: lower case lower case :D

jlpoltrack: Thanks for the patience :) I see the pattern now:
- lower for variables
- upper for defines

olliw42: 👍

olliw42:
MANY THX sir !!!

PR #55, opened 2023-03-05
Image calibration

jlpoltrack:
Added based on datasheet 9.2.1.  Supposed to help Rx performance - saw no real change in RSSI on the bench with and without.

### DIFF
--- a/mLRS/Common/sx-drivers/sx126x_driver.h
+++ b/mLRS/Common/sx-drivers/sx126x_driver.h
@@ -165,14 +165,14 @@ class Sx126xDriverCommon : public Sx126xDriverBase
         //Default for SX1261/2 is 902 to 902 MHz
         //Default for SX1268 is 470 to 510 MHz
         switch (Setup.FrequencyBand) {
-			case SETUP_FREQUENCY_BAND_915_MHZ_FCC: CalibrateImage(SX126X_CAL_IMG_902_MHZ_1, SX126X_CAL_IMG_902_MHZ_2); break;
-			case SETUP_FREQUENCY_BAND_868_MHZ: CalibrateImage(SX126X_CAL_IMG_863_MHZ_1, SX126X_CAL_IMG_863_MHZ_2); break;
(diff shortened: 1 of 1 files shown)


mLRS/Common/sx-drivers/sx126x_driver.h
```diff
@@ -160,6 +160,19 @@ class Sx126xDriverCommon : public Sx126xDriverBase
         if (osc_configuration != SX12xx_OSCILLATOR_CONFIG_CRYSTAL) {
             SetDio3AsTcxoControl(osc_configuration, 250); // set output to 1.6V-3.3V, ask specification of TCXO to maker board
         }
+
+        //Image Calibration per datasheet 9.2.1
```
olliw42, line 164: blank after //
in the next lines first char lower case pl :)

olliw42:
this looks well done, just a pititesse

I would however wish we could get more testing/confirmation on this, before merging, no idea how to achieve that though

let's try to get a feedback from the original author of the sx126x lib

@Imen-D: can you maybe comment on this PR. Is this what one should do? Thx.

mLRS/Common/sx-drivers/sx126x_driver.h
jlpoltrack: Updated.

olliw42: THX

Imen-D:
Hi Guys,

Image calibration is part of calibration process when device startup whether from POR or sleep. mLRS didn't make any command to chip to sleep. So we can focused on POR. Our lib will wait this process when startup and will be done before we setup TCXO, CMIIW. As mLRS keep develop and changing, hence we can use different frequency range, i.e E22 900M can use 915 or 868 band, and sx1268 with default 470~510 which is not our common frequency (433), so we must do image calibration.

olliw42:
@Imen-D 
many thx for chimming in :)
So, you confirm what ImageCalibration shoudl be done. 

Could you pl also have a look at the code which @jlpoltrack is suggesting in this PR, and confirm/disconfirm if it is in the correct place?

Imen-D:
@olliw42, @jlpoltrack , yes it is good for me too

olliw42:
@jlpoltrack @Imen-D 

ok, so I'm merging :)

MANY THX folks!

Issue #57, opened 2023-03-07
Proposal | Include 24 dBm (251 mW) instead of 23 dBm (200 mW)


jlpoltrack:
This change would affect:

- Frsky R9M
- All E28 variants

Reason - there is no 200 mW CRSF element but there is one for 250 mW.  Let me know and I can create a PR.

olliw42:
ähm hmmm uhm ... I have no strong opinion on this ... both is fine with me. I would be interesting to get more opinons. Somehow.

olliw42:
done
THX!!

PR #58, opened 2023-03-08
24 dBM instead of 23

jlpoltrack:
As mentioned here: #57

### DIFF
--- a/mLRS/Common/hal/rx-hal-R9M-868-f103c8.h
+++ b/mLRS/Common/hal/rx-hal-R9M-868-f103c8.h
@@ -278,7 +278,7 @@ const rfpower_t rfpower_list[] = {
     { .dbm = POWER_3_DBM, .mW = 2 },
     { .dbm = POWER_10_DBM, .mW = 10 },
     { .dbm = POWER_20_DBM, .mW = 100 },
-    { .dbm = POWER_24_DBM, .mW = 251 },
+    { .dbm = POWER_24_DBM, .mW = 250 },
     { .dbm = POWER_27_DBM, .mW = 500 },
     { .dbm = POWER_30_DBM, .mW = 1000 },
 };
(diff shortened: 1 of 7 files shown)


mLRS/Common/hal/rx-hal-R9M-868-f103c8.h
```diff
@@ -239,8 +239,10 @@ void rfpower_calc(int8_t power_dbm, uint8_t* sx_power, int8_t* actual_power_dbm,
     voltage_mV = 1730;
     *actual_power_dbm = 27;
   } else if (power_dbm > 22) {
-    voltage_mV = 1390;
-    *actual_power_dbm = 23;
+    //voltage_mV = 1390;
+    //*actual_power_dbm = 23;
+    voltage_mV = 1475;
```
olliw42, line 244: did you guesstimate this value, or did you had a chance to adjust it using a power meter?

mLRS/Common/hal/rx-hal-R9M-868-f103c8.h
```diff
@@ -276,7 +278,7 @@ const rfpower_t rfpower_list[] = {
     { .dbm = POWER_3_DBM, .mW = 2 },
     { .dbm = POWER_10_DBM, .mW = 10 },
     { .dbm = POWER_20_DBM, .mW = 100 },
-    { .dbm = POWER_23_DBM, .mW = 200 },
+    { .dbm = POWER_24_DBM, .mW = 251 },
```
olliw42, line 281: I guess 250 should be accurate enough and potentially less confusing
dito in below occurences

mLRS/Common/hal/rx-hal-R9M-868-f103c8.h
jlpoltrack: Guessed 1450, needed 1475 to be consistently over 24 dBm on the IRC power meter.  I now recall there were concerns on R9 overheating, can lower have a safer value if we want.

mLRS/Common/hal/rx-hal-R9M-868-f103c8.h
jlpoltrack: Fair enough, will update.

olliw42:
many thx, sir !!!

PR #61, opened 2023-03-10
Fix path directory delimiter for Linux

brad112358:
Old habits?  :-)


olliw42:
I certainly didn't had this in mind when I typed this LOL
so: yes
good that you are watching :):)
MANY thx

PR #62, opened 2023-03-10
Adjust the bounds and don't delay/rate limit state transitions which don't need the delay

brad112358:
This is includes changes I suggested earlier.  It works well in my quick test, eliminating "overshoot" in slowing down during parameter download.


olliw42:
MANY thx.
Pl put that into the function called for the option "brad". Pl feel free to also consider a better name for that option.
:)

olliw42:
I think can be closed since PR https://github.com/olliw42/mLRS/commit/1e4ff4fdc74ce17fa2c69afd4bc6473430445404 was merged.
MANY THX again :)

PR #63, opened 2023-03-14
Rewrite handle_txbuf_px4()

brad112358:
This is mostly adopted from Olli's latest code in handle_txbuf_ardupilot() with some changes intended to improve response time.
Also change setup string from brad to other

In my testing, this seems to work at least as well for simple cases as the more complex code it replaces for both Ardupilot and PX4 and I think it should also adapt faster to extra traffic routed from cameras, gimbals, etc.

### DIFF
--- a/mLRS/Common/setup_list.h
+++ b/mLRS/Common/setup_list.h
@@ -91,7 +91,7 @@
   X( Setup.Rx.FailsafeMode,       LIST, "Rx FailSafe Mode", "RX_FAILSAFE_MODE", 0,0,0,"", "no sig,low thr,by cnf,low thr cnt,ch1ch4 cnt", MSK_ALL )\
   X( Setup.Rx.SerialBaudrate,     LIST, "Rx Ser Baudrate",  "RX_SER_BAUD",      0,0,0,"", SETUP_OPT_RX_SERIAL_BAUDRATE, MSK_ALL )\
   X( Setup.Rx.SerialLinkMode,     LIST, "Rx Ser Link Mode", "RX_SER_LNK_MODE",  0,0,0,"", SETUP_OPT_SERIAL_LINK_MODE, MSK_ALL )\
-  X( Setup.Rx.SendRadioStatus,    LIST, "Rx Snd RadioStat", "RX_SND_RADIOSTAT", 0,0,0,"", "off,ardu_1,brad", MSK_ALL )\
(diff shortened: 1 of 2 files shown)


mLRS/Common/setup_list.h
```diff
@@ -91,7 +91,7 @@
   X( Setup.Rx.FailsafeMode,       LIST, "Rx FailSafe Mode", "RX_FAILSAFE_MODE", 0,0,0,"", "no sig,low thr,by cnf,low thr cnt,ch1ch4 cnt", MSK_ALL )\
   X( Setup.Rx.SerialBaudrate,     LIST, "Rx Ser Baudrate",  "RX_SER_BAUD",      0,0,0,"", SETUP_OPT_RX_SERIAL_BAUDRATE, MSK_ALL )\
   X( Setup.Rx.SerialLinkMode,     LIST, "Rx Ser Link Mode", "RX_SER_LNK_MODE",  0,0,0,"", SETUP_OPT_SERIAL_LINK_MODE, MSK_ALL )\
-  X( Setup.Rx.SendRadioStatus,    LIST, "Rx Snd RadioStat", "RX_SND_RADIOSTAT", 0,0,0,"", "off,ardu_1,brad", MSK_ALL )\
+  X( Setup.Rx.SendRadioStatus,    LIST, "Rx Snd RadioStat", "RX_SND_RADIOSTAT", 0,0,0,"", "off,ardu_1,other", MSK_ALL )\
```
olliw42, line 94: "other" is better than "brad" but still somewhat ugly :)
so I suggest "mtd_b1"

mLRS/CommonRx/mavlink_interface_rx.h
```diff
@@ -448,83 +446,95 @@ if(txbuf>90) dbg.puts("-20 "); else dbg.puts("0   ");
 }
 #endif
 
-
+// This method should be selected for PX4 and currently may be a useful alternative for Ardupilot
 bool MavlinkBase::handle_txbuf_px4(uint32_t tnow_ms)
```
olliw42, line 450: could you pl rename this function to "xxx_method_b1"
unless you want to keep it "_px4" to emphasize it's default nature for px4, but when I would argue also the option should be named "px4" and not "mth_b1" :)

olliw42:
great!

few minor inline comments :)

olliw42:
MANY THX!

PR #67, opened 2023-04-03
Update SX126x to handle RSSI 16 Bit

jlpoltrack:
Needed for https://github.com/olliw42/sx12xx-lib/pull/5


olliw42:
duplication, I've pushed this already :)
MANY thx nevertheless :)

PR #69, opened 2023-04-09
Add support for M5STACK STAMP PICO for use with R9M

brad112358:


esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -65,7 +66,7 @@ int port_udp = 14550; // connect to this port per UDP // MissionPlanner default
 int baudrate = 57600;
 
 // WiFi channel
-int wifi_channel = 13; // 1 is the default, 13 (2461-2483 MHz) has the least overlap with mLRS 2.4 GHz frequencies
+int wifi_channel = 13; // 1 is the default, 13 (2461-2483 MHz) has the least overlap with mLRS 2.4 GHz frequencies, but is not used in the US
```
olliw42, line 69: maybe "... frequencies; also is not used in the US"

brad112358: The but is important.  Wifi Channel 13 is generally not available in the US.  It only goes to 11 :)

How about ... frequencies;  Note, channels higher than 11 may not work with US WiFi Clients.

olliw42: ah, it's probably a language issue on my side but the but still reads odd, it triggers a wrong thought. I find your explanation better. So, maybe_ "... frequencies. Note: Channel 13 is generally not available in the US, where 11 is max." ?
Since the line may become long you simply could put the whole comment inbetween // Wifi ... and int wifif_...

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -128,8 +131,8 @@ void setup()
 
     size_t rxbufsize = SERIAL.setRxBufferSize(2*1024); // must come before uart started, retuns 0 if it fails
     size_t txbufsize = SERIAL.setTxBufferSize(512); // must come before uart started, retuns 0 if it fails
-#ifdef SERIAL_RXD // if SERIAL_TXD is not defined the compiler will complain, so all good
-    SERIAL.begin(baudrate, SERIAL_8N1, SERIAL_RXD, SERIAL_TXD);
+#ifdef SERIAL_RXD // if SERIAL_TXD or SERIAL_INVERT is not defined the compiler will complain, so all good
```
olliw42, line 134: my comment on this seems to have disappeared, sory for that
could you please change it as follows so that one can define rxd,txd without invert

#ifdef SERIAL_RXD
#ifdef SERIAL_INVERT
Serial.begin(..., SERIAL_INVERT);
#else
Serial.begin(...);
#endif
#endif

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -188,7 +191,7 @@ void loop()
     int packetSize = udp.parsePacket();
     if (packetSize) {
         int len = udp.read(buf, sizeof(buf));
-        SERIAL.write(buf, len);
+        SERIAL.write(bufp, len);
```
olliw42, line 194: ?

brad112358: Editor command got inserted.  Sorry!

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
brad112358: This makes little sense to me, especially since you didn't allow defining SERIAL_RXD without SERIAL_TXD.  But, it's not worth debate.

olliw42: LOL

olliw42:
MANY thx!

PR #70, opened 2023-04-09
Don't send so many very short UDP messages to avoid message loss

brad112358:
With QGround control, on both Linux and Android, I got lots of lost MAVLink messages without this change.  When I checked, I saw that most UDP packets contained only a few bytes (1-3) of payload.  I don't know which end the messages were being lost on.

The intent of this change is to wait for more bytes to be received on the serial link before sending a UDP packet, but still constrain the latency to a reasonable 5ms.

### DIFF
--- a/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
+++ b/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
@@ -111,7 +111,7 @@ bool led_state;
 
 bool is_connected;
 unsigned long is_connected_tlast_ms;
-unsigned int send_tlast_ms;
+unsigned long send_tlast_ms;
 
 
 void serialFlushRx(void)
--- a/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
+++ b/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
@@ -169,7 +169,8 @@ void setup()
 
 void loop() 
 {
-    int tnow_ms = millis();
+    unsigned int tnow_ms = millis();
+    static unsigned int last_send = 0;
     if (is_connected && (tnow_ms - is_connected_tlast_ms > 2000)) { // nothing from GCS for 2 secs
         is_connected = false;
     }
@@ -192,12 +193,14 @@ void loop()
         is_connected = true;
         is_connected_tlast_ms = millis();;
     }
-
-    while (SERIAL.available()) {
+    int avail = SERIAL.available();
+    // Don't send nearly empty messages so often
+    if ((avail > 30) || (avail && (tnow_ms - last_send > 5))) {
         int len = SERIAL.read(buf, sizeof(buf));
         udp.beginPacket(ip_udp, port_udp);
         udp.write(buf, len);
-        udp.endPacket();    
+        udp.endPacket();
+        last_send = tnow_ms;
     }   
 
 #else // TCP
--- a/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
+++ b/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
@@ -110,7 +110,8 @@ int led_tlast_ms;
 bool led_state;
 
 bool is_connected;
-int is_connected_tlast_ms;
+unsigned long is_connected_tlast_ms;
+unsigned int send_tlast_ms;
 
 
 void serialFlushRx(void)
@@ -169,15 +170,14 @@ void setup()
 
     is_connected = false;
     is_connected_tlast_ms = 0;
-
+    send_tlast_ms = 0;
     serialFlushRx();
 }
 
 
 void loop() 
 {
-    unsigned int tnow_ms = millis();
-    static unsigned int last_send = 0;
+    unsigned long tnow_ms = millis();
     if (is_connected && (tnow_ms - is_connected_tlast_ms > 2000)) { // nothing from GCS for 2 secs
         is_connected = false;
     }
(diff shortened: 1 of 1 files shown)


esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -169,7 +169,8 @@ void setup()
 
 void loop() 
 {
-    int tnow_ms = millis();
+    unsigned int tnow_ms = millis();
```
olliw42, line 172: ah, good catch
just googled, according to arduino docs millis() is unsigned long, if so, could probably be better to use that

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -169,7 +169,8 @@ void setup()
...
+    static unsigned int last_send = 0;
```
olliw42, line 173: could you pl make this a global variable, and init it in setup(), and maybe unsigned long
I guess I would like send_tlast_ms better (see is_connected_tlast_ms)

olliw42:
great, this makes lots of sense
I guess the 30 and 5ms has worked out well experimentally, couldn't this be increased to 10ms or even 20ms?
could you also test with MissionPlanner? (not asking for, I'm just curious)

brad112358:
I haven't tested with mission planner yet, but I will let you know when I do.  30 and5 was the first thing I tried and it helped so much I didn't think I needed to try others.  But, after I submitted this PR, I discovered a few messages were still being lost so I have already increased the delay and message size which helped even more.  I'll update this PR later.

I also have another idea for a heuristic which wouldn't add latency as much as this does:  mLRS sends a mavlink message at a time in a burst, so we could actually wait for the break in serial reception, when the receive size stops growing for a ms or so, then send the packet.  This should be especially effective at a slightly higher baud rate.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
brad112358: > could you pl make this a global variable, and init it in setup(), and maybe unsigned long I guess I would like send_tlast_ms better (see is_connected_tlast_ms)

I suppose, if you insist, but can you elaborate on why do you want to do it this way?  

Function local static variables occupy the same storage segment and have the same lifetime as global variables, but with scope limited to the function they are declared in.  This helps make clear that they are not and can not be used anywhere else.  Unless setup() is called more than once in the lifetime of the program, the initialization here in the static variable declaration should have exactly the same effect as initializing a global variable in setup().  It is initialized once when the program starts.  Initializing it here, closer to where it is used makes it easier to find the initialization when reading the code.  This use of static function local variables is generally considered best practice for these reasons.  

I'm guessing you already know all of this, but this is why I prefer the static variable.

olliw42: there are also a couple of disadvantages with statis which I'm guessing you also know all, and I do not agree with "is generally considered best practice", that's a matter of who you ask :)
you can keep them as you want, but I'm going to change anyway :)

pl check out if one shouldn't use u long, and consider the renaming

olliw42:
I'm really not concerned about 20ms vs 5ms
in other applications of mine, like the uavcan mavlik tunnel I did many years back, few 10-ish ms worked great
anyway, put in what you think is the best, it always can be further improved later

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
brad112358: Yes, I have no objection to your other requests, just the static.  And I explained why I prefer it.

As I said, if you insist. But I asked why;  What specific disadvantage of the static local approach is motivating your decision to avoid it?  Answering my question will help us avoid wasted time in future PRs

brad112358:
I did code up the heuristic to align UDP payload with MAVLink messages and it works quite well, allowing wireshark to decode mavlink messages and further limiting UDP message rate without much latency.  I think I'll wait for your answer on why you don't like static locals before I push it.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
olliw42: this is a good code addition, can we just get it done without drama pl. I sure could spend time to put down muh text, but would it change your opinion, I guess no, so what's the point. Do it or don't do it. :)

brad112358: I'm sorry you consider this drama.  Yes, it might not change my mind, or it might.  Either way, it's your project so I'll do it your way.  But I'd appreciate it if you would do me the courtesy of explaining.  Besides, understanding your reasoning might help me write code you like better in the future with minimum drama.

olliw42: the simple answer would be that I myself have spend too much time and have seen others spending way too much time in finding bugs which at the end were just not reset or re-inited static variables. You put out "is generally considered best practice" which is a pretty mighty and absolutistic claim and I have no intention to convince you otherwise. There is no real relevant advantage of statics but real downsides in both that it is error prone and not compiler efficient. Your statement implies you don't see any disadvantage, and haven't hit the downsides, good for you, but I try to avoid them. I have no itention to produce a list of situations were a static is not only not the best practice but in fact just silly except that in mLRS we have such situations.
btw, I'm pretty sure we have a reset/reinitialization issue in mLRS which I couldn't find yet. I'm surprised the phenomenon hasn't been reported yet. So much on "proper initialization of variables".

brad112358:
I'm sorry if I've misunderstood what you wrote.  My goal really is that you
shouldn't feel the need to rework everything I submit.

brad112358:
I did test mission planner on Windows 10 both with and without this change at 19Hz.  Both worked fine with no sign of dropped messages.  I'm not sure why QGC drops so many messages on Linux and Android with the original code.

Have you tested UDP bridge yet with Android?

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -110,7 +110,8 @@ int led_tlast_ms;
 bool led_state;
 
 bool is_connected;
-int is_connected_tlast_ms;
+unsigned long is_connected_tlast_ms;
+unsigned int send_tlast_ms;
```
olliw42, line 114: should be u long too, shouldn't it?

brad112358: crap.  Of course.  I had already changed that in another branch and missed it in this one.

olliw42:
MANY THX :)

PR #72, opened 2023-04-19
Add support for M5STAMP_C3U_MATE configured for R9M

brad112358:
This works every bit as cleanly with the R9M as the M5STAMP PICO does.  Just 4 easy to solder thru-hole pins.  And yes, it's easier to program.  Just hold down the central button when plugging it in to USB


esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
```diff
@@ -215,6 +228,44 @@ GPIO15 = RTC_GPIO13
...
+    #endif
+
+    #undef USE_SERIAL_DBG1
+    #define USE_SERIAL1_DBG
+
+    #undef LED_IO
+    #define USE_LED
+    #include <Adafruit_NeoPixel.h>
+    #define NUMPIXELS  1
+    Adafruit_NeoPixel pixels(NUMPIXELS, 2, NEO_GRB + NEO_KHZ800);
```
olliw42, line 244: I'm a bit confused. We have 4 boards now with NeoPixels, yet the code looks quite different each time. Two have PIN_NEOPIXEL defined, two not. Two use #ifdef NEOPIXEL_POWER, two not. What is the pattern here? Can this somehow be streamlined? I understand it's nfc, but it just doesn't look nice and systematic.
My guts feeling is that:
- here we also want to define/use PIN_NEOPIXEL, but do not want to add the #ifdef NEOPIXEL_POWER stuff
- for the M5 STAMP PICO we want to remove the #ifdef NEOPIXEL_POWER stuff
- for the M5Stamp C3 Mate we want to use PIN_NEOPIXEL

olliw42:
GREAT!

just one observation

olliw42:
so, if I understand the 5V issue, the procedure is to take off the stamp, connect to usb and flash, and then to put it back on, right?

I would wish also the sbus pin would be connected to one of the esp gpio's, but not in reach LOL

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
```diff
@@ -215,6 +228,44 @@ GPIO15 = RTC_GPIO13
...
+    }
+
+    void led_off(void)
+    {
+        pixels.fill(0x000000); // off
+        pixels.show();
+    }
+
+    #define SERIAL_RXD  1 // = RX1
+    #define SERIAL_TXD  0 // = TX1
```
olliw42, line 265: interesting
these are the pads labelled porta G1,G0, aren't they?
maybe one could put down a respective comment, I see you referred to IO1/IO0, maybe one could add how they are labelled in M5 dialect

brad112358:
> so, if I understand the 5V issue, the procedure is to take off the stamp, connect to usb and flash, and then to put it back on, right?

Correct.

vrquaeler:
I love seeing more options to connect the R9M to the GCS  computer. 

> I would wish also the sbus pin would be connected to one of the esp gpio's, but not in reach LOL

May I ask what the purpose of connecting the sbus bin to the ESP32 could be?
I did not get it.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
brad112358: I've just been following the example of a similar board, so these inconsistencies were already present.

But I'm also confused.
I don't see anything that ever defines  NEOPIXEL_POWER.  Is any of that code needed at all?
I don't see anything that defines PIN_NEOPIXEL when MODULE_ADAFRUIT_QT_PY_ESP32_S2 is defined.  Does this board compile?

vrquaeler: The "ADAFRUIT_QT_PY_ESP32_S2" has a dedicated pin to be able to switch the WS2812 on an off for IoT Low Power applications. This feature may be not available on the M5-Stamp-C3.

[https://learn.adafruit.com/adafruit-qt-py-esp32-s2/pinouts#neopixel-led-3107926](url)

olliw42:
> May I ask what the purpose of connecting the sbus bin to the ESP32 could be?

I'm thinking about a way to configure the esp from the tx module. This would e.g. allow to set the wifi credentials, or the baudrate, and such, in the esp from the lua,cli,display. 
In order to tell the esp when the serial trafic via the uart is not data but for configuration, the approach I have in mind is to set one pin to e.g. zero.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
olliw42: yes, the PyQt sure does compile (and the power stuff is needed)(and enclosing it in the if comes from an example of Adafruit)

olliw42: based on my version for one of the M5's I suspect that the power stuff isn't needed for all three m5's, hence the suggestion to remove it there. Defining PIN_NEOPIXEL looks nice to me as it tells what the number means, hence my suggestioon to do that for all three.

brad112358:
> In order to tell the esp when the serial trafic via the uart is not data but for configuration, the approach I have in mind is to set one pin to e.g. zero.

I suppose this could be done neatly enough by carefully trimming a double wide female header connector or using the existing connector with some glue and soldering a wire to the extra pin.  But this significantly complicates the build,
Instead, what about a software solution where the transmitter always sends the configuration at power on.  Requiring an extra power cycle after configuration changes should be acceptable I'd think.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
olliw42: if in this PR or a follow up PR I don't mind, I think it would be just easier to grab this occasion
I can't test the two newly added M5's, so, preferably you do it, either here or in a follow up, otherwise I would need explict confirmation that the suggested changes are ok so I could do the follow up PR.

olliw42:
I was sure thinking about a software only solution, would be just so much nicer, but there was an issue which I don't recall which made me get back to the gpio thing
your other PR, when we have syncing to silent gaps, may aleviate my older issues however
your idea of doing config only at power up is probably a very good one, the tx and rx are power cycled on "store" anyways, and one could deal with the esp equally as the receiver is dealt with ... this sounds indeed like a very possible approach, I will keep it in mind, thx

olliw42:
may I ask you for anorther slight change:

given that your two additions are clearly not canonical but targeted for use with the R9M, could you please change the name of the defines to
MODULE_M5STAMP_PICO_FOR_FRSKY_R9M
MODULE_M5STAMP_C3U_MATE_FOR_FRSKY_R9M

two further ideas, not for this PR, but I just like to grab the occasion:
- I think the C3 Mate should also be a great candidate, and work essentially like the C3U mate, shouldn't it? I wonder if one shouldn't add a MODULE_M5STAMP_C3_MATE_FOR_FRSKY_R9M, in a follow up PR.
- Can't we use the C3/C3U also as serial inverter, and thus also for flashing the firmware using the system bootloader (aka "upgrading"). The idea would be to use the esp's 2nd uart and to just establish a passthrough to the uart connected to the R9M. It shoudl be actually possible to use this in conjunction with the UDP/TCP path, i.e., at the same time. This would allow to (1) get the mavlink stream also via a serial (if for some reason this is preferred over wireless) but also (2) to upgrade the firmware. It is sad the C3/C§U have this 5V issue, since when it could be done via it's USB port. Yet, being able to upgrade the firmware with just a regular usb-tll adapter connected to teh esp could be very significant simplification.

brad112358:
> I think the C3 Mate should also be a great candidate, and work essentially like the C3U mate, shouldn't it? I wonder if one shouldn't add a MODULE_M5STAMP_C3_MATE_FOR_FRSKY_R9M, in a follow up PR.

Shouldn't the same binary should work?.  Perhaps add a comment rather than another target define.

> Can't we use the C3/C3U also as serial inverter

Yes, but once we get the ELRS firmware install method working, that method is even easier than the bootloader in the case of the R9M

olliw42:
> Shouldn't the same binary work? Perhaps add a comment rather than another target define.

I don't know but wouldn't think so. The C3 has a usb-ttl adapter, the C3U uses native usb.

> Yes, but once we get the ELRS firmware install method working, that method is even easier than the bootloader in the case of the R9M

yeah, if that will work out, if not it could be an option

MANY THX!
will merge now :)

PR #73, opened 2023-04-22
tx-diy-E28-dual-f103cb add test of buzzer pin

vrquaeler:
Hi Olli,
The buzzer pin is broken but not used yet. Here is a small enhancement of the test code. 

PS: Looking forward to see some "buzzer" code.


olliw42:
THX!!

there is buzzer code already, see other hal files as example, and https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/hal.h#L43. If you want to have a physical buzzer one needs to add some hardware though, the buzzer, a transistor, a diode and 2 resistors, see https://github.com/olliw42/mLRS-hardware/blob/master/olliw-stm32-based/tx-module-02/mLRS-tx-dual-base-v031-np-scheme.png.

PR #76, opened 2023-05-03
50 Hz baud rate increase

jlpoltrack:
I tried this out on Arduplane 4.3.5 and 4.4.0-Beta and this gave a nice increase in speed when downloading parameters over MAVFTP on 50 Hz.

- 115200 baud = 11,520 bytes / sec
- Ardu limits to 30%
- 11,520 * 0.3 = 3,456 bytes / sec
- 50 Hz downlink is 4,100 bytes / sec

### DIFF
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -202,7 +202,7 @@ void MavlinkBase::putc(char c)
 #if MAVLINK_OPT_FAKE_PARAMFTP > 1
         bool force_param_list = true;
         switch (Config.Mode) {
-        case MODE_50HZ: force_param_list = (Config.SerialBaudrate > 57600); break; // 57600 bps and lower is ok for mftp
+        case MODE_50HZ: force_param_list = (Config.SerialBaudrate > 115200); break; // 115200 bps and lower is ok for mftp
         case MODE_31HZ: force_param_list = (Config.SerialBaudrate > 57600); break; // 57600 bps and lower is ok for mftp
         case MODE_19HZ: force_param_list = (Config.SerialBaudrate > 38400); break; // 38400 bps and lower is ok for mftp
         }


olliw42:
good call
I'm a bit confused since I did this math before as well as some testing and concluded for 57600. Maybe I just wanted to leave some more room and probably considered it fast enough anyway
I will test

olliw42:
I now remember why I didn't do this

the "txbuf" flow mechanism further below tries to keep the stream in the range of 75%-85%, which means that if there is a MAVFtp flow of 0.3 * 115200 = 3456 bytes/s or 3456/4100 = 84% it will trigger nearly all time and drain out the "normal" stream. To avoid this is the reason that the numbers had been adjusted such to be below the 75% threshold, and not adjusted to be "only" below 100%.

Note that in ArduPilot the MAVFtp stream is NOT controlled by txbuf, while the "normal" stream is, which creates the situation.

While for 19&31Hz mode we get above 50% stream usage, for the 50Hz mode the situation is unfortunate since with 57600 we only get 42%, and with 115200 it are your 84%. So, we have a typical situation there the decission strongly depends on how exactly the advantages/disadvantages are balanced out, and a slight shift in the balance, because some other persons just have different preferences, has strong effect on the result. 

So much regards why 57600 was choosen as limit in the first place. I'm not yet sure what I would want to conlcude however. I see your point.

jlpoltrack:
Ah I see.  Bummer that Ardu doesn't offer more flexibility with baud rates to allow for a more optimal setting.  Fine to close if you want - just noticed when I was trying out 31 / 50 Hz (now that these play nice with EdgeTx) and didn't see much of a difference in parameter download.

olliw42:
yeah, Ardu could have tried to get things right, but I think it's much better with 4.4 than it was before, and it's not too bad.

I'm not saying on shouldn't do MAVFtp for 115200, apparently you found it to work for you, it's just I'm not sure what the best compromise is. I apparently rather wanted to play it a bit safe than to squeeze out the best param upload speed, because I found the upload speed ok for me. 
So, the question is how important the higher param upload speed really is to us. Is it just to have this "waauuu" effect? 
I'm undecided on this matter.

olliw42:
@jlpoltrack 
could this be closed?
it seems to me that there is no real "pressure" for it, and things are largely satisfying as they are. At least for the moment (one always can come back)

jlpoltrack:
Fine to close

PR #79, opened 2023-05-17
IN 866 Fixes

jlpoltrack:
- Small channel list fix
- Spelling in setup_list

### DIFF
--- a/mLRS/Common/fhss.cpp
+++ b/mLRS/Common/fhss.cpp
@@ -71,7 +71,7 @@ void FhssBase::generate(uint32_t seed)
         // ensure it is not too close to the previous
         // do only if we have plenty of channels at our disposal
         bool is_too_close = false;
-        if (((config_i != FHSS_CONFIG_433_MHZ) || (config_i !=FHSS_CONFIG_866_MHZ_IN)) && (k > 0)) { // TODO: use smarter method, e.g., cnt < 2/3
+        if (((config_i != FHSS_CONFIG_433_MHZ) && (config_i !=FHSS_CONFIG_866_MHZ_IN)) && (k > 0)) { // TODO: use smarter method, e.g., cnt < 2/3
             int8_t last_ch = ch_list[k - 1];
             if (last_ch == 0) { // special treatment for this case
                 if (ch < 2) is_too_close = true;
--- a/mLRS/Common/fhss.cpp
+++ b/mLRS/Common/fhss.cpp
@@ -71,7 +71,7 @@ void FhssBase::generate(uint32_t seed)
         // ensure it is not too close to the previous
         // do only if we have plenty of channels at our disposal
         bool is_too_close = false;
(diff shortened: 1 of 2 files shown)


mLRS/Common/fhss.cpp
```diff
@@ -71,7 +71,7 @@ void FhssBase::generate(uint32_t seed)
         // ensure it is not too close to the previous
         // do only if we have plenty of channels at our disposal
         bool is_too_close = false;
-        if ((config_i != FHSS_CONFIG_433_MHZ) && (k > 0)) { // TODO: use smarter method, e.g., cnt < 2/3
+        if (((config_i != FHSS_CONFIG_433_MHZ) || (config_i !=FHSS_CONFIG_866_MHZ_IN)) && (k > 0)) { // TODO: use smarter method, e.g., cnt < 2/3
```
olliw42, line 74: shouldn't this be && ??

jlpoltrack: I thought there would only be 1 frequency value stored in config_i so we'd have to check for 433 _OR_ 866  IN? I could be wrong :)

olliw42:
yes, config_i can be only one of these values

therefore "not 433_MHZ or not 866_MHZ_IN" will be always true and the condition is obsolete

what we want is "not (433_MHZ or 866_MHZ_IN)" which is equa to "(not 433_MHZ and not 866_MHZ_IN)"

:)

jlpoltrack:
Now I see - updated :)

mLRS/Common/fhss.cpp
```diff
@@ -71,7 +71,7 @@ void FhssBase::generate(uint32_t seed)
         // ensure it is not too close to the previous
         // do only if we have plenty of channels at our disposal
         bool is_too_close = false;
-        if (((config_i != FHSS_CONFIG_433_MHZ) || (config_i !=FHSS_CONFIG_866_MHZ_IN)) && (k > 0)) { // TODO: use smarter method, e.g., cnt < 2/3
+        if (((config_i != FHSS_CONFIG_433_MHZ) && (config_i !=FHSS_CONFIG_866_MHZ_IN)) && (k > 0)) { // TODO: use smarter method, e.g., cnt < 2/3
```
olliw42, line 74: the () around the config_i checks can go :)

jlpoltrack: Updated

olliw42: not exactly as I had in mind but equally well, makes it also better readable
good

olliw42:
have you tested it?
I'm inclined to merge :)

jlpoltrack:
I only tested that it compiles.  We can give heads up on Discord for further testing :)

olliw42:
we both think it's correct, right?
maybe I'm just hilarious with asking for a test :)

jlpoltrack:
Hehe - alright.  Flashed to E5 Mini:

- 866 Label
- Can switch to 866 sucessfully

image

olliw42:
MANY THX !!!!
merging

PR #83, opened 2023-05-28
add lua supports B/W screen,5 fixed parameters

jasonwang01:
add lua for small B/W screen radio. Only save five fixed parameters infomation to avoid out of memory. They are BindPhrase/Mode/TxPwr/RxPwr/RxOutMode

### DIFF
--- a/lua/mLRS-bw.lua
+++ b/lua/mLRS-bw.lua
@@ -1,0 +1,952 @@
+--local toolName = "TNS|mLRS Configurator|TNE"
+----------------------------------------------------------------------
+-- Copyright (c) MLRS project
+-- GPL3
+-- https://www.gnu.org/licenses/gpl-3.0.de.html
+-- OlliW @ www.olliw.eu
+-- modify for B/W screen by Jason Wang
+----------------------------------------------------------------------
+-- Lua TOOLS script
+----------------------------------------------------------------------
+-- copy script to SCRIPTS\TOOLS folder on OpenTx SD card
+-- works with mLRS v0.1.13 and later, mOTX v33
+
+local version = '2023-04-28.00'
+
+local CUSTOM_COLOR = INVERS
+local TEXT_COLOR = 0 
+local MENU_TITLE_COLOR = INVERS
+local TEXT_DISABLE_COLOR = 0
+
+local function liney(line)
+    return 8*line;
+end
+
+-----------------------------custom param list ---------------------------
+--------param (max 5)
+local BindPhrase_idx = 0 -- idxes of options on main page
+local Mode_idx = 1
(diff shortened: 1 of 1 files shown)


olliw42:
@jasonwang01 
this is absolutely outstanding. awesome.

I'm on vaccation currently and can't really look at it, but few comments come to mind
- I'm surprised that you only keep bind phrase and mode, and not also teh frequency band (i.e. all three common parameters on the "main" page). All three parameters are relevant for establishing a connection.
- Generally, the indices for the parameters can change over time! So, if it's not possible to make the script independent, one probably should put them into an easy to find and acess location in the script, and define them there.
- Some of the functions&variables are the same as in the color script, while some others are obviously different. I guess it would be usefull to make that better visiable. For instance, one could add a _bw to thos functions which have been changed or are specific for bw.
- no need for by_index parameter upload? or still desirable?

just nitpicking :)

maybe you could post your script in the rcgroups and/or discord threads, to seek testers. I recall some users indicated they could test.

again, great work!

jasonwang01:
i don't know how to optimize lua to avoid the out of memory, it's hard to do :(
so just ignore storing many parameters except those five. even can't add a bootloader option because radio throw out not enough memory
it's a temporary solution, is better to post in the rcgroups/discord instead merge in github,  sorry i hurriedly share it :)
i'm reading elrs lua for get some ideas, maybe need reduce the parameter storing by multipart load. for example, read main list/tx list/rx list when entering their page

have a wonderful vacation

olliw42:
@jasonwang01 
don't appologize
it is fantastic that you make his effort, and that you share your result here !!
Even if you may not yet find it perfect, it will give ideas and trigger further work. It is just great to see.

For that reason I also reopen this PR, so that it doesn't disappear and stays on top for us to discuss.

Probably one indeed needs what you suggested elsewhere, namely this param_by_index function. I will see if I can add it next week, so to not block the bw efforts.

Are you running this on EdgeTx, or OpenTx?

olliw42:
@jasonwang01 
there is a positive report on your script by VRqualer here: https://www.rcgroups.com/forums/showpost.php?p=51056625&postcount=1732

olliw42:
@jasonwang01 
I have added support for calling a parameter by index. In the latest commit you also find an example lua script where this mechanism is used instead of the RequestList approach. It shoudl help you to realize how the byIndex mechanism is used.

hopefully this helps you to get better along with a bw script

note: the code can be much simplified, in the example code I tried to do just minor changes, I uesd it to test/confirm that the new mechanism is working
note 2: a simplification is alos possible by stripping out all this mbridge stuff, on bw radios only crsf is a viable option. This may also save few bytes.

PR #102, opened 2023-08-04
FLRC Bind updates

jlpoltrack:
Untested at the moment, my Rx hardware don't have bind buttons.


olliw42:
looks like the right idea
but can't be correct as is since SX1280_xxx is SX128x only ...
will test

jlpoltrack:
Might need to call 'setStandby' per the datasheet:

image

SetToIdle does a SetFs:

image

olliw42:
yes, this seems to make it work.
merging with followup
THX!

PR #103, opened 2023-08-05
Add Linux support to some python scripts

brad112358:
Note that I commented out the code which copies the license.md files.  The case of these file names is inconsistent in the source path, so this code didn't always work on Linux.  This is clearly not needed for building so I'm not sure why you bothered when you didn't copy other documentation files.  Let me know if you think it is needed.


tools/run_copy_st_drivers.py
```diff
@@ -22,15 +22,22 @@
...
+    else:
+        os.system('md "'+directory+'"')
 
+            
 def copy_files_in_dir(target, source):
     dirlist = os.listdir(source)
     for f in dirlist:
         if os.path.isfile(os.path.join(source,f)):
-            shutil.copy(os.path.join(source,f), target)
+             shutil.copy(os.path.join(source,f), target)
```
olliw42, line 40: indention error, tab should be 4 spaces

tools/run_copy_st_drivers.py
```diff
@@ -149,7 +156,7 @@ def copy_cmsis_driver(folder, chip, clean=False, silent=False):
         copy_dir(target, source)
 
         # copy licence
-        shutil.copy(os.path.join(source,'..','LICENSE.md'), os.path.join(target,'..'))
+        #shutil.copy(os.path.join(source,'..','LICENSE.md'), os.path.join(target,'..'))
```
olliw42, line 159: pl copy licence

tools/run_copy_st_drivers.py
```diff
@@ -226,7 +233,7 @@ def copy_hal_driver(folder, chip, clean=False, silent=False):
             copy_wexclude(target, source, f, files_to_exclude)
 
         # copy licence
-        shutil.copy(os.path.join(source,'..','LICENSE.md'), os.path.join(target,'..'))
+        #shutil.copy(os.path.join(source,'..','LICENSE.md'), os.path.join(target,'..'))
```
olliw42, line 236: pl copy licence

tools/run_make_firmwares3.py
```diff
@@ -30,7 +30,7 @@ def findSTM32CubeIDEGnuTools(search_root):
     st_cubeide_dir = ''
     st_cubeide_ver_nr = 0
     for file in os.listdir(search_root):
-        if 'STM32CubeIDE' in file:
+        if 'stm32cubeide' in file or 'STM32CubeIDE' in file:
```
olliw42, line 33: maybe better something line `if 'stm32cubeide' in file.lower(): # to make it work on several os`

tools/run_make_firmwares3.py
```diff
@@ -41,16 +41,26 @@ def findSTM32CubeIDEGnuTools(search_root):
...
+                gnu_dir = dirpath 
+        else:
+            if 'mcu.externaltools.gnu-tools-for-stm32' in dirpath and 'win32' in dirpath:
+                gnu_dir = dirpath 
     
     return st_dir, gnu_dir    
 
-ST_DIR,GNU_DIR = findSTM32CubeIDEGnuTools(os.path.join("C:/",'ST'))
+if os.name == 'posix':
+    ST_DIR,GNU_DIR = findSTM32CubeIDEGnuTools(os.path.join("/opt",'st'))
```
olliw42, line 61: ????
that's just you or everyone?

olliw42:
many thx for this!
I think few points need to be looked at.

- first off, can I assume this was also tested on Win, or should I do?
- run_make_firmware and run_replace_xxx are empty. Pl correct this. (merging this PR will kill them, won't it?)
- run_copy_stdrivers: The issue is that the folder should not be empty since STM32CubeIDE drops it then and a user may run into hazzles. Hence some file needs to be there, and my not so random random pick was LICENSE.md. So, pl find a way that it is coped. I don't understand what "_The case of these file names is inconsistent in the source path, so this code didn't always work on Linux_" means, but if it means what it appears to say it must be s strange Lunix attitude to change the case of files.
- run_make_firmware3: a good number of the `if os.name == 'posix':` come to separate the mkdir and rm operations. I think one could do that better by defining somewhere in the top helper functions to md and rm, and call them later.
- run_make_firmware3: the path things I find very strange. Are these the canonical paths as any user would get them when they install STM32CubeIDE on thier linux system, or are these the paths which you made them to be on your system?


also some inline comments 

:)

brad112358:
> first off, can I assume this was also tested on Win, or should I do?

I should have mentioned no testing on Windows.  Sorry, I only have one computer with windows installed, and it runs Linux 99% of the time.  It also has no development tools installed as I  don't do development on Windows.

> run_make_firmware and run_replace_xxx are empty. Pl correct this. (merging this PR will kill them, won't it?)

No, I've only changed the modes on these files to make them executable on Linux.  Github is misleading in this case. They are not empty in my branch.

> I don't understand what "The case of these file names is inconsistent in the source path, so this code didn't always work on Linux" means

Some of the source paths you are copying have a file named "License.md" and some are named "LICENSE.md".  By default Windows doesn't care about the case in file names, but Linux does.

In the places where I commented out the copy of these files, the folders where you are copying the license files are not completely empty.  They obviously contain a folder since you are copying the files to a path that ends in "..".  Are you sure you really need this file in these two cases?

> run_make_firmware3: the path things I find very strange. Are these the canonical paths as any user would get them when they install STM32CubeIDE on thier linux system, or are these the paths which you made them to be on your system?

The IDE install package from ST installs itself in /opt/st.  This is a common convention for third party Linux packages which don't use a specific Linux distribution's native package format and thus, aren't managed by the Linux system.  I find "C:" strange also.  :)

tools/run_make_firmwares3.py
brad112358: This is the path that the Linux version of the IDE package uses.

olliw42:
> > first off, can I assume this was also tested on Win, or should I do?
> 
> I should have mentioned no testing on Windows. Sorry, I only have one computer with windows installed, and it runs Linux 99% of the time. It also has no development tools installed as I don't do development on Windows.

ok, no problem. I will then.

> > run_make_firmware and run_replace_xxx are empty. Pl correct this. (merging this PR will kill them, won't it?)
> 
> No, I've only changed the modes on these files to make them executable on Linux. Github is misleading in this case. They are not empty in my branch.

not clear to me what was then changed. But, ok.

> > I don't understand what "The case of these file names is inconsistent in the source path, so this code didn't always work on Linux" means
> 
> Some of the source paths you are copying have a file named "License.md" and some are named "LICENSE.md". By default Windows doesn't care about the case in file names, but Linux does.

ah oh ah

> In the places where I commented out the copy of these files, the folders where you are copying the license files are not completely empty. They obviously contain a folder since you are copying the files to a path that ends in "..". Are you sure you really need this file in these two cases?

you are right, I confused with the copy of LICENSE.txt in do_for_each_target(), which I found to be needed

anyway, I would prefer to copy them, for licence sakes correctness, so it is impossible to copy with Lunix, pl keep the lines for Win

> > run_make_firmware3: the path things I find very strange. Are these the canonical paths as any user would get them when they install STM32CubeIDE on thier linux system, or are these the paths which you made them to be on your system?
> 
> The IDE install package from ST installs itself in /opt/st. This is a common convention for third party Linux packages which don't use a specific Linux distribution's native package format and thus, aren't managed by the Linux system. I find "C:" strange also. :)

ok. I just wanted to be sure these two path manglings you've added are not just your special locations :)

Overall, I would wish there would be a simpler and less verbose and spilling way to make the scripts linux&win compatible. Maybe one could have one place in the top which separates out the two cases, but I guess for now it's what it is.

So, two points left:
- could you pl check if you can find a not too clunky line to copy the license mds, or to keep the line only for win
- could you pl see if it makes the firmware3 code look simpler by introducing md & mkdir functions

:)

brad112358:
> could you pl check if you can find a not too clunky line to copy the license mds, or to keep the line only for win

If you insist on having these files, I guess we can test which specific version of the file exists before copying it.  Is that good enough?

> could you pl see if it makes the firmware3 code look simpler by introducing md & mkdir function

Perhaps I don't understand what you are asking for here.  At first I thought you didn't like the duplication of the code in create_clean_dir(), so I simplified it in my last commit by calling the existing functions.  Is this not simple enough?

olliw42:
> If you insist on having these files

for win I would insist, for linux I don't care :)

> , I guess we can test which specific version of the file exists before copying it. Is that good enough?

I'm surprised that linux hasn't some kind of wildcard or whatever to make copy case insensitive, but, hey, what do I know about linux LOL

as said, if you find the resulting code to complex, just do the copy only for win

> > could you pl see if it makes the firmware3 code look simpler by introducing md & mkdir function
> 
> Perhaps I don't understand what you are asking for here.

you did already (I missed to see you did some changes already before I commented) :)

brad112358:
> I'm surprised that linux hasn't some kind of wildcard or whatever to make copy case insensitive

There are ways of doing this from the shell on Linux by setting the nocaseglob option, but it's a little ugly and you gave me an easy way out so I'll take it.  :)

olliw42:
THX!
I will merge now.
There might be some follow up changes, to see if one can't separate the difference better. Not sure yet.
Anyways. GREAT and THX again.

brad112358:
Thanks!

olliw42:
@brad112358 

I made some little changes, mostly cosmetics. Can you pl quickly double check if the ymake sense and if it's still all good for you?

all thanks to you :)

PR #105, opened 2023-08-08
rx, transmit diversity

jlpoltrack:
Works on the bench - but not sure if I'm missing anything here :)

### DIFF
--- a/mLRS/CommonRx/mlrs-rx.cpp
+++ b/mLRS/CommonRx/mlrs-rx.cpp
@@ -659,8 +659,17 @@ RESTARTCONTROLLER:
         }break;
 
     case LINK_STATE_TRANSMIT: {
-        // TODO: transmit antenna diversity
-        do_transmit((USE_ANTENNA1) ? ANTENNA_1 : ANTENNA_2);
+    	if (USE_ANTENNA1 && USE_ANTENNA2) {  // check for diversity
+    		if (stats.last_rssi1 >= stats.last_rssi2) {  // check which antenna had the better rssi
+    			do_transmit(USE_ANTENNA1);
+    		}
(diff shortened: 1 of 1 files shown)


olliw42:
ah, very interesting. I just happen to have started to continue my work on this, albeit I'm focusing for the moment on the tx side. I've just made me a logger to log the complete real-time transmitter stats data, to work out a scheme.

I can't claim for sure, but I think this is too simplistic. Especially I'm concerned about what's happening near connection loss (i.e. in the presence of substantial frame losses) as well as when disconnected. I also somehow feel that there should be some time constant, e.g. from some filtering, as well as some hysteresis. The point is that the momentary interference or loss of packet on a frequency does not typically mean that the over-the-air channel properties have changed, e.g. becasue the vehicle has changed it position oin space. So, the simplistic scheme may actually most of the time jump to the bad antenna. Just my thoughts on the matter.

BTW, I'm going to do my experiments for tezh tx transmit diversity with a 2dBi omni antenna on one port and a 7 dBi directional antenna on the other port. Would be a cool general configuration, I guess. (doesn't make sense on rx side, of course)

jlpoltrack:
Fair enough - will be interesting to see your findings.  I will try to do at least 1 flight with this basic implementation and share what I see.

Hardware wise - I seem to be moving towards single antenna (directional, e.g. moxon) Tx and diversity on the Rx side.

olliw42:
any technical reason for closing which would be worth knowing?

jlpoltrack:
No technical reason, just tidying :)

olliw42:
ha, ok, thx!

you have seen this PR https://github.com/olliw42/mLRS/pull/108 ?

PR #108, opened 2023-08-15
Transmit diversity

olliw42:
this PR is just so that those who want to test can do test :)

@jlpoltrack you may find this interesting

with regards to testing, it should be noted that it is somewhat difficult to actually "see" if it is working correctly, since as a matter of fact always one antenna will be choosen for transmission, so the unit will always transmit ...


olliw42:
transmit diversity is now in main
closing

PR #110, opened 2023-08-18
Added std. "ESP32 Dev Module"

rotorman:
Added std. "ESP32 Dev Module" (e.g. NodeMCU ESP32-WROOM-32) as a new option. Including also an inverted serial version for usage e.g. with FrSky R9M.

According to Discord user hooman, this code works with his R9M and NodeMCU ESP32-WROOM-32.

Update: tested myself with ESP32-DevKitC V4 with ESP32-WROOM-32D as well.

### DIFF
--- a/esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
+++ b/esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
@@ -120,9 +120,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD 16 // = RX2
     #define SERIAL_TXD 17 // = TX2
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #undef USE_LED
 
@@ -139,9 +136,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD 16 // = RX2
     #define SERIAL_TXD 17 // = TX2
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #ifndef LED_IO
         #define LED_IO  2
@@ -177,9 +171,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD 3 // = RX
     #define SERIAL_TXD 1 // = TX
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #ifndef LED_IO
         #define LED_IO  13
@@ -199,9 +190,6 @@ GPIO15 = RTC_GPIO13
     
     #define SERIAL_RXD 18 // = RX1
     #define SERIAL_TXD 17 // = TX1
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #undef LED_IO
     #define USE_LED
@@ -221,9 +209,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD 18 // = RX1
     #define SERIAL_TXD 19 // = TX1
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #undef LED_IO
     #define USE_LED
@@ -243,9 +228,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD  32 // = RX1
     #define SERIAL_TXD  33 // = TX1
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #undef LED_IO
     #define USE_LED
@@ -265,9 +247,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD  1 // = RX1
     #define SERIAL_TXD  0 // = TX1
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #undef LED_IO
     #define USE_LED
@@ -287,9 +266,6 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD 32 // = RX1
     #define SERIAL_TXD 26 // = TX1
-  #if defined USE_SERIAL_INVERTED
-    #define SERIAL_INVERT true
-  #endif
 
     #undef LED_IO
     #define USE_LED
--- a/esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
+++ b/esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
@@ -18,11 +18,19 @@ IO9/IO10: U1RXD/U1TXD, is Serial1
 IO16/IO17: U2RXD/U2TXD, uses IO16/IO17 for internal flash, hence not available as serial
 
 ------------------------------
-NodeMCU ESP32-Wroom-32 or ESP32-DevKitC V4
+Espressif ESP32-DevKitC V4
+------------------------------
+board: ESP32 Dev Module
+https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-devkitc.html
+IO3/IO1: U0RXD/U0TXD, connected via usb-ttl adapter to USB port, is Serial, spits out lots of preamble at power up
+IO16/IO17: U2RXD/U2TXD, is Serial2
+
+
+------------------------------
+NodeMCU ESP32-Wroom-32
 ------------------------------
 board: ESP32 Dev Module
 https://esphome.io/devices/nodemcu_esp32.html
-https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-devkitc.html
 IO3/IO1: U0RXD/U0TXD, connected via usb-ttl adapter to USB port, is Serial, spits out lots of preamble at power up
 IO16/IO17: U2RXD/U2TXD, is Serial2
 
@@ -100,25 +108,8 @@ GPIO15 = RTC_GPIO13
 //-------------------------------------------------------
 // board details
 //-------------------------------------------------------
-//-- NodeMCU ESP32-Wroom-32 or ESP32-DevKitC V4
-#if defined MODULE_NODEMCU_ESP32_WROOM32
-    #ifndef ARDUINO_ESP32_DEV // ARDUINO_BOARD != ARDUINO_ESP32_DEV
-	      #error Select board ESP32 Dev Module!
-    #endif
-
-    #undef USE_SERIAL_DBG1
-    #undef USE_SERIAL1_DBG
-    #define USE_SERIAL2_DBG
-
-    // ESP32-DevKitC V4 does not have embedded LED
-    #ifndef LED_IO
-        #define LED_IO  2
-    #endif    
-    #define USE_LED
-
-
-//-- NodeMCU ESP32-Wroom-32 or ESP32-DevKitC V4 (inverted serial)
-#elif defined MODULE_NODEMCU_ESP32_WROOM32_SERIAL_INVERTED
+//-- Espressif ESP32-DevKitC V4
+#if defined MODULE_ESP32_WROOM32
     #ifndef ARDUINO_ESP32_DEV // ARDUINO_BOARD != ARDUINO_ESP32_DEV
 	      #error Select board ESP32 Dev Module!
     #endif
@@ -129,9 +120,29 @@ GPIO15 = RTC_GPIO13
 
     #define SERIAL_RXD 16 // = RX2
     #define SERIAL_TXD 17 // = TX2
+  #if defined USE_SERIAL_INVERTED
     #define SERIAL_INVERT true
+  #endif
 
-    // ESP32-DevKitC V4 does not have embedded LED
+    #undef USE_LED
+
+
+//-- NodeMCU ESP32-Wroom-32
+#elif defined MODULE_NODEMCU_ESP32_WROOM32
+    #ifndef ARDUINO_ESP32_DEV // ARDUINO_BOARD != ARDUINO_ESP32_DEV
+	      #error Select board ESP32 Dev Module!
+    #endif
+
+    #undef USE_SERIAL_DBG1
+    #undef USE_SERIAL1_DBG
+    #define USE_SERIAL2_DBG
+
+    #define SERIAL_RXD 16 // = RX2
+    #define SERIAL_TXD 17 // = TX2
+  #if defined USE_SERIAL_INVERTED
+    #define SERIAL_INVERT true
+  #endif
+
     #ifndef LED_IO
         #define LED_IO  2
     #endif    
@@ -164,6 +175,12 @@ GPIO15 = RTC_GPIO13
     #undef USE_SERIAL1_DBG
     #undef USE_SERIAL2_DBG
 
+    #define SERIAL_RXD 3 // = RX
+    #define SERIAL_TXD 1 // = TX
+  #if defined USE_SERIAL_INVERTED
+    #define SERIAL_INVERT true
+  #endif
+
     #ifndef LED_IO
         #define LED_IO  13
     #endif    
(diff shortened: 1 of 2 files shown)


esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
```diff
@@ -192,7 +236,7 @@ GPIO15 = RTC_GPIO13
 //-- M5Stack ATOM Lite
 #elif defined MODULE_M5STACK_ATOM_LITE
     #ifndef ARDUINO_M5Stack_ATOM // ARDUINO_BOARD != ARDUINO_M5Stack_ATOM
-	      #error Select board ARDUINO_M5Stack_ATOM!
+	      #error Select board M5Stack-ATOM!
```
olliw42, line 239: this appears to be the correction of a bug unrelated to this PR. Maybe separate out. It seems to me that there are more of this kind of bugs, e.g. in line 141. We probably want to go over all the #error and get them right in one stroke, and a separate PR.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -42,6 +44,8 @@ for more details on the boards see mlrs-wifi-bridge-boards.h
 // Board
 // un-comment what you want
 //#define MODULE_GENERIC
+//#define MODULE_ESP32_DEV
+//#define MODULE_ESP32_DEV_INV
```
olliw42, line 48: pl give the flag a more descritive name. With the number of flags more and more growing it becomes more and more difficult to follow through. 
Maybe MODULE_ESP32_DEV_SERIAL_INVERTED

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -80,6 +84,7 @@ int wifi_channel = 13;
 // comment all for default behavior, which is using only Serial port
 //#define USE_SERIAL_DBG1 // use Serial for communication and flashing, and Serial1 for debug output
 //#define USE_SERIAL1_DBG // use Serial1 for communication, and Serial for debug output and flashing
+//#define USE_SERIAL2_DBG // use Serial2 for communication, and Serial for debug output and flashing
```
olliw42, line 87: adding this option should mean that in the boards.h file there should be also plenty of #undef USE_SERIAL2_DBG in all the other modules, shouldn't it

olliw42:
first, many thx
on a first quick inspection, I have three comments

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -42,6 +44,8 @@ for more details on the boards see mlrs-wifi-bridge-boards.h
 // Board
 // un-comment what you want
 //#define MODULE_GENERIC
+//#define MODULE_ESP32_DEV
```
olliw42, line 47: ok, another one: 
This is for the  NodeMCU ESP32-Wroom-32, so the flag should be called  MODULE_NODEMCU_ESP32_WROOM32

the idea of this pattern is to make it somewhat easy for a user to figure out which flag she/he needs to uncomment

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
rotorman: Done.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
rotorman: Done.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
rotorman: Done.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
rotorman: Done.

olliw42:
this lgtm, but let it rest for a bit before I merge, in case more thoughts come up. 

very nice

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
```diff
@@ -91,14 +99,51 @@ GPIO15 = RTC_GPIO13
 //-------------------------------------------------------
 // board details
 //-------------------------------------------------------
+//-- ESP32 Dev Module
```
olliw42, line 102: should be NodeMCU ESP32-Wroom-32
dito in line 118

rotorman: Done, added also ESP32-DevKitC V4 as this runs with the same code as well.

olliw42: since it is not hardware compatible I guess a separate entry for the  ESP32-DevKitC V4 would be nicer.

you probably are going to also add the inverted version ... I'm not sure I like that huge increase of cases ...

what do you think about this:

the two cases defined MODULE_NODEMCU_ESP32_WROOM32 and defined MODULE_NODEMCU_ESP32_WROOM32_SERIAL_INVERTED could be put into one common #if, and whithin it the case MODULE_NODEMCU_ESP32_WROOM32_SERIAL_INVERTED could set the SERIAL_INVERT true. Should make the whole section much more compact.

I'm still not sure we should add the inverted variant for each and every module type coming up in life ... for the R9M we did it as a curtosy to Brad and his particular mod, which is now also in the docs, simply becasue it really fits so neatly. But to do it for each and every ?!?!?!

Maybe we rather should provide an option USE_SERIAL_INVERTED, like we do with WIFI_PROTOCOL or WIFI_POWER ... I momentarily tend towards this ...

what do you say?

rotorman: Good idea. Please have a look at latest commit.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
```diff
@@ -17,6 +17,23 @@ IO3/IO1: U0RXD/U0TXD, connected via usb-ttl adapter to USB port, is Serial, spit
...
 
+------------------------------
+Espressif ESP32-DevKitC V4
+------------------------------
+board: ESP32 Dev Module
+https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-devkitc.html
+IO3/IO1: U0RXD/U0TXD, connected via usb-ttl adapter to USB port, is Serial, spits out lots of preamble at power up
+IO16/IO17: U2RXD/U2TXD, is Serial2
+
+
```
olliw42, line 28: line too much

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
```diff
@@ -91,14 +108,50 @@ GPIO15 = RTC_GPIO13
 //-------------------------------------------------------
 // board details
 //-------------------------------------------------------
+//-- Espressif ESP32-DevKitC V4
+#if defined MODULE_ESP32_WROOM32
```
olliw42, line 112: should be named MODULE_EPS32_DEVKITC_V4

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -14,6 +14,8 @@
 /*
 for more details on the boards see mlrs-wifi-bridge-boards.h
 
+- NodeMCU ESP32-Wroom-32
```
olliw42, line 17: missing
- Espressif ESP32-DevKitC V4
board: ESP32 dev Module

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -42,6 +44,8 @@ for more details on the boards see mlrs-wifi-bridge-boards.h
 // Board
 // un-comment what you want
 //#define MODULE_GENERIC
+//#define MODULE_ESP32_WROOM32
```
olliw42, line 47: should be MODULE_EPS32_DEVKITC_V4, shouldn't it

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -133,7 +140,7 @@ void setup()
     size_t rxbufsize = SERIAL.setRxBufferSize(2*1024); // must come before uart started, retuns 0 if it fails
     size_t txbufsize = SERIAL.setTxBufferSize(512); // must come before uart started, retuns 0 if it fails
 #ifdef SERIAL_RXD // if SERIAL_TXD is not defined the compiler will complain, so all good
-  #ifdef SERIAL_INVERT
+  #ifdef USE_SERIAL_INVERTED
```
olliw42, line 143: I need to think about this, but for the moment ok
will not compile, right, since SERIAL_INVERT isn't always defined

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
rotorman: Fixed.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge-boards.h
rotorman: Done.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
rotorman: Done.

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
rotorman: Done (although with correct *ESP32*).

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
rotorman: It does compile, please check.

olliw42: I'm just looking at the code, but could it do the correct thing when SERIAL_INVERT is not defined? (which it seems to me it isn't). Maybe it compiles because that field has a default?

rotorman: https://github.com/rotorman/mLRS/blob/9c5891f0937efc8e606da33ffba4c40e9eb97336/esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino#L144-L152

rotorman: Ah darn, now I get it, my bad.

olliw42: :)

rotorman: Fixed now, sorry my bad.

olliw42:
I guess I'm going to repeat myself

this lgtm, but let it rest for a bit before I merge, in case more thoughts come up.

very nice, MANY thx so far, really much appreciated

olliw42:
@brad112358 in case you have the chance, could you test this PR on your R9M ESP32 setup? I just like to be sure we haven't broken it.

IJdewangan:
i am getting an error while uploading code in NodeMcu can anyone help?
 Code in flash (default, ICACHE_FLASH_ATTR), used 232100 / 1048576 bytes (22%)
║   SEGMENT  BYTES    DESCRIPTION
╚══ IROM     232100   code in flash

rotorman:
Do you have ESP32 NodeMCU? Which board did you select in Arduino IDE? Should be Boards -> esp32 -> ESP32 Dev Module.

brad112358:
I'm not sure why we need the #undef USE_SERIAL... everywhere since they will be undefined by default.
Given we now have 3 options for USE_SERIAL..., would it make more sense to eliminate all 3 and just move the SERIAL and DBG defines into the board details  #if defined blocks?  Then the DBG_PRINT(x) can use the DBG macro and be conditional on it being defined.  This all results in a bit less code (especially with the #undef USER_SERIAL...) and also easily allows the remaining combinations of Serial, Serial1, ... for future boards.

brad112358:
Another thought.... 
The current PR seems to require a change to the R9 documentation we just updated to mention the need to define USE_SERIAL_INVERTED.  That's fine, but then we may as well remove the "for Frsky R9M" references everywhere.  Alternatively, we could add "#define USE_SERIAL_INVERTED" to the "For Frsky R9M" boards and move the #define for this down in the .ino file to after the #include "mlrs-wifi-bridge-boards.h"

olliw42:
the intention of the flags is to be able to define your properties entirely in the main sketch, i.e., the .ino file, if the pre-prepared modules don't work for you. Hence in the boards.h you cannot rely on what you think are "defaults". I would like to keep it that way.
yes, for the R9M variants, the flag in teh boards.h file needs to be changed as you indicate, pl see the open review which Risto hasn't yet accounted for.
We could change things for the R9M cases, but I am happy to have it easy for a user for these two specific boards.

brad112358:
Perhaps I'm missing something.  I don't see any review comments not marked resolved.

olliw42:
oh damed ... I did it again ... grrrrrrr
THX for mentioning!

esp/mlrs-wifi-bridge/mlrs-wifi-bridge.ino
```diff
@@ -50,6 +56,8 @@ for more details on the boards see mlrs-wifi-bridge-boards.h
 //#define MODULE_M5STAMP_PICO_FOR_FRSKY_R9M
 //#define MODULE_M5STACK_ATOM_LITE
 
+// Uncomment, if you need inverted serial
+#define USE_SERIAL_INVERTED
```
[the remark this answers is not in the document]
rotorman: I hope I fixed it in latest commit as requested, please check. Thx!

olliw42:
MANY THX
sorry for having forgotten to finalize the pending review :(
lgtm
I will wait a bit to see if there is some confirmation/disconfirmation from Brad if it still works for him.

brad112358:
Yes, I tested both M5STAMP_C3U_MATE_FOR_FRSKY_R9M and MODULE_M5STAMP_PICO_FOR_FRSKY_R9M and they worked as expected once I remembered that I needed to change the !#*@&% channel from 13 to something that works in North America.

I don't usually need to remember this since I'm stuck maintaining my own branch anyway (because I want to pack an entire MAVLink messages in a single UDP packet to allow Wireshark to decode it), but it might be nicer especially for new users not to have to mess with this.

Does using channel 13 still bring a significant benefit for mLRS users on 2.4 GHz?  Or do I remember seeing there is now a way to avoid an arbitrary WiFi channel?  If we don't want to change the default channel to something that works everywhere, I suppose I should update the R9 doc as I forgot to mention it there.

olliw42:
> Yes, I tested both M5STAMP_C3U_MATE_FOR_FRSKY_R9M and MODULE_M5STAMP_PICO_FOR_FRSKY_R9M and they worked as expected 

very cool. MAYN THX. So I will merge.

> Does using channel 13 still bring a significant benefit for mLRS users on 2.4 GHz? Or do I remember seeing there is now a way to avoid an arbitrary WiFi channel? If we don't want to change the default channel to something that works everywhere, I suppose I should update the R9 doc as I forgot to mention it there.

Yes, mLRS has the capability now to exclude specific bands, see https://github.com/olliw42/mLRS-docu/blob/master/docs/PARAMETERS.md#bind-phrase. This is - at least for me, unfortunately no reports - very situation dependent however, and the "best" needs to be determined by trial and error. I e.g. exclude #6. And it is not a 100% perfect solution anyway, I believe because of receiver desensitation, it just mitigates the issue significantly. 

Anyway, if 13 is not an option in the US, we should choose something which is. 13 was meanigfull since it's really at the end of the fhss table. Moreover, for all those not on 2.4 GHz it doesn't matter, so one could argue that the 2.4 GHz folks should carry the burden. Any suggestion instead? Pl raise a PR.

olliw42:
@rotorman @brad112358 MAY THX for this. I'm merging.

PR #111, opened 2023-08-19
fix: Missing defines and paths to build in debug mode for FRM303 TX

rotorman:


olliw42:
many thx for this.

I guess I will not merge this however.

- it would have to be done consistently for all targets.
- it opens a completely new can of worms which I simply don't have the capacity to deal with.
- I also think it may too easily confuse new users as they might build in debug and just not realize the mistake, since "it works" ...
- you maybe may understand that the recent FRM303 sport pin case furthered my opinion that SWD debuging is not soo valuable and rarely very helpfull for the tricky issues in embeeded systems. The cases where it is usefull, one almost always can track well by simple debug prints.

The info in here is however very valuable. I think it would be well and better placed in the docs somewhere.

As a general comment, pl note that while a number of individuals have made some really remarkable code contributions to mLRS, the overall the dev base is enourmously thin. The code and project is already quite diffragmented and there is a huge tendency recently, explosion like, for further diffragmentation, to a point that the project is endangered to become unmanagable. I can't see how supporting now also debug techniques is feasible. Just to give a hint for the general hesitancy on my side to go with certain things. 

OK?

olliw42:
what I was trying to say that debug isn't configured on purpose, for the reasons listed
:)

olliw42:
suggest to add the procedures to the documentations, https://github.com/olliw42/mLRS-docu/issues/79

closing

PR #112, opened 2023-08-19
Add URL to ESP32 WiFi Power options list

rotorman:


olliw42:
MANY thx.
the problem with this sort of links is that they are volatile and not static in time. I tried to achieve the same by adding the comment "// Note: In order to find the possible options, right click on WIFI_POWER_19_5dBm and choose "Go To Definiton""  in commit https://github.com/olliw42/mLRS/commit/7c04aafae52292c8425a2dc38e8fd292f2fd4d79.
Don't hesitate to raise another PR if this isn't suitable.
Closing.

PR #113, opened 2023-08-19
Provide an option to lower CPU frequency for M5Stack ATOM Lite.

rotorman:
Keeps ESP32 CPU running cooler and taking less energy. Experimentally found out that 160 MHz is still enough at least for UDP connectivity. Lowering further down to 80 MHz, did not get nice results anymore.

### DIFF
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge-boards.h
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge-boards.h
@@ -277,6 +277,8 @@ GPIO15 = RTC_GPIO13
     #define NUMPIXELS  1
     #define PIN_NEOPIXEL  27
 
+    #define CPUFREQMHZ 160 // Options 80, 160, 240 MHz
+
 
 //-- Generic
 #else
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
@@ -187,6 +187,9 @@ void serialFlushRx(void)
 
 void setup()
 {
+#ifdef CPUFREQMHZ
+    setCpuFrequencyMhz(CPUFREQMHZ);
+#endif
     led_init();
     dbg_init();
     delay(500);


olliw42:
interesting.
are there any data as regards how effective this really is? I find it frankly a bit difficult to imagine that it should do a lot.

vrquaeler:
Hi, I do very similar things (like clocking down the MCU) for the "ESP-Now" flashed modules. It just works. The ESP's run cooler. Did also current measurements on the ESP modules, but can't remember the numbers.

jlpoltrack:
On my ESP-Now setup I run 80 MHz no problem - I can run some high speed captures if there's interest.   Here's some numbers pulled from Espressif forum.  Looks like moving to 160 MHz saves ~20 mA.  The ESP32 dev boards that I have usually have are powered from 5 V linear regulators so would be 100 mW in savings.

image

https://esp32.com/viewtopic.php?t=2662

rotorman:
Hmm... was not my intention to close this...

olliw42:
hahaha ... yuchee ... I'm not the only one to whom this happens :D:D:D

rotorman:
I measured the current draw of an Espressif ESP32-DevKitC V4 running `mlrs-wireless-bridge`. The results:

TCP mode (while client is connected to it, no data flowing): 240 MHz - 139 mA; 160 MHz - 126 mA; 80 MHz - 113 mA
UDP mode (while client is connected to it, no data flowing): 240 MHz - 139 mA; 160 MHz - 126 mA; 80 MHz - 113 mA
UDP client mode (not connected to AP): 240 MHz -  139 mA; 160 MHz - 114 mA; 80 MHz - 107 mA
UDP client mode (connected to ESP8266 running mLRS UDP AP, no data flowing): 240 MHz - 65 mA; 160 MHz - 53 mA; 80 MHz - 39 mA
Bluetooth mode (paired state, no data flowing): 240 MHz - 73 mA; 160 MHz - 60 mA; 80 MHz - 48 mA

From the results above I can conclude that having a lower clock, does lower the power consumption of an ESP32 board. Especially for modules in closed containers, such as M5Stack ATOM Lite at the back of Flysky FRM303 micro JR bay, as in my case, this will be IMO beneficial.

olliw42:
cool to see some data.
I think the question however is not so much if it does draw less power ... I mean, it's a priori clear that it will draw lower power the lower the frequency, that's just a basic principle for this sort of electronics ... I think this is hardly disputed. To me the question rather would be if it's mandatory in the sense that it has a substantial effect, e.g., on the temperature, so that one just should do it to preserve the well-functioning, or if it is just a yeah it's better but one can live well also without it.

rotorman:
The question is other way around - is it mandatory to have a higher clock frequency paired with higher energy consumption.

olliw42:
I tend to go with the default and to divert only if needed, not the other way around :)
for the STM32s we also just go with the max frequency
for as long as it doesn't hurt, it doesn't hurt ...
:)

olliw42:
appears not to be super relevant
closing

PR #115, opened 2023-08-22
Adds Bluetooth connectivity option to mLRS Bridge.

rotorman:
This PR changes mLRS WiFi bridge to mLRS WiFi & Bluetooth bridge -> thus a new name `mLRS Wireless Bridge`.

With define `WIRELESS_PROTOCOL` either WiFi TCP (=0), WiFi UDP (=1) or Bluetooth (=2) can be selected.

Quick test with an Espressif ESP32-DevKitC V4 gave promising results. Would be sweet to get some further tests with other hw before merging.

**Update:** longer test with M5Stack ATOM Lite shows that this works nicely.

### DIFF
--- a/esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
+++ b/esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
@@ -239,16 +239,15 @@ void loop()
         if (led_state) led_on(is_connected); else led_off();
     }
 
-    //-- here comes the core code, handle wifi connection and do the bridge
+    //-- here comes the core code, handle WiFi or Bluetooth connection and do the bridge
 
-    const int bufsize = 256;
-    uint8_t buf[bufsize]; // working buffer
+    uint8_t buf[256]; // working buffer
 
 #if (WIFI_nBLUETOOTH == 0) // Bluetooth
     uint8_t idx = 0;
     while (SerialBT.available()) {
       buf[idx++] = SerialBT.read();
-      if (idx >= bufsize) break;
+      if (idx >= sizeof(buf)) break;
     }
     if (idx > 0)
     {
--- a/esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
+++ b/esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
@@ -57,16 +57,9 @@ List of supported modules, and board which need to be selected
 // (for the generic module also SERIAL_RXD and SERIAL_TXD need to be defined)
 //#define USE_SERIAL_INVERTED
 
-#define WIFI_nBLUETOOTH 1 // Bluetooth = 0, WiFi = 1
+#define PROTOCOL_TYPE 1 // 0 = WiFi TCP, 1 = WiFi UDP, 2 = Bluetooth
 
-#if (WIFI_nBLUETOOTH == 0) // Bluetooth
-  String localName = "mLRSBTbridge"; // Bluetooth name
-#endif
-
-#if (WIFI_nBLUETOOTH == 1) // WiFi
-  // Wifi Protocol 0 = TCP, 1 = UDP
-  #define WIFI_PROTOCOL  1
-
+#if ((PROTOCOL_TYPE == 0) || (PROTOCOL_TYPE == 1)) // WiFi
   // Wifi credentials
   String ssid = "mLRS AP"; // Wifi name
   String password = ""; // "thisisgreat"; // WiFi password, "" makes it an open AP
@@ -75,22 +68,22 @@ List of supported modules, and board which need to be selected
 
   int port_tcp = 5760; // connect to this port per TCP // MissionPlanner default is 5760
   int port_udp = 14550; // connect to this port per UDP // MissionPlanner default is 14550
-#endif
 
+  // WiFi channel
+  // 1 is the default, 13 (2461-2483 MHz) has the least overlap with mLRS 2.4 GHz frequencies.
+  // Note: Channel 13 is generally not available in the US, where 11 is the maximum.
+  int wifi_channel = 13;
+
+  // WiFi power
+  // comment out for default setting
+  // Note: In order to find the possible options, right click on WIFI_POWER_19_5dBm and choose "Go To Definiton"
+  #define WIFI_POWER  WIFI_POWER_2dBm // WIFI_POWER_MINUS_1dBm is the lowest possible, WIFI_POWER_19_5dBm is the max
+#elif (PROTOCOL_TYPE == 2) // Bluetooth
+  String localName = "mLRSBTbridge"; // Bluetooth name
+#endif
 
 // baudrate
 int baudrate = 115200;
-
-// WiFi channel
-// 1 is the default, 13 (2461-2483 MHz) has the least overlap with mLRS 2.4 GHz frequencies.
-// Note: Channel 13 is generally not available in the US, where 11 is the maximum.
-int wifi_channel = 13;
-
-// WiFi power
-// comment out for default setting
-// Note: In order to find the possible options, right click on WIFI_POWER_19_5dBm and choose "Go To Definiton"
-#define WIFI_POWER  WIFI_POWER_2dBm // WIFI_POWER_MINUS_1dBm is the lowest possible, WIFI_POWER_19_5dBm is the max
-
 
 // serial port usage (only effective for the generic module)
 // comment all for default behavior, which is using only Serial port
@@ -107,14 +100,11 @@ int wifi_channel = 13;
 // Includes
 //-------------------------------------------------------
 
-#if (WIFI_nBLUETOOTH == 0) // Bluetooth
+#if ((PROTOCOL_TYPE == 0) || (PROTOCOL_TYPE == 1)) // WiFi
+  #include <WiFi.h>
+#elif (PROTOCOL_TYPE == 2) // Bluetooth
   #include <BluetoothSerial.h>
 #endif
-
-#if (WIFI_nBLUETOOTH == 1) // WiFi
-  #include <WiFi.h>
-#endif
-
 
 
 //-------------------------------------------------------
@@ -128,20 +118,18 @@ int wifi_channel = 13;
 // Internals
 //-------------------------------------------------------
 
-#if (WIFI_nBLUETOOTH == 0) // Bluetooth
-  BluetoothSerial SerialBT;
-#endif
-
-#if (WIFI_nBLUETOOTH == 1) // WiFi
+#if ((PROTOCOL_TYPE == 0) || (PROTOCOL_TYPE == 1)) // WiFi
   IPAddress ip_udp(ip[0], ip[1], ip[2], ip[3]+1); // speculation: it seems that MissionPlanner wants it +1
   IPAddress netmask(255, 255, 255, 0);
-  #if WIFI_PROTOCOL == 1 // UDP
+  #if PROTOCOL_TYPE == 1 // UDP
     WiFiServer server(80);
     WiFiUDP udp;
   #else // TCP
     WiFiServer server(port_tcp);
     WiFiClient client;
   #endif
+#elif (PROTOCOL_TYPE == 2) // Bluetooth
+  BluetoothSerial SerialBT;
 #endif
 
 bool led_state;
@@ -183,15 +171,11 @@ void setup()
     DBG_PRINTLN(rxbufsize);
     DBG_PRINTLN(txbufsize);
 
-#if (WIFI_nBLUETOOTH == 0) // Bluetooth
-    SerialBT.begin(localName);
-#endif
-
-#if (WIFI_nBLUETOOTH == 1) // WiFi
+#if ((PROTOCOL_TYPE == 0) || (PROTOCOL_TYPE == 1)) // WiFi
     // AP mode
-    //WiFi.mode(WIFI_AP); // seems not to be needed, done by WiFi.softAP()?
+    // WiFi.mode(WIFI_AP); // seems not to be needed, done by WiFi.softAP()?
     WiFi.softAPConfig(ip, ip, netmask);
-  #if WIFI_PROTOCOL == 1
+  #if (PROTOCOL_TYPE == 1)
     String ssid_full = ssid + " UDP";
   #else    
     String ssid_full = ssid + " TCP";
@@ -208,9 +192,11 @@ void setup()
   #ifdef WIFI_POWER
     WiFi.setTxPower(WIFI_POWER); // set WiFi power, AP or STA must have been started, returns false if it fails
   #endif    
-  #if WIFI_PROTOCOL == 1
+  #if (PROTOCOL_TYPE == 1)
     udp.begin(port_udp);
   #endif
+#elif (PROTOCOL_TYPE == 2) // Bluetooth
+    SerialBT.begin(localName);
 #endif
 
     led_tlast_ms = 0;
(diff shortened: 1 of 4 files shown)


esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
```diff
@@ -213,9 +241,37 @@ void loop()
 
     //-- here comes the core code, handle wifi connection and do the bridge
 
-    uint8_t buf[256]; // working buffer
+    const int bufsize = 256;
+    uint8_t buf[bufsize]; // working buffer
 
-#if WIFI_PROTOCOL == 1 // UDP
+#if (WIFI_nBLUETOOTH == 0) // Bluetooth
```
olliw42, line 247: could you pl move the BT to after the udp/tcp stuff?

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
```diff
@@ -213,9 +241,37 @@ void loop()
...
+    while (SerialBT.available()) {
+      buf[idx++] = SerialBT.read();
+      if (idx >= bufsize) break;
+    }
+    if (idx > 0)
+    {
+      is_connected = true;
+      is_connected_tlast_ms = millis();
+      SERIAL.write(buf, idx);
+    }
```
olliw42, line 258: why is this so different than for the UDP case? What am I missing when I think it could be basically similar?

SerialBT.available() returns the number of bytes, pl consider not using while but changing to 
int len = SerialBT.available() 
... ...

it seems SerialBT does not have a read() function usable as SerialBT.read(buf, sizeof(buf)), right?

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
rotorman: My reasoning was that option 0 comes before option 1. No?

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
```diff
@@ -213,9 +241,37 @@ void loop()
 
     //-- here comes the core code, handle wifi connection and do the bridge
 
-    uint8_t buf[256]; // working buffer
+    const int bufsize = 256;
```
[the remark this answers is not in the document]
rotorman: Fixed in latest commit.

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
olliw42: yo, sad. You can mimic it with a simple loop. So I still think that the code could look similar to UDP.

olliw42:
I understand the BT is not wifi argument, but as said, it IMHO introduces lots of complexity and make-it-difficult-for-users for what is really just semantics. I think the decission is simply, do we want to demonstrate that we understand that BT is not wifi or do we want new users with potentially low code skills enter.

the problem I have with things like PROTOCOL_TYPE is that they really say nothiong, I mean, is this MAVLink v1 vs v2? Or LTM vs transparent?

> The indentation of #if's IMO really helps the readability.

I will see when I see the whole thing :)

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
rotorman: Fixed in latest commit, as now using:
0: WiFi TCP
1: WiFi UDP
2: Bluetooth

rotorman:
Plz see last commit. If you are not cool with `PROTOCOL_TYPE` then how about `WIRELESS_PROTOCOL`?

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
```diff
@@ -75,23 +68,23 @@ List of supported modules, and board which need to be selected
...
+  // WiFi channel
+  // 1 is the default, 13 (2461-2483 MHz) has the least overlap with mLRS 2.4 GHz frequencies.
+  // Note: Channel 13 is generally not available in the US, where 11 is the maximum.
+  int wifi_channel = 13;
+
+  // WiFi power
+  // comment out for default setting
+  // Note: In order to find the possible options, right click on WIFI_POWER_19_5dBm and choose "Go To Definiton"
+  #define WIFI_POWER  WIFI_POWER_2dBm // WIFI_POWER_MINUS_1dBm is the lowest possible, WIFI_POWER_19_5dBm is the max
+#elif (PROTOCOL_TYPE == 2) // Bluetooth
```
olliw42: don't like PROTOCOL_TYPE :)
since you renamed files to wireless, WIRELESS_PROTOCOL suggests itself

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
rotorman: Cannot unfortunately follow your reasoning for a change here, as from Bluetooth read one can fish out only byte by byte as long as available returns true and you do want to stop when buffer is filled. Thus a while loop with a break IMO is the optimal solution. No?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -1,15 +1,15 @@
...
 // https://www.gnu.org/licenses/gpl-3.0.de.html
 //*******************************************************
-// Basic but effective & reliable transparent WiFi<->serial bridge
+// Basic but effective & reliable transparent WiFi and Bluetooth<->serial bridge
 //*******************************************************
 // 22. Aug. 2023
 //*********************************************************/
 // inspired by examples from Arduino
-// ArduinoIDE 2.0.3, esp32 by Espressif Systems 2.0.6
+// ArduinoIDE 2.1.1, esp32 by Espressif Systems 2.0.11
```
olliw42, line 12: is this just what you happen to have or are these in fact the minimal versions required?

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
```diff
@@ -75,23 +68,23 @@ List of supported modules, and board which need to be selected
...
+  String localName = "mLRSBTbridge"; // Bluetooth name
```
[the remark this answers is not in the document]
olliw42: one can't also set the power for BT, right? (class does not seem to have such function, but maybe it's possible otherwise?)

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
```diff
@@ -57,16 +57,9 @@ List of supported modules, and board which need to be selected
...
-#if (WIFI_nBLUETOOTH == 0) // Bluetooth
-  String localName = "mLRSBTbridge"; // Bluetooth name
-#endif
-
-#if (WIFI_nBLUETOOTH == 1) // WiFi
-  // Wifi Protocol 0 = TCP, 1 = UDP
-  #define WIFI_PROTOCOL  1
+#define PROTOCOL_TYPE 1 // 0 = WiFi TCP, 1 = WiFi UDP, 2 = Bluetooth
 
+#if ((PROTOCOL_TYPE == 0) || (PROTOCOL_TYPE == 1)) // WiFi
```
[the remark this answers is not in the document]
rotorman: Can you please elaborate more what you suggest. I did not yet fully understand your request.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -291,7 +291,7 @@ void loop()
         udp.endPacket();
     }
 #elif (PROTOCOL_TYPE == 2) // Bluetooth
-    uint8_t idx = 0;
+    uint16_t idx = 0;
```
olliw42, line 294: make it int, this seems to be a preferred practice in Arduino. Also, some functions return tsize_t which is int.

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
rotorman: Done in latest commit.

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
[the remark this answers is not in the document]
olliw42: it is not need to enclose the variables and defines by this #if here
the compiler will either take care or we will just not care the minimal waste
the user config section will look cleaner without these #if's
I rather would add a more visible comments thing to see better where the BT stuff sits

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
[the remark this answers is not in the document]
rotorman: Done.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: This is what I used to test.

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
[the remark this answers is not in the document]
olliw42: nearly :)
bluetooth_name ? bluetooth_device_name? instead of localName?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
olliw42: ok, fair enough. Thx for the clarification. I will test too, and when we will see :)

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
olliw42: it might be that one actually has to do if (len > 0) (I think some available() can also return -1)
probably we should to be on the safe side

rotorman:
[setPin](https://github.com/espressif/arduino-esp32/blob/099b432d10fb4ca1529c52241bcadcb8a4386f17/libraries/BluetoothSerial/src/BluetoothSerial.cpp#L898-L919) does not seem to work for Bluetooth device in client mode. At least I tried various combinations, and got never asked for PIN to connect to. If you have an idea, I am open to hear it.

What I did was:
```c++
#define USE_BLUETOOTH_PIN
const char *bluetoothPin = "1234";
```
and
```c++
SerialBT.begin(bluetooth_device_name);
#if defined(USE_BLUETOOTH_PIN)
  SerialBT.setPin(bluetoothPin);
#endif
```
with no luck.

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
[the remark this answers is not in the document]
rotorman: OK, now I see what you were after. Please see latest commit if that was what you meant.

olliw42:
> [setPin](https://github.com/espressif/arduino-esp32/blob/099b432d10fb4ca1529c52241bcadcb8a4386f17/libraries/BluetoothSerial/src/BluetoothSerial.cpp#L898-L919) does not seem to work for Bluetooth device in client mode. 

I haven't yet tried my self, but what you tried looks identical to what common tutotials say to do ... sad. I have no better idea unfortunately.

esp/mlrs-wifi-bt-bridge/mlrs-wifi-bt-bridge.ino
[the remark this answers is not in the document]
rotorman: Regarding Bluetooth power - no API call for that unforunately via BluetoothSerial: https://github.com/espressif/arduino-esp32/blob/master/libraries/BluetoothSerial/src/BluetoothSerial.h

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -57,10 +57,11 @@ List of supported modules, and board which need to be selected
...
 //#define USE_SERIAL_INVERTED
 
-// Wifi Protocol 0 = TCP, 1 = UDP
-#define WIFI_PROTOCOL  1
+#define WIRELESS_PROTOCOL 1 // 0 = WiFi TCP, 1 = WiFi UDP, 2 = Bluetooth
 
-// Wifi credentials
+
+//**********************//
+//*** Wi-Fi settings ***//
```
olliw42, line 64: WiFi not Wi-Fi
we use the former througout :)

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -83,6 +81,17 @@ int wifi_channel = 13;
 #define WIFI_POWER  WIFI_POWER_2dBm // WIFI_POWER_MINUS_1dBm is the lowest possible, WIFI_POWER_19_5dBm is the max
 
 
+//**************************//
+//*** Bluetooth settings ***//
+String bluetooth_device_name = "mLRS BT"; // Bluetooth device name
+
+
+//************************//
+//*** Generic settings ***//
```
olliw42, line 90: consider "General" instead of Generic

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -279,7 +275,49 @@ void loop()
...
+    int avail = SERIAL.available();
+    if (avail <= 0) {
+        serial_data_received_tfirst_ms = tnow_ms;      
+    } else {
+      if ((tnow_ms - serial_data_received_tfirst_ms) > 10 || avail > 128) { // 10 ms at 57600 bps corresponds to 57 bytes, no chance for 128 bytes
+        serial_data_received_tfirst_ms = tnow_ms;
+        int len = SERIAL.read(buf, sizeof(buf));
+        SerialBT.write(buf, len);
+      }	
+    }
```
olliw42, line 320: there is one level of {} too much
can be
```
} else
if () {
}
```

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -279,7 +275,49 @@ void loop()
...
+
+        int len = SERIAL.read(buf, sizeof(buf));
+        udp.beginPacket(ip_udp, port_udp);
+        udp.write(buf, len);
+        udp.endPacket();
+    }
+#elif (WIRELESS_PROTOCOL == 2) // Bluetooth
+    int len = SerialBT.available();
+    if (len > 0) {
+      if (len > sizeof(buf)) len = sizeof(buf);
```
olliw42, line 303: indention is 4 spaces

olliw42:
this really lgtm now.
just few very tiny things

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: Done.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: Done.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: This is intentional, as without explicit { } this is error prone according to previous experience, when editing in the future.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: Done.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
olliw42: can you explain, what is error prone here?

rotorman: If you add a line to else clause, one easily oversees that the brackets were missing in the first place. This is the reason, I prefer to always use brackets for else case.

olliw42: well, I don't share that view and don't see any increased error proness ... the indenting shows where one is. Editing is generally dangerous.
anyway, this shall not block it. It will naturally go eventually. :)

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
olliw42: it compiles and works partially (see below). So this version change is not needed and can be reverted.

olliw42:
I'm trying this code now and have a number of issues.

1) It took me few minutes to realize that teh BluetoothSerial library is apparanetly not available for many boards, but just few. I actually installed it and they it complained it finds two installs LOL. Anyway, when merged I will do some changes to give the user a better chance here.

2) Much more problematic, the serial traffic seems to work only in one direction, namely the direction tx -> MisssionPlanner, but not MissionPlanner -> tx.
Btw, I have tested it also with TCP and UDP, and both worked fine, so it's not setup mistake like bad wire or so.

That is, I do see all the telemetry messages (HUD etc. pp) in MissionPlanner, but e.g. param upload just doesn't work. This suggests that something doesn't work with SerialBT.available()/SerialBT.read() functions. To see if it's the code changes I put upon you in this section I reverted to what you had initially with the same sympthom. 

Any idea what I may do wrong?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -279,7 +275,49 @@ void loop()
...
+        udp.beginPacket(ip_udp, port_udp);
+        udp.write(buf, len);
+        udp.endPacket();
+    }
+#elif (WIRELESS_PROTOCOL == 2) // Bluetooth
+    int len = SerialBT.available();
+    if (len > 0) {
+        if (len > sizeof(buf)) len = sizeof(buf);
+        for (int i = 0; i < len; i++) buf[i] = SerialBT.read();
+        Serial.write(buf, len);
```
olliw42, line 305: BUG !!!! This must be SERIAL

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: Great, thanks. Reverted.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
rotorman: Thx, fixed.

olliw42:
many thx sir, this is very cool!

brad112358:
I had been thinking doing this same thing because of the same issue as people were mentioning on discord:  I can't seem to get my phone to connect to the internet via Cellular while the WiFi is connected to the esp32, even though the phone knows the WiFi network has no internet connectivity.  I think this worked back in my crossfire days.  I wonder if the esp32 AP code is advertising a default route.  We don't want that as it might cause the phone to route everything to the WiFi even when the cellular network is available.

Anyway, I decided to test this PR and it works great with the R9M and the M5Stack M5Stamp Pico.  Very nice!  I don't see any lost messages is most connection sessions.  Sometimes just a few messages are lost when first connecting.

Unfortunately, the ESP32C3 chip doesn't support BT SPP, only BT LE.

PR #117, opened 2023-08-24
stack check

olliw42:
@rotorman 
attempt at checking stack usage by painting

### DIFF
--- a/mLRS/Common/stack_check.h
+++ b/mLRS/Common/stack_check.h
@@ -41,7 +41,8 @@ void stack_check_init(void)
 {
     // paint the stack memory area
     uint32_t ptr = (uint32_t)&_estack - (uint32_t)&_Min_Stack_Size;
-    uint32_t end = (uint32_t)&_estack;
+    //uint32_t end = (uint32_t)&_estack;
+    uint32_t end = __get_MSP() - 1; // let's use MSP to avoid concerns with used stack space at time of function call
     while (ptr < end) {
         *(uint8_t*)ptr = 0xAA;
         ptr++;
--- a/mLRS/Common/stack_check.h
+++ b/mLRS/Common/stack_check.h
@@ -1,0 +1,52 @@
+//*******************************************************
+// Copyright (c) MLRS project
+// GPL3
+// https://www.gnu.org/licenses/gpl-3.0.de.html
+// OlliW @ www.olliw.eu
+//*******************************************************
+// Stack check
+//*******************************************************
+#ifndef STACK_CHECK_H
+#define STACK_CHECK_H
+#pragma once
+
+
+#include <inttypes.h>
+
+
+//-------------------------------------------------------
+//
+//-------------------------------------------------------
+// see
+// - https://mcuoneclipse.com/2016/11/01/getting-the-memory-range-of-sections-with-gnu-linker-files/
+// - https://stackoverflow.com/questions/55622174/is-accessing-the-value-of-a-linker-script-variable-undefined-behavior-in-c
+extern uint32_t _estack;
+extern uint32_t _Min_Stack_Size;
+
+
+uint32_t stack_check_used(void)
+{
+    uint32_t ptr = (uint32_t)&_estack - (uint32_t)&_Min_Stack_Size;
+    uint32_t end = (uint32_t)&_estack;
+    while (ptr < end) {
+        if (*(uint8_t*)ptr != 0xAA) break;
+        ptr++;
+    }
+
+    return (end - ptr);
+}
+
+
+void stack_check_init(void)
+{
+    // paint the stack memory area
+    uint32_t ptr = (uint32_t)&_estack - (uint32_t)&_Min_Stack_Size;
+    uint32_t end = (uint32_t)&_estack;
+    while (ptr < end) {
+        *(uint8_t*)ptr = 0xAA;
+        ptr++;
+    }
+}
+
+
+#endif // STACK_CHECK_H
--- a/mLRS/CommonTx/mlrs-tx.cpp
+++ b/mLRS/CommonTx/mlrs-tx.cpp
@@ -76,6 +76,7 @@ v0.0.00:
 #include "../Common/common.h"
 #include "../Common/micros.h"
 #include "../Common/channel_order.h"
+#include "../Common/stack_check.h"
 //#include "../Common/test.h" // un-comment if you want to compile for board test
 
 #include "in.h"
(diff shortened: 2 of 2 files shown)


mLRS/Common/stack_check.h
```diff
@@ -0,0 +1,52 @@
...
+
+
+void stack_check_init(void)
+{
+    // paint the stack memory area
+    uint32_t ptr = (uint32_t)&_estack - (uint32_t)&_Min_Stack_Size;
+    uint32_t end = (uint32_t)&_estack;
+    while (ptr < end) {
+        *(uint8_t*)ptr = 0xAA;
+        ptr++;
```
rotorman, line 47: Isn't the pointer incrementing here with 4 bytes/32-bits. Also line 33 above?
I would suggest to use:
```c++
*ptr = 0xAA55AA55;
ptr++;
```

olliw42: was worried first about this too, but IMHO not, since ptr is not a pointer but just a uint32_t variable

this refutes also your suggestion :)

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -648,6 +649,7 @@ int main_main(void)
 #ifdef BOARD_TEST_H
   main_test();
 #endif
+  stack_check_init();
```
rotorman, line 652: It is probably too late in `main()` to fill the full stack, as you are likely overwriting some variables here. Alternative would be to do it in `startup_*.s` in assembly or fill only the portion that has not yet been used. See, e.g.: https://stackoverflow.com/questions/71816900/c-code-to-paint-an-embedded-stack-with-a-pattern-say-0xabababab-just-after-mai

rotorman: Here my suggestion: https://github.com/rotorman/mLRS/commit/ddbc64acaf993eb511330a74d0ff7bd9c82180c4
@olliw42 Interested to hear what you think about going with assembly like this.

mLRS/Common/stack_check.h
rotorman: Would need to verify with in-circuit-debugger what happens with ptr++.
Alternatively:
```c++
if (*(uint32_t*)ptr != 0xAAAAAAAA) break;
```
?

olliw42: it is a ptr in that you access the loaction with it, i.e. use *ptr to set it. But pointer are just uint32_t by itself, since they are just 32bit adressess. So, I think it's all fine. I sure admit though that I also puzzeled few moments what exactkly needs to be done here.

mLRS/CommonTx/mlrs-tx.cpp
olliw42: haha ... I knew this argument would be raised :):):) 
(from which you can determine that I had thought about what I'm doing here LOL)

I want to avoid by all possible means to mangle with the .ld or .s files. This has, not only but also, to do with that we are using CubeIDE and not a full build system. E.g. just a mistaken click on the .ioc file will replace these files, and you are screwed. And not everyone may realize this or even may know what then to do. So, I wouldn't want to merge such changes to main.

Also, I don't really see your argument. It is technically correct but somewhat academic; there are no variables on stack which are used upon entry of main (AFAICS). If worried, I rather would simply only paint the upper part of the stack, so that anything happing in the lower part is still intact. I mean, what we care about is stack overflow, not if it is used by 20 or 100 or 300 bytes.

In combination of these arguments I really favor the taken approach. I guess I agree though that one should simply add e.g. 128 to ptr. Just to be on the safe side.

rotorman: Following idea to test this: https://github.com/rotorman/mLRS/commit/b1a5e2add783bc6971e9bf61673518cedb4ce9fe
Assembler code is painting with 0x55. Will put a breakpoint, when this is done and then observe 0x20003C00 to 0x20003FFF. Then will put a breakpoint before your stack painting with 0xAA starts and will see if all bytes between 0x20003C00 to 0x20003FFF are still 0x55.

olliw42: as said I will change the above code such as to not touch the first 128 bytes.

your test is (potentially) flawed. .s calls two  functiuons, and these functions may put variables on the stack, and if they do the respective stack area will be unpainted, even though nothing bad has happened. 

What you probably want to do is to check the SP, and see if it has advanced. 

this actually could suggest a slightly better approach than to subract 128, namely to just paint from SP downwards... I may like that ...

rotorman:
Here some results.

Breakpoint after assembler painting with 0x55:
grafik

Breakpoint before C++ painting:
grafik

Hardfault when executing mlrs-tx.cpp:L655 `stack_paint()`:
grafik
due to overwriting necessary information.

rotorman:
I removed the C++ painting and changed the assembly painting to `0xAA`. Here how TX side FRM303 stack looks like after some minutes, when connected to RX side FRM303 and Mission Planner, via serial and mlrs-wireless-bridge loaded on M5Stack ATOM Lite:

grafik

rotorman:
I'm gonna put it this way - the correct way is the assembler way before any C/C++ preparation code runs...

rotorman:
The state after 10 min - free space has not decreased:
grafik

I will let it run to see if I get telem lost at some point.

olliw42:
> We can see that stack from top 0x20003FFF down to 0x20003E50 is in use (432 bytes, thus limiting with 128 would have been too little).

I disagree. I think you have not yet picked up yet in which way this argument is flawed. You should look at MSP at time of painting.

I put stack_check_init() AFTER init() (that is even much later than it is now) and added debug printout of some variables, especially of MSP. And MSP was only at 80 bytes below estack. 

Anyway, I have changed it to not paint above MSP, which really should remove any concerns.

I too - on my tx and rx diy boards I currently use (not FRM303) - see stack occupancies of ca 800 bytes. If that upholds for long time stack might be ok. I wonder though if we could explain this 800-ish number. I mean, the static stack analyzer gives just quite low values.

:)

Anyway, this is something I wanted to do since long, also for other projects of mine, and I'm glad I did it now. Can't hurt.

olliw42:
> Here how TX side FRM303 stack looks like after some minutes, when connected to RX side FRM303 and Mission Planner, via serial and mlrs-wireless-bridge loaded on M5Stack ATOM Lite:
grafik
We can see only 192 unused bytes. Will keep observing to see if this decreases over time.

that's super weired ... is it normal that there are gaps in the stack ??? I would not think so ... how does that come ?!?

rotorman:
After ca. 40 minutes, I now have a telemetry lost and, steady green LED on TX-side FRM303, blinking red LED on RX side FRM303. Stack looks like this:
grafik

olliw42:
largely settled, check check optionadded to main

PR #118, opened 2023-08-24
Added mLRS-Wireless-Bridge for ESP8266

rotorman:
As ESP8266 cannot do Bluetooth, this edition supports WiFi only.

Tested with ESP-01S, but needs more testing.

### DIFF
--- a/esp/mlrs-wireless-bridge-esp8266/mlrs-wireless-bridge-esp8266.ino
+++ b/esp/mlrs-wireless-bridge-esp8266/mlrs-wireless-bridge-esp8266.ino
@@ -1,0 +1,298 @@
+//*******************************************************
+// mLRS Wireless Bridge for ESP8266
+// License: GPL v3
+// https://www.gnu.org/licenses/gpl-3.0.de.html
+//*******************************************************
+// Basic but effective & reliable transparent WiFi<->serial bridge
+//*******************************************************
+// 28. Aug. 2023
+//*********************************************************/
+// Adapted from mlrs-wireless-bridge for ESP32
+// ArduinoIDE 2.1.1, esp8266 by ESP8266 Community 3.1.2
+// ------------------------------
+// Tested with Ai-Thinker ESP-01S
+// ------------------------------
+// board: Generic ESP8266 Module
+// http://www.ai-thinker.com/pro_view-88.html
+// IO3/IO1: U0RXD/U0TXD is Serial
+//
+// Please be aware that this code here is unmaintained
+// and would also benefit from further testing feedback.
+
+#include <ESP8266WiFi.h>
+
+#ifndef ESP8266 // ARDUINO_BOARD != Generic ESP8266 Module
+  #error Select board Generic ESP8266 Module!
+#endif
+
+#define LED_IO  2
+
+//-------------------------------------------------------
+// User configuration
+//-------------------------------------------------------
+
+// Wireless protocol
+// 0 = WiFi TCP, 1 = WiFi UDP, 2 = Wifi UDPCl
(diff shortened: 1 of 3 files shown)


olliw42:
very nice. I like that, and I like you put the esp8266 into an extra file.

however, please don't rename the esp32 folder&files. Few days ago it all was slim and comprehensible, but now it's getting somewhat out of hands :):)
also, remind, each such change induces a chain of actions in the docs (they still don't even reflect the last change fully)
let's pl stick with mlrs-wireless-bridge for the EPS32 branch :)

rotorman:
OK, reverted the folder and file name change. It surely would be possible to have only a single *.ino file, but there is IMO no way around having some #ifdefs to differentiate between ESP32 and ESP8266.

olliw42:
THX.

could you please rename the esp8266 folder/file to only use - insteasd of _. I.e. mlrs-wireless-bridege-esp8266.

rotorman:
Sure, done.

esp/mlrs-wireless-bridge-esp8266/mlrs-wireless-bridge-esp8266.ino
```diff
@@ -0,0 +1,287 @@
+//*******************************************************
+// mLRS Wireless Bridge for ESP8266
+// License: GPL v3
+// https://www.gnu.org/licenses/gpl-3.0.de.html
+//*******************************************************
+// Basic but effective & reliable transparent WiFi<->serial bridge
+//*******************************************************
+// 24. Aug. 2023
+//*********************************************************/
+// Adapted from mlrs-wireless-bridge for ESP32 by OlliW42
```
olliw42, line 10: pl drop the "by OlliW42". 
first, it can be missleading as it could be read as "adapted ... by..." and also the credential is already in teh other file.

olliw42:
> Before merging, would be nice to have at least one user provide feedback about success of using this code.

would be nice ... but may potentially delay merge to indefinite

olliw42:
lgtm from my side. MANY THX!

concenring "before merging, would be nice to have at least one user provide feedback", pl drop a note when you want it to get merged :)

rotorman:
I had a look how to integrate ESP-01S support into existing mlrs-wireless-bridge and it does not look bad: https://github.com/rotorman/mLRS/commit/7db29ef33b9027735a2d75fa48d04152ed78796f
Should we close this here and I re-open a new PR with the integrated code?

Note that the rename of SERIAL to SERIALx stems from the issue that ESP8266 compiler does not seem to differentiate between upper and lower case, thus `#define SERIAL serial` gave me a build error.

olliw42:
I was briefly looking at this too, and it indeed doesn't look too bad ... but:
- it would mangle in a less maintained hardware ionto this file, and it would be less possible to make the distinction clear
- while it may blend in well now, it is much less capable hardware and it's not so clear how it would blend in in the future. E.g., if we would add ESP-Now, or have some config storage, and so on. I just note that in other projects whioch support both ESP32 and ESP8266 there tend to be lots of define dirt.

I am hesitant.

I guess we can keep it as it is. It's good to know however how it woudl look like in one file.

> that ESP8266 compiler does not seem to differentiate between upper and lower case

I frankly find that hard to believe. I would think that there is probably another issue at work.

rotorman:
> I frankly find that hard to believe. I would think that there is probably another issue at work.

I you have an idea what else it could be, would be interested to learn it.

olliw42:
it's just that I never have come across a C compiler which would be case insensitive
my first guess would be that some library defines it already itself

olliw42:
many THX !!!

PR #119, opened 2023-08-26
Increase stack size for TX-side FRM303

rotorman:
Testing with in-circuit-debugger shows that 1024-bytes (0x400) is likely not enough for stack for TX-side FRM303.

Below I include a Segger Ozone screenshot, showing an overflow of 0x400 region (stack top 0x20003FFF down to stack bottom 0x20003C00).
For testing, a stack size 0x600 was used (stack top 0x20003FFF down to 0x20003A00) that was painted in assembly start script with 0xAA: https://github.com/rotorman/mLRS/blob/e1ba91e23f023697bf900fc342d835da41ea3e25/mLRS/tx-FRM303-f072cb/Core/Startup/startup_stm32f072cbtx.s#L88-L101

We can see that the stack memory has been used down to 0x20003BE0, which is lower than 0x400 stack bottom address of 0x20003C00:

grafik

This PR increases stack size to 1.5Kbytes (150%).

### DIFF
--- a/mLRS/tx-FRM303-f072cb/STM32F072CBTX_FLASH.ld
+++ b/mLRS/tx-FRM303-f072cb/STM32F072CBTX_FLASH.ld
@@ -39,7 +39,7 @@ ENTRY(Reset_Handler)
 _estack = ORIGIN(RAM) + LENGTH(RAM); /* end of "RAM" Ram type memory */
 
 _Min_Heap_Size = 0x0 ; /* required amount of heap */
-_Min_Stack_Size = 0x400 ; /* required amount of stack */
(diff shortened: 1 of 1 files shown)


olliw42:
the problem will be solved soon at it's root cause :)

also, it would be better have to be done from within CubeMX, sicne otherwise it would be overwritten too easily.

MANY THX nevertheless. :)

olliw42:
this https://github.com/olliw42/mLRS/commit/baebddff21bebdde54e3df0a2263aef2dfb8666f reduces stack use quite a bit, and also improves characteristic.
Should settle the issue.
This closing. :)
THX again!

PR #122, opened 2023-09-14
high refresh crsf tweak

jlpoltrack:
Have been doing some testing with 100, 111 and 125 Hz modes for FLRC and this change was needed for all of them.

### DIFF
--- a/mLRS/CommonTx/crsf_interface_tx.h
+++ b/mLRS/CommonTx/crsf_interface_tx.h
@@ -333,7 +333,7 @@ bool tTxCrsf::TelemetryUpdate(uint8_t* task, uint16_t frame_rate_ms)
     	  telemetry_start_next_tick = false;
 
         // slow it down if frame time is too short
-        if (frame_rate_ms <= 10) {
+        if (frame_rate_ms <= 19) {
         	  static uint8_t cnt = 0;
         	  if (!cnt) telemetry_state = 0;
         	  cnt++;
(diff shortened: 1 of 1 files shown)


olliw42:
@jlpoltrack wouldn't 10ms/100Hz be ok with just slowing down by a factor 2 (i.e. with cnt >1)?
I wonder if it would be worth to add two cases, one for rate_ms < 20 ms which slows by factor of 2, and one with < 10ms which slows by a factor of 3, like now.
Any insights/thoughts?

jlpoltrack:
I don't notice any difference on the radio and since it is wanted to only have 1 FLRC mode, I think just having one condition makes more sense.

olliw42:
thx
couldn't it make sense to set the limit more conversatively, I mean, the next is at 20 ms, and we probably want to start slowing down sonner than at 10 ms (at let's say 15 ms there already wouldn't be space for telemrtry frames). So maybe we should change to rate_ms <= 19 or so ...

jlpoltrack:
> thx couldn't it make sense to set the limit more conversatively, I mean, the next is at 20 ms, and we probably want to start slowing down sonner than at 10 ms (at let's say 15 ms there already wouldn't be space for telemrtry frames). So maybe we should change to rate_ms <= 19 or so ...

Yea, this makes sense to me.  Will update

olliw42:
MANY THX!!

PR #123, opened 2023-09-15
Relay Mode

olliw42:
for those who want to play and test :)

this adds the new option "crsf tx" to the RxOutMode parameter. This option should be selected on the receiver on the relay station (= secondary receiver). This is, in principle, all as regards of the required configuration.

However, in addition some configurations need to be done to have it all working together, e.g. also the baudrate should be selected to match the tx module on the relay station (= main tx module), usually 115200. Also RxChOrder should be selected to match that in the tx module in the radio (= secondary tx module). "crsf" should be selected in both tx modules, and mavlink must be enabled in all devices.

The out pin of the secondary receiver should be connected to the jrpin5/in pin on the main tx module, as well as the serials. With regards to the serials, it is mandatory to connect the tx line on the secondary receiver, the rx line can be left open, but of cousre also be connected if full mavlink shall be relayed.

The RC data is properly transmitted, also the mavlink data, and yappu works on the radio as well as the crsf telemetry shows the stats corresponding to the main link, not the relay link.

The main tx module and main receiver cannot be configured via the lua script. That is for configuring the main link, and easy approach would be to just plug the main tx module into a radio and set things up for them.


rotorman:
Here two images that might help to understand how the relay mode could be used (just showing two examples, more constellations are possible):

grafik

grafik

jiri-jirus:
I am here to test relay mode. would it be possible to prepare up-to-date firmwares for matek 2.4&868 RX/TX please? thanks

PR #124, opened 2023-09-16
feat: added support for VCP on G4

Gurkengewuerz:
as discussed on the dirscord the first PR for supporting VCP on G4 MCUs. you @olliw42 already mentioned there isn't (a currently known) better way to define the includes for the G4 series in `mLRS/modules/stm32-usb-device/usbd_conf.h`.

Edits/Reviews/Suggestions to this PRs are as always welcome.

PR #125, opened 2023-09-16
feat: added g431cb support for firmware3 build script

Gurkengewuerz:
like #124 also discussed on #hardware channel in discord. this is needed to support the g431**cb** series different with more I/O and 48 pins from the already supported g431**kb** series.

here also applies reviews/edits/suggestions are welcome.


olliw42:
MANY THX!
merging

PR #128, opened 2023-09-24
flrc downlink fix

jlpoltrack:
Believe downlink speed should be 9111 Bytes / sec:

(1000/9) * 82 = 9111.111

Uplink is calculated correctly:

(1000/9) * 64 = 7111.111


olliw42:
:+1:

PR #132, opened 2023-10-14
e5, SPI2 define

jlpoltrack:
Copied from Rx Hal, does compile with diversity enabled but untested on hardware (don't have)


olliw42:
MANY THX!

so, I suggest we wait few days in case jdavies does a report.

olliw42:
seems to have been confirmed. Merging. MANY THX!

PR #135, opened 2023-10-17
SysTask decrement for Rx

lithuak:
Hey! It looks like this change https://github.com/olliw42/mLRS/commit/2edf65745f0bc0449e148fd05fb953ff0201ac6a has affected transmitter, but not receiver.

Changing rx logic in the same way looks right.. Otherwise e.g. [the next lines](https://github.com/olliw42/mLRS/blob/a137f3d0ad05cbd48604c7bf2706d7906ae86775/mLRS/CommonRx/mlrs-rx.cpp#L526) would be decremented by one tick regardless the actual number of missed ticks?..

### DIFF
--- a/mLRS/CommonRx/mlrs-rx.cpp
+++ b/mLRS/CommonRx/mlrs-rx.cpp
@@ -520,7 +520,7 @@ RESTARTCONTROLLER:
     //-- SysTask handling
 
     if (doSysTask) {
-        doSysTask = 0;
+        doSysTask--;
 
         if (connect_tmo_cnt) {
             connect_tmo_cnt--;


olliw42:
thx!
this on purpose, on the rx side the incrementing doesn't have to be done, just the flag needs to be set. All good here.

closing

lithuak:
sorry for bothering... my point was: the incrementing happens now (after the mentioned change) on both rx and tx side, because it was done in common.h which both side use. But the decrement happens in different ways (tx decrements, rx resets the flag). May be not a big deal right now - it only (potentially) affects the difference in how sides calculate the connection timeouts, how leds blink, etc. Could become a source of desynchronization later on, if extending rx code.

olliw42:
it's fine
the point is just that you start from a wrong point and thus arrive at wrong conclusions. The issue is not the rx side, but the tx side. As said, this is all on purpose. It won't create issues later in time. Fundamentally, a systick should never be missed and if it is then there this causes much more serious issues than connection timeout or led blinking. You may need to get a bit more aquianted with the details of the timing in the code.

lithuak:
thanks! just starting to learning the code, can get confused :)

olliw42:
it's fantastic that you are venturing into learning the code, very much appreciated, but some aspects may not be immediately obvious. I'm the last who would say the code hasn't any bugs, but not everything which may look suspicious at first sight is a bug but is intentional :)

lithuak:
@olliw42 what's the best channel to quick ask/check about such non-obvious things w/o opening issue or PR?.. thank you

olliw42:
good question, I guess discord ... maybe we should add a dev channel there, minique may do it

olliw42:
@lithuak we do have now a "development-talk" channel in the discord server :)

PR #136, opened 2023-11-17
Sx126x fsk

jlpoltrack:
Tested and worked on 2x E5 Mini.

- New FSK Mode
- SX126x driver uses GFSK references to align with datasheet


olliw42:
M A N Y T H X!

PR #141, opened 2023-11-30
Fix flow control broken by USE_FEATURE_MAVLINKX

brad112358:
USE_FEATURE_MAVLINKX had a couple of problems with respect to txbuf calculations which badly broke flow control.

1) The calculation of buffered received serial data in serial_in_available() needs to include the data which is in the process of being parsed as well as the data which has already been parsed but not yet sent over the radio link.

2) We should not parse and transfer to fifo_link_out more than one message in the while loop since we only checked that there is space for one message.  We also shouldn't spend too much time here without doing flow control.

### DIFF
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -108,6 +108,7 @@ class MavlinkBase
     void generate_cmd_ack(void);
 
     uint8_t _buf[MAVLINK_BUF_SIZE]; // temporary working buffer, to not burden stack
+    uint16_t parsing_in_cnt;
 };
 
 
@@ -132,6 +133,7 @@ void MavlinkBase::Init(void)
     radio_status_txbuf = 0;
     txbuf_state = TXBUF_STATE_NORMAL;
 
+    parsing_in_cnt = 0;
     bytes_serial_in = 0;
     bytes_serial_in_cnt = 0;
     bytes_serial_in_rate_filt.Reset();
@@ -201,6 +203,7 @@ void MavlinkBase::Do(void)
     if (fifo_link_out.HasSpace(290)) { // we have space for a full MAVLink message, so can safely parse
         while (serial.available()) {
             char c = serial.getc();
+	    parsing_in_cnt++;
             if (fmav_parse_and_check_to_frame_buf(&result_serial_in, buf_serial_in, &status_serial_in, c)) {
 
                 // TODO: this could be be done more efficiently by not going via msg_link_out
@@ -216,10 +219,12 @@ void MavlinkBase::Do(void)
                 }
 
                 fifo_link_out.PutBuf(_buf, len);
+		parsing_in_cnt = 0;
 
                 if (msg_link_out.msgid == FASTMAVLINK_MSG_ID_COMMAND_LONG) {
                     handle_cmd_long();
                 }
+		break; // We may not have space for a second message and need to check txbuf now
 
             }
         }
@@ -415,7 +420,7 @@ void MavlinkBase::send_msg_serial_out(void)
 uint16_t MavlinkBase::serial_in_available(void)
 {
 #ifdef USE_FEATURE_MAVLINKX
-    return fifo_link_out.Available();
+    return fifo_link_out.Available() + serial.bytes_available() + parsing_in_cnt;
 #else
     return serial.bytes_available();
 #endif


olliw42:
interesting
I thought I did got it right, but maybe not 
could share what sympthoms you were seeing which made you conclude that it badly broke flow control?

mLRS/CommonRx/mavlink_interface_rx.h
```diff
@@ -216,10 +219,12 @@ void MavlinkBase::Do(void)
                 }
 
                 fifo_link_out.PutBuf(_buf, len);
+		parsing_in_cnt = 0;
 
                 if (msg_link_out.msgid == FASTMAVLINK_MSG_ID_COMMAND_LONG) {
                     handle_cmd_long();
                 }
+		break; // We may not have space for a second message and need to check txbuf now
```
olliw42, line 227: the 2nd part of the comment seems not correct, since it's not garanteed that by jumping out txbuf is checked, may or may not happen
I guess what we want to do is to check space again and if too small jump out
`if (!fifo_link_out.HasSpace(290)) break; // not enough space for another message `
It's sure safer to do this, but I'm not convinced this makes any difference. I guess my "argument" is that it should never happen that in that loop a 2nd message can occur, Mavlink:Do() is called often but it needs quite some time for the bytes of a full next message to flow in. What evidence/proof do you have that not having the break gives troubles?

mLRS/CommonRx/mavlink_interface_rx.h
```diff
@@ -132,6 +133,7 @@ void MavlinkBase::Init(void)
     radio_status_txbuf = 0;
     txbuf_state = TXBUF_STATE_NORMAL;
 
+    parsing_in_cnt = 0;
```
olliw42, line 136: maybe in analogy bytes_parser_in_cnt ?

brad112358:
The symptoms without these fixes were lost messages at all baud rates between 38400 and 230400 and with all combinations of mavlink, mavlinkx, and the two txbuf methods.  The message loss was 3 to 40% depending on settings as measured by QGC in the application->mavlink page. Slow and failed parameter download was also observed.  Message loss was highest during bulk parameter download, but non-zero in steady state as well.  As you know, I only have R9, so 19Hz. My QGC configuration requests flow rates slightly above what the radio link can handle (some flows run at about half the requested rate because of flow control).  This should be enough info for you to reproduce the problems if you want proof.  I fixed the byte counting problem first, but that wasn't enough to eliminate the symptoms, so I added the break which completely solved the problem.

Unless you check for lost messages based on sequence numbers, parameter download might be the only place you would notice the problem.  Perhaps you didn't notice the problems because you use higher radio message rates which allows ftp for parameter download at some baud rates.

Note that I normally use 230400 baud to minimize the latency caused by needing to parse mavlink messages at both ends of the radio link and I normally use method_b for flow control (even for ardupilot).  But other configurations also lost messages.

I'm not sure why it's not obvious to you as it is to me that both of these changes are required.  I will try to outline my thinking.

Without the counting fix, we will never trigger on the tests for serial_in_available() > 1024 because the buffer is only 512 in size and that's as big as it can get, so we *will* overflow our buffers during parameter download which won't stop until we send txbuf < 50.  We also won't slow transmission as quickly for other flows.

Even with the counting fix but without the break or some other code change inside the loop we will loose messages.  Anytime the autopilot is sending faster than the radio link, the two buffers, serial, and fifo_link_out will buffer more and more data over time. Over the course of several calls to Do(), first, fifo_link_out will first fill to less than 290 bytes remaining.  Then, eventually in subsequent calls to Do(), serial will also begin to fill until serial_in_available() gets higher than 1024. Only then, we will stop parameter flow or rapidly slow down normal flows.  At some point, we will have more than 290 bytes (perhaps up to about 1024-512+290) of data waiting in serial and the while loop will attempt to parse all of it in one call to Do() and will overflow fifo_link_out, loosing at least one message.

The break is better than adding an extra test inside the loop because it limits the time we spend in the loop a bit and reduces the time we parse data without checking to see if we need to set TXBUF_STATE_BURST or do other txbuf related checks.  Parsing at most one message helps reduce our flow control response time to bursts of messages from the FC (remember, ardupilot is very bursty) and reduces the chances of buffer overflow and reduces buffering related message latency.  The significance of these effects depends on the time it takes to parse and copy a couple of hundred bytes.  Unless you know for sure that this time is much shorter than the time it takes to receive a message at 230400 baud, the break is better.  Even if you know this timing precisely, the break is simpler.

mLRS/CommonRx/mavlink_interface_rx.h
brad112358: I've addressed both of these points in my comment below, but phrasing it a different way may help.  The second message inside the loop can easily happen because when messages arrive on the serial port faster than the radio link can remove them, fifo_link_out is drained more slowly than messages may come into our input buffer which is larger than fifo_link_out so, serial may have more than one message (and even more than 512 bytes) available.  Would it be more clear if I said "check the size of serial" rather than "check txbuf" here?

mLRS/CommonRx/mavlink_interface_rx.h
brad112358: Sure, if you prefer, though I think what I have is at least as clear.

olliw42:
many thx for your explanation. Much appreciated.
I'll merge, and do some minor cosmetics on it

I do regularly also look at the behavior in 19 Hz mode, but I hadn't noticed anything like "uuuupppssss". What you describe sounds like rather catastrophic, so I'm not sure why I didn't notice but i won't rule out that I may not have investigated as carefully as you may have done.

olliw42:
@brad112358 
maybe you recall, I'm quite confused momentarily: what was `bytes_serial_in` for? Why isn't it relevant to take into account either? 
It's used to calculate the rate on which the txbuf value is determined. So I would think it shoudl relate to the bytes as they are flowing in.
To me it seems however that it actually should better be called `bytes_link_out`, since it's counted in MavlinkBase::getc(), which is called to get the bytes for the link. Which is not the count of the inflowing bytes.
It somehow seems to me we may have a more serious bug here.

brad112358:
Isn't it amazing how quickly code we write becomes less than obvious?  As you may remember (or, perhaps not, but you will see in the code), this variable is used in calculating the percentage of the radio link used in the last second so we can base txbuf on data rate rather than buffer used.  It's name seemed fine to me when you wrote the code, but, I had the same thought when reviewing the code while investigating the message loss last week.  Unfortunately, MAVLinkX translation complicates things.  I still don't like this approach much which is why I hadn't tested much with MAVLINKX code enabled and missed the flow control problem earlier.

In the presence of compression and to a lesser extent, even with just MAVLinkX translation, the amount of data we want to send over the radio link in a second no longer exactly matches the number of bytes we read from the serial link.   We can probably make a more accurate estimate of average radio link usage by counting bytes after the translation to MAVLinkX.  So, not only should the name change, the calculation should also change to count bytes placed into fifo_link_out.  

Unfortunately, this will also introduce latency in responding to surges in traffic because fifo_link_out is relatively small and bytes received can't always be translated and counted as soon as they come in over the serial link.  It's not clear to me how much this latency might slow down our response (via txbuf) to changes in flow rates.  But, if we are only counting serial input bytes and not compressed size, I don't see how compression was actually providing any useful extra link capacity since flow control was based on uncompressed size.  If this change causes problems, we might need to increase the size of fifo_link_out.  I also have some other ideas on how to increase flow control responsiveness in the case of bursts of traffic which I'll save for another discussion.

PR #145, opened 2024-03-10
cli freq typo

jlpoltrack:
Title


olliw42:
THX!

olliw42:
@jlpoltrack 
OH NO!! I was too quick ... it is supposed to mean f_reg, for register value!!

could you maybe do another PR with changing it to f_reg?

PR #147, opened 2024-03-28
single led update

jlpoltrack:
Title


mLRS/Common/fail.h
```diff
@@ -102,6 +101,7 @@ void fail(tSerialBase* dbg, uint8_t led_pattern, const char* msg)
 
     led_red_on();
     led_green_on();
+#endif
```
olliw42, line 104: should go after the last while(1) (will never be reached)

jlpoltrack: Addressed

PR #148, opened 2024-04-10
Incorporate code for mLRS compatiability ESP8285 ELRS receiver hardware

tmcadam:
This code incorporates changes to mLRS code to allow the system to operate on ELRS ESP8285 RX devices.


olliw42:
maybe you "just" need to do an interactive rebase on your main ... might be nasty to do if you've never done since you'll get the conflicts which you need to resolve interactively

tmcadam:
Yeah, I'm just looking at these conflicts. I think some changes in olliw42/main have already been incorporated in tmcadam/main. I can't rebase cleanly. I can merge in olliw42/main and commit the merge changes. Which should allow this to go in. I suspect the buttons will be enabled when we resolve the merge conflicts. 

What do you propose as the version number to go with this?

olliw42:
>  I suspect the buttons will be enabled when we resolve the merge conflicts.
usually at least the [Resolve conflicts] button is enabled ... I'm surprised it is not

> What do you propose as the version number to go with this?
with what is in main, i.e. 1.2.01

tmcadam:
> GREAT somewhat magic to me, but you seem to have pulled the right strings :):)
> 
> I'll merge this evening

Let's review a little more before merging.  Maybe I stuffed something up in the merge conflict fixes and few outstanding questions.

olliw42:
just as a note, this PR doesn't have to be perfect in all these details. They always can be worked on after this inauguration PR :)

jlpoltrack (approved):
I tested today with 2400 PA-D with 2400 PA Target (yes a bit rude) but had no problems connecting and getting MAVLink data.

PR #151, opened 2024-04-13
esp, rx,tx, hal cleanify, sync with stm, move SYSTICK define, nfc

olliw42:
moves SYSTICK defines from device hal files to hal.h and timer.h
this follows the cleanify for the stm32 code
it's nfc
tested to compile for all esp targets, and work for speedybee


jlpoltrack (approved):
Looks good, tested on Generic 900 w/ FC.

olliw42:
THX !!

PR #152, opened 2024-04-13
Esp interrupt cleanup

jlpoltrack:
Rationale: ESP32 doesn't like noInterrupts / Interrupts functions.  

This removes them from the ESP8266 code, tested with Generic 900.

- Removing from rxclock had no impact.
- In order for EEPROM write to succeed, DIO interrupt needs to be disabled.  Created a new function in the SX127x and SX128x drivers called 'deInit' which does this.

If this makes sense, will also need to update all of the Hals to detach the DIO interrupt.

### DIFF
--- a/mLRS/Common/esp-lib/esp-eeprom.h
+++ b/mLRS/Common/esp-lib/esp-eeprom.h
@@ -182,15 +182,11 @@ EE_STATUS_ENUM ee_writedata(void* data, uint16_t datalen)
 EE_STATUS_ENUM status;
 
     // Write data to page0
-    noInterrupts();
     status = _ee_write_to(EE_PAGE0, data, datalen);
-    interrupts();
     if (status != EE_STATUS_OK) return status;
 
     // Write data to page1
-    noInterrupts();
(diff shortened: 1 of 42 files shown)


olliw42:
I think we had this discussion, at least it looks to me we did ... and it totally does not make sense to me. Not protecting a critical block is just not ok. Frankly, to me it looks like we just don't know how to do it correctly.

olliw42:
also I think this shouldn't go into main but a esp dev branch. IMHO we shouldn't fumble with main until we see how it all unfolds and pieces together for esp32

jlpoltrack:
> also I think this shouldn't go into main but a esp dev branch. IMHO we shouldn't fumble with main until we see how it all unfolds and pieces together for esp32

Good point, I will close this PR.

I understand your point on the critical RXClock piece.  Are you strongly against the DeInit concept?

olliw42:
> Are you strongly against the DeInit concept?

not neccessarily, I just don't see it should be needed

I guess I will do a search in the elrs code base for attachInterrupt and deattach, I feel we are missing something

PR #154, opened 2024-04-14
esp, add mavlink params, add esp82xx tx support, add tx-generic-2400-pa target

olliw42:
as the title says, this PR does 3 thiongs, separated in 3 commits:
- adds mavlink params for tx
- adds basic tx support for esp82xx
- adds tx-generic-2400-pa target

tested to compile for all targets
tested to work on speedybee receiver as tx (with no com, no debug, no in, no jrpin5, only with serial)(debug I did also test) 

note: I made a change in esp-rxclock.h, which is inspired by the finding that initializing MS_C with 0 did not start the timer for tx. It's a bug, and clear from the ISR, which first increases CNT, so that the following if check never gets true. In the case of the old rxclock.h this didn't bite since Reset() was called in Init.

Not all edge cases may be fully worked out yet, especially with the param support. So let's consider this WIP, and it's here for you to see, and maybe (hopefully) also to play with.

The intention is to use a pair of ELRS receivers as SiK replacement. 
This was possible already with the FrsKy receivers, since these provide additional pins, which could be used for in & out. The challenge with the ELRS receivers is that they only have Tx,Rx.


rotorman:
Just as an info regarding the above Express LRS receiver ports limitation: the RadioMaster RP4TD (with ESP32) has 2 accessible serial ports: https://www.radiomasterrc.com/products/rp4td-expresslrs-2-4ghz-diversity-receiver

olliw42:
> Just as an info regarding the above Express LRS receiver ports limitation: the RadioMaster RP4TD (with ESP32) has 2 accessible serial ports: https://www.radiomasterrc.com/products/rp4td-expresslrs-2-4ghz-diversity-receiver

YES ... that's why I bought it :D :D :D

we currently however have only esp82xx working stable, and many other ELRS receivers don't have these pads

in some sense it's also just to start getting into tx, and I wanted to try this mavlink params thing for a while :)

tmcadam:
This is very cool!! How would I send a mavlink command to change a tx or rx param?

olliw42:
I have tested it with MissionPlanner ... sadly MP has a bit it's weirednesses, but when knowing and respecting them it's not so bad

have rx and tx connected, on the serial of the tx attach a usb-ttl adapter and connect to MissionPlanner, now connect in MissionPlanner, and you should see now TWO components, one the fight ontroller, and one the tx module (show up as telemetry radio). Select, the radio, go to parameter page, and click "refresh params", this should download the parameters

tmcadam:
I have tested with 2 x generic 900 rx and is working very nicely. Very cool!!!!

image

Only issue I can see is some inconsistent behavior around the led being off or on when bound i.e. sometimes its off, sometimes it is on. 

This is really a good entry to the TX game. I will try and run this now on esp32 tx hardware.

olliw42:
COOL! THX for testing!

> Only issue I can see is some inconsistent behavior around the led being off or on when bound i.e. sometimes its off, sometimes it is on.

what do you mean? you enter bind and the led does not blink?

> Also, are the options and description able to be defined, so they are visible in MP parameter list?

AFAIK yes, but not soo easily. Maybe we should re-research.
(I had proposed a generic MAVLink way many years ago, but sadly it never took off)

tmcadam:
> COOL! THX for testing!
> 
> > Only issue I can see is some inconsistent behavior around the led being off or on when bound i.e. sometimes its off, sometimes it is on.
> 
> what do you mean? you enter bind and the led does not blink?
> 
> > Also, are the options and description able to be defined, so they are visible in MP parameter list?
> 
> AFAIK yes, but not soo easily. Maybe we should re-research. (I had proposed a generic MAVLink way many years ago, but sadly it never took off)

When bound, sometimes the led is solid on, sometimes it is solid off. The blinking is fine.

olliw42:
what do you mean with "bound" ??? 
it should be always bound to establish a connection in the first place
also, led behavior should be independent on this mavlink thing

usually, if led changes from solid to something else, it means that the tx-rx disconnect. is this that, is your link not stably connected?

tmcadam:
> what do you mean with "bound" ??? it should be always bound to establish a connection in the first place also, led behavior should be independent on this mavlink thing
> 
> usually, if led changes from solid to something else, it means that the tx-rx disconnect. is this that, is your link not stably connected?

Sorry, I didn't explain well. If I switch on the tx and rx switched off, then it blinks slowly. I switch on the rx, the blinking stops and the led goes off. But the tx and rx are connected, working fine.

olliw42:
> If I switch on the tx and rx switched off, then it blinks slowly. I switch on the rx, the blinking stops and the led goes off. But the tx and rx are connected, working fine, just led is off (should be on in this state).

ah, looks like a bug ... has nothing to do with mavlink params, right?

tmcadam:
> > If I switch on the tx and rx switched off, then it blinks slowly. I switch on the rx, the blinking stops and the led goes off. But the tx and rx are connected, working fine, just led is off (should be on in this state).
> 
> ah, looks like a bug ... has nothing to do with mavlink params, right?

Correct

Also, LQ was good 100% 0 packets lost in MP

olliw42:
this is really strange 
just looked at the led code, and I can't see anything which would explain this. Also, I can't see why it should be different for the tx than it is for the rx. 
The rx's led works as expected, right?

> If I switch on the tx and rx switched off, then it blinks slowly. I switch on the rx, the blinking stops and the led goes off.

Is this repeatable, i.e., is it always so that when you first power up the tx and then th erx, the led on the tx goes from blink to off?
Implication, when you first power up rx and then tx, led goes to on.
Or is this more random?

mLRS/Common/hal/esp/tx-hal-generic-2400-pa-esp8285.h
```diff
@@ -0,0 +1,127 @@
...
+//-------------------------------------------------------
+
+#define DEVICE_HAS_SINGLE_LED
+//#define DEVICE_HAS_JRPIN5
+//#define DEVICE_HAS_IN
+#define DEVICE_HAS_NO_COM
+#define DEVICE_HAS_NO_DEBUG
+//#define DEVICE_HAS_NO_SERIAL
+
+// #define USE_FEATURE_MAVLINK_PARAMS // has no CLI, no Lua, hence needs this
```
jlpoltrack, line 20: Should this be defined here?  Currently it is commented out?

olliw42:
> Sorry, I didn't explain well. If I switch on the tx and rx switched off, then it blinks slowly. I switch on the rx, the blinking stops and the led goes off. But the tx and rx are connected, working fine, just led is off (should be on in this state).

I cannot reproduce it, and I also cannot see anything in the code which would suggest this.

I start to think this is not a LED thing but some other problem with your system.

mLRS/Common/hal/esp/tx-hal-generic-2400-pa-esp8285.h
olliw42: that's for the moment
I'm not yet sure how we should do that, if it should be available for all modules, just those which can't have oled,cli,jrpin5, or what ...

olliw42:
kind of obsolete, hence closing

PR #158, opened 2024-04-16
esp, rx enable true diversity

jlpoltrack:
Enable true diversity on ESP32 True Diversity Receivers, bunch of optimizations as well.


olliw42:
it has the workflow yml in it, accident or on purpose?

jlpoltrack:
> it has the workflow yml in it, accident or on purpose?

Accident - I will remove, so it can be handled separately

tmcadam:
I would be happy to merge this based on your own testing @jlpoltrack. Better to have it in `dev-esp-rx-td` as soon as as possible. The real testing can take place on that branch, and code review in PR#156 ;-).

@olliw42 Anyway we can hold onto the commit history on `dev-esp-` branches. Maybe the merge setting is for the whole repo?

olliw42:
> @olliw42 Anyway we can hold onto the commit history on `dev-esp-` branches. Maybe the merge setting is for the whole repo?

maybe, probably, I have no clue ... I wouldn't want to do too many complicated setups and configs in a too sort time ... I'm pretty conservative on such points since I just don't want to spend my time on such things ... I like to code, make progress, improve, add features ... not setting up repos etc pp ... so, in some sense I don't want to have a clue here :D
suggestion: lets go on and KIS
:) :)

PR #160, opened 2024-04-18
rx, tx SX127x, rfo define

jlpoltrack:
New define 'SX_USE_RFO' for SX127x hardware that uses the RFO pin instead of the PA_BOOST pin for output.  Tested by Joost on BR3.


jlpoltrack:
I prefer the shorter name since that's how the datasheet refers to these pins: 

image

No rush to merge.

PR #163, opened 2024-04-21
dev-esp-rx-td, some tidies to branch dev-esp-rx-td, nfc

olliw42:
this does some tidies and cleanups
they all should be nfc
the large majority are really just formatting, blanks, etc.
the major changes were done to the use of DIVERSITY_SINGLE_SPI. It seems to me the driver had been essentially identical and needed only minor change, so I put it into one driver.
tested to not have changed the working on a RP4TD


jlpoltrack:
Tested on SuperD after erasing flash, merging.

- FC connection works
- EEPROM works
- FLRC works

olliw42:
THX !!

PR #165, opened 2024-04-21
esp. sync with other branches, tidies

olliw42:
brings in to main some things from the dev-esp-rx-td branch, which were considered good there
mostly tidies, hence nfc
but also few nominally non-nfc things

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -23,22 +23,22 @@ monitor_speed = 115200
 platform = espressif8266@4.2.1
 board_build.f_cpu = 160000000L
 build_type = release
+build_flags = -Os
 
-; RX
+;-- RX --
 
 [env_common_rx]
 build_src_filter =
   -<*>
   +<CommonRx/mlrs-rx.cpp>
   +<CommonRx/out.cpp>
-  +<Common/common_types.cpp>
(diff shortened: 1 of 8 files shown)


platformio.ini
```diff
@@ -23,23 +23,23 @@ monitor_speed = 115200
 platform = espressif8266@4.2.1
 board_build.f_cpu = 160000000L
 build_type = release
+build_flags = -Os
```
jlpoltrack, line 26: If we're thinking about this, I would also suggest increasing the flash speed to 80 MHz (qio is not supported).

olliw42: I was thinking that -Os is the default, isn't it? So that this is actually nfc and just there to have a build_flags entry here ....

jlpoltrack: My bad :) I read as 'O2'

I would say O2 is one thing that can be explored (ELRS does this) https://github.com/ExpressLRS/ExpressLRS/blob/master/src/targets/common.ini#L122

jlpoltrack:
Tested on Matek ELRS-D with Generic 2400 PA + FC.  Looks good.

olliw42:
MANY THX !!
so merging

PR #166, opened 2024-04-21
dev-esp-rx-td, more tidies, mostly nfc

olliw42:
title says it, more tidies from my side
- the large majority are nfc, and really just formatting
- in esp-uartb.h I have replaced some inline by static inline (Note: inline is useless, it needs to be static inline)
- in platformio there was a bug, in there a esp826xx target was taking esp build flags
- added RM RP4TD as selected target device
- changed things in the dev targets, hope I got that correct


jlpoltrack:
Looks good to me:

- All builds compile
- Tested with SuperD connects, can change params and serial hooked up to FC looks good.

PR #168, opened 2024-04-22
dev-esp-rx-td, hopefully last set of tidies II

olliw42:
title says it

also added a new target 2400-pa-dcdc, is consequence of me testing the betafpv nano 2.4G receiver ... appears to work equally well without and with DCDC, let's see. Thought better to have it for now.
We could do the same approach of a dcdc target for the betafpv-superd-2400, but can do this later too LOL

tested on betafpv nano 2.4G and nano 900 MHz

### DIFF
--- a/mLRS/Common/hal/esp-glue.h
+++ b/mLRS/Common/hal/esp-glue.h
@@ -19,11 +19,6 @@ void __disable_irq(void) {}
 void __enable_irq(void) {}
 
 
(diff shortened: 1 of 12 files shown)


jlpoltrack:
Looks good, will merge.

Tested on SuperD - binds, connects, serial to FC works.

olliw42:
THX !!

PR #169, opened 2024-04-23
stm, fix GCC protection

jlpoltrack:
Title


olliw42:
uuuppsss ... how did that happen
THX !!

PR #170, opened 2024-04-30
Add RadioMaster BR3

jlpoltrack:
- Adds RadioMaster BR3 (ESP8285) 868/915 MHz 500 mW Receiver
- Validated RF Power
- MAVLink looks good when hooked up to FC

image


olliw42:
can't it be a generic layout?
I'm not super fond of adding it as "selected device". First, we do have already plenty. Second, it's antenna diversity, and thus a bit clumsy and non-ideal (and expensive) to really promote.

jlpoltrack:
> can't it be a generic layout?

It has a unique target in ELRS, but am fine to update hal and device_conf so that it lives in the 'generic' section if that's what you're thinking.

olliw42:
how unique is the layout as compare to a generic layout?

jlpoltrack:
> how unique is the layout as compare to a generic layout?

Somewhat different:

- TX enable pin (generic 900 PA doesn't have)
- Antenna select pin
- RFO define
- Different gain than generic 900 PA

olliw42:
many thx

hm, I think we need to think about what we want to do with the situation, to me it seems that the generic concept doesn't really work out, there seem to be way too many exceptions ...
I'm not sure if we shouldn't just create a firmware for any device out there ...

jlpoltrack:
I would propose an alternative :) since the above approach would require a rather large table to be maintained in the Docs (I think we now support over 50 boards).

- In the mLRS repo:
    - Have a 'Generic' and 'Device-Specific' section.   'Selected' hardware could be a subset in the 'Device-Specific' section or another section.
- In the mLRS docs:
    - The 'Supported' list would mention any device-specific boards that don't use a generic target as a base (e.g. RM BR3) in addition to generic targets.
    - The 'Selected' list would only include boards that we choose - with the thought that just because there is a device specific target doesn't mean that board would end up in the list (e.g. RM BR3)

olliw42:
interesting ideas
so I'll add my thoughts

I think the mixture of different "classes" will just be confusing to any users, if they are not pretty much self explaining. IMHO the docs and the repo are linked in that the suitable pre-compiled binary should be easy to find and identify.
So far, the idea was to have only generic, and very few selected devices, with implication that there would be only few pre-compiled files with device names, and all other with generic in their name.

However things like BR3 adds a new variety, and it IMHO really makes it hard to understand why for some device one should have to look up a list to find the corresponding generic file while for others one just has to look for the name, among which some are selected and some are not.

This made me think that we should go to simply generate named pre-compiled binaries for basically every device. Could be added by hand, since the current approach does allow already to use one hal file for multiple devices or use some automated approach. The doc could also be automated, or even simpler, in the doc only the selected are explicitely mentioned and for the other it is simply said to look into the folder and pick the correct one.

Either way, I would not add info like tested/not tested, since this most likely will be just a mometary thing, since we likely won't be able to retest em all with every firmware update,

mLRS/Common/hal/device_conf.h
```diff
@@ -386,6 +386,14 @@ The default selection of frequency bands can be overruled by feature defines.
   #define FREQUENCY_BAND_915_MHZ_FCC
 #endif
 
+#ifdef RX_ELRS_RADIOMASTER_BR3_900_ESP8285
+  #define DEVICE_NAME "RadioMaster BR3 900"
```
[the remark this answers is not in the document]
jlpoltrack: Updated

platformio.ini
```diff
@@ -14,6 +14,7 @@ src_dir = mLRS
 [env]
 framework = arduino
 monitor_speed = 115200
+upload_speed = 921600
```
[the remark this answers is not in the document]
jlpoltrack: Default for ESP32 is 921600 and ESP8266 is 230400 - but never had a problem flashing ESP8266 at faster speed.  Fine to remove if we want.

jlpoltrack:
Merge conflicts now resolved.  BR3 entries in the generic section.

mLRS/Common/hal/device_conf.h
```diff
@@ -381,4 +381,4 @@ The default selection of frequency bands can be overruled by feature defines.
 #define USE_FEATURE_FLRC
 #if defined TX_FRM303_F072CB || defined RX_FRM303_F072CB // is short of RAM for tx, and possibly too slow
   #undef USE_FEATURE_FLRC
-#endif
+#endif
```
olliw42, line 384: nit pick
could you please add the line so this file goes away from the changed files :)

mLRS/Common/hal/hal.h
```diff
@@ -429,4 +429,4 @@ Note: Some "high-level" features are set for each device in the device_conf.h fi
   void fiveway_init(void) {}
 #endif
 
-#endif // HAL_H
+#endif // HAL_H
```
olliw42, line 432: dito

platformio.ini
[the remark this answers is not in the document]
olliw42: aha
I'll leave it to you guys to decide :)

mLRS/Common/hal/device_conf.h
jlpoltrack: Huh - I copied these files directly from Main...

What's the difference?

Main:

image

My Branch:

image

jlpoltrack: Resolved

mLRS/Common/hal/hal.h
jlpoltrack: Resolved

PR #171, opened 2024-05-01
esp, separate out device_config.h, hal.h, streamline diy boards

olliw42:
@tmcadam @jlpoltrack 

the number of esp targets seem to grow, and I alos have an impression that we are rather moving towards supporting any kind of elrs device, even devices which may not be the most reasonable choice such e.g. the BR1 or BR3, I suggest to separate outthe esp target definitions into their own files ... 

in some future we potentially may consider making them even more canonical or even auto generated, as the esp hal files seem to be very redundant

the pr also renames the diy/dev boards to a canoncial naming scheme. A canonical naming schem can be helpful, e.g. in the run py script for the stm32 targets this allows a more fine grained selection of what to do.

This PR should come before https://github.com/olliw42/mLRS/pull/170, and https://github.com/olliw42/mLRS/pull/170 would have to be rebased accordingly.


platformio.ini
```diff
@@ -170,7 +170,7 @@ build_flags =
 
 ;-- ESP DIY boards --
 
-[env:rx-diyboard-900]
+[env:rx-diy-dev-900-esp8266]
```
tmcadam, line 173: I think this makes sense. I have been thinking about changing this as well.

tmcadam:
I like this. These files were getting a bit big to work with.

olliw42:
I suppose @jlpoltrack has seen it by now and ddin't disagree with you. So, based on approval by one esp dev I merge. :)

THX !!

PR #174, opened 2024-05-04
dev-esp-br3-power

jlpoltrack:
- Update SX127x to use 11.4 dBM max power when using RFO
- Update BR3 power levels accordingly


olliw42:
THX!

PR #175, opened 2024-05-04
dev-esp-RP4TD-out-2

jlpoltrack:
Adds out functionality to RadioMaster RP4 TD receiver.  Needs to be verified on RP4 hardware.


mLRS/CommonRx/out_interface.h
```diff
@@ -33,7 +37,9 @@ class tOut : public OutBase
     {
         OutBase::Init(&Setup.Rx);
         out_init_gpio();
+#ifndef ESP32
```
olliw42, line 40: this is not so nice
three possible suggestions
- add this function to the uart lib and let it just do the normal init (i.e. add to uart_template.h)
- add an empty prototype in the top of this file where there is the branch for esp32 anyway
- add an empty prototype to the hal file
I guess I like (1) best

olliw42:
very cool

the invert/normal does work ??

olliw42:
so, tested, seem good. Merging.

GREAT JOB. THX!!

PR #177, opened 2024-05-07
esp, esp lib improvements, comments&tidy

olliw42:
some few additions/changes to the esp lib files, and a comment
this brings over some stuff from the owpr-esp-tx branch
- less IRAM static inline
- gpio handling for pin >= 32
- some comments nfc

should be a light review :)


jlpoltrack:
Will test more later, but just a quick IRAM check -no change: 

SuperD, current main:

image

SuperD, this branch:

image

I believe the reason for this is because the configs in platformio.ini are taking precedence.

olliw42:
very cool, thx!
just did another small commit to change y to 1 in the platformio.ini ... should have been changed to 1 already, not sure where this was scrambled up...

tmcadam:
Looks fine to me. I can test on hardware tonight if you would like another hardware test? If not I am happy to approve as is.

olliw42:
you guys could merge?

jlpoltrack:
Seems like everyone is good - will merge.

PR #179, opened 2024-05-09
tx aetr default

jlpoltrack:
AETR is default order for OpenTx, EdgeTx and AP - so I think this would mean one less setting for folks to change.

### DIFF
--- a/mLRS/Common/common_conf.h
+++ b/mLRS/Common/common_conf.h
@@ -49,7 +49,7 @@
 
 #define SETUP_TX_CHANNELS_SOURCE        1 // 0: none, 1: Crsf (pin5), 2: In (In or pin1), 3: mBridge (pin5)
 
-#define SETUP_TX_CHANNEL_ORDER          CHANNEL_ORDER_ETAR
+#define SETUP_TX_CHANNEL_ORDER          CHANNEL_ORDER_AETR
 
 #define SETUP_TX_IN_MODE                0 // 0: IN_CONFIG_SBUS, 1: IN_CONFIG_SBUS_INVERTED
 


olliw42:
somehow not for me ... ;)

jlpoltrack:
Okay, I can close - out of curiosity do you know where this comes from?  Spektrum is TAER...

geofrancis:
Does it matter on a module? It doesn't know what the channel is doing, as it just sees channels 1-16.

olliw42:
> Okay, I can close - out of curiosity do you know where this comes from? Spektrum is TAER...

well, it probably comes from that I never really understood how this thing works on opentx with the internal 4in1 module ... when I got my first otx radio few years back I was in fact using spektrum ... and I found it just chaotic ... I was glad I eventually got something which just was working for me 

so might be really just my stupidity ...

I guess what I should do now is to rework all my moddels since I don't have anything on spektrum anymore ... (but I'm too lazy LOL)

jlpoltrack:
> Does it matter on a module? It doesn't know what the channel is doing, as it just sees channels 1-16.

Yes, because mLRS will move things around based on:

- Tx Ch Order
- Rx Ch Order

olliw42:
> Does it matter on a module? It doesn't know what the channel is doing, as it just sees channels 1-16.

it's unfortunately not as simple as that ... 
it affects eg failsafe, and if e.g. you want one vehicle to work with two radios which are setup internally differently ecasue on you set up for spektrum with the internal 4in1 and another one for something else ... it' not like everyone has a radio for each vehicle ;)

I spend quite a bit of time on this in the early days. And my conclusion is that you need a way on each side to convert to a ota standard.

jlpoltrack:
> internal 4in1

Interesting - the only reference online I could find for ETAR was for Orange (HobbyKing) hardware - so maybe this is the link :)

olliw42:
not orange specific

geofrancis:
I have never seen a ETAR layout on any radio. to get it as default you would have had to specifically select it when flashing your radio or it should have defaulted to AETR as that's the default setting for opentx and edgetx, ardupilot, inav and betalfight.

olliw42:
well, frankly, when using "never" you make quite some claim ... sounds like you are the one who should be the know it all ... it seems I have managed to make multiprotocol do something which it can't do, what a wizzard I am ;)

having said this, I don't object the notion that AETR is probably the more expected default

olliw42:
lgtm :)

geofrancis:
I have due to working as the main technician at a drone retailer doing hundreds of custom builds over 4 years I have actually setup every radio system on the market until around 7 years ago.

PR #181, opened 2024-05-13
dev MspX

olliw42:
do not merge
for those who want to play with iNav and see if some robuistification to the MSP protocol is worth it


tmcadam:
I'm planning to test this very soon!

Will it work with one of the lua telemetry widget (inav or yapuu)?

olliw42:
> I'm planning to test this very soon!

this would be very cool

> Will it work with one of the lua telemetry widget (inav or yapuu)?

inav: I can't tell. I just don't know how that app works.

yapuu: not really, you'll find the code fragments in this branch which makes it showing the attidue in the hud and also recognizes armed/disarmed, but that's it. 

I actually kind of abandoned the idea to get yapuu working since it's too complex to start with. I'm working on getting the standard CRSF telemetry elements at least. If you happen to have a vehicle with battery and gps I can inform you and update the branch once I have that a bit better going.

olliw42:
@tmcadam I've updated the branch to include CRSF telemetry
I could not test it really, so some conversions might be wrong, I hope that a test with a reall vehicle should tell what needs to be corrected, b ut in pronciple all CRSF telemetry sensors should be there
the only cathc I'm aware of at this point is the that mode field FM isn't showing the correct mode, I still haven't fully worked this out (https://discord.com/channels/1005096100572700794/1237748187662647327/1241327812439179326)

olliw42:
ok, so the branch got much improved
- did some "dirty" workaround for gettijng the flight modes to tx module & radio. Seems to work fine though
- added compression to the MSP_BOXNAMES message ... this is the message which is 340 bytes long (in my case) ... and which almost never will get fully thoruigh if the LQ is not stable at 100% ... now the message is compressed down to 45 bytes ... so, much much more likely to get through 

some real world testing on a real vehicle is what is missing though

olliw42:
@tmcadam you were asking if an app like inav app would work ... just brioefly looked at it, and it seems to me it gets all it's data from lua getValue() ... this means that it actually should work now ... at least as much as it can work with a CRSF system (CRSF has not that many telemetry items, so some dta fields won't work) ...
:)

PR #182, opened 2024-05-16
Dev esp32 900 td + SX127x diversity support

jlpoltrack:
- Adds Generic 900 True Diversity PA Target
- Updates SX127x Driver to allow for diversity
- Functionally tested by Enzaii on Discord
- Power uses values from ELRS


mLRS/Common/sx-drivers/sx127x_driver.h
```diff
@@ -430,9 +430,153 @@ class Sx127xDriver : public Sx127xDriverCommon
...
+
+        SetStandby(); // should be in STDBY after reset
+        delay_us(1000); // is this needed ????
+    }
+
+    //-- high level API functions
+
+    void StartUp(tSxGlobalConfig* global_config)
+    {
+//XX        // this is not nice, figure out where to place
```
olliw42, line 548: you can delete these comments here, it's sufficient to have them in teh driver above :)

olliw42:
looks excellent, just one possible minor adjustment

when you or Tom merge, pl take care to change the commit title to something starting with `esp, ...`

(it affects a non-esp library, but I think that's so canonical I would consider this a pure esp merge :)

MAYN THX !!

mLRS/Common/sx-drivers/sx127x_driver.h
jlpoltrack: Removed comments from StartUp function

tmcadam:
Looks great. This looks like a really interesting piece of hardware to add. 

Great to get end user test on this one too!

jlpoltrack:
@tmcadam Could you kindly do the merge if you approve? (Try not to merge my own PRs :) )

tmcadam:
@jlpoltrack Done!

PR #183, opened 2024-05-21
dev-esp-iflight-2.4

jlpoltrack:
- Adds two targets: iFlight 250 mW True Diversity 2.4G, iFlight 500 mW 2.4G
- Cleanup of platformio.ini to align with esp-hal and esp-device_conf

### DIFF
--- a/mLRS/Common/hal/esp/esp-device_conf.h
+++ b/mLRS/Common/hal/esp/esp-device_conf.h
@@ -76,6 +76,20 @@
   #define FREQUENCY_BAND_2P4_GHZ
 #endif
 
+#ifdef RX_ELRS_IFLIGHT_ELRS_TD_2400_ESP32
+  #define DEVICE_NAME "IFLIGHT 2400 TD PA"
+  #define DEVICE_IS_RECEIVER
+  #define DEVICE_HAS_SX128x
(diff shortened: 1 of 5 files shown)


jlpoltrack:
> you did confirm the power levels yourself, or Joost did?

Power levels are unconfirmed but use ELRS values.  I'm pretty sure these use the same RF Frontend as the E28-2G4M27S based on what I see in the ELRS repo.

olliw42:
in discord you presented a power level chart, did show consistently 9 dB too much. 
so, this here is correcting for the 9 dB, right?
any reason not to double check (from the chart I conclude you have power meter :)) 

anyway, I'm just nitpicking, as always. If you guys want to merge you have my lgtm :):):)

please do not forget to change the commit title to something starting with "esp, ....."

jlpoltrack:
Well the challenge is that I don't have this hardware (and am also not sure if Joost has power meter) :)

olliw42:
ah ... so the power level chart was based on what you know, not what was actually measured ...

I guess we could ask Joost to retry with the new hal firmware, and he when at least should not see issues in the highest pwoer level, right?

olliw42:
if I read discord correctly, it was just confrimed to be all good

from my side hence: pl go ahead with merging. lgtm

MANY THX!!!!

PR #187, opened 2024-05-25
tools, py script to flash receivers via fc passthru, based on brad's

olliw42:
title says it
replaces @brad112358 's https://github.com/olliw42/mLRS/pull/186
requires @brad112358 's PR to ArduPilot https://github.com/ArduPilot/ardupilot/pull/27152


brad112358:
Are we actually allowed to redistribute this ST code?

brad112358:
Never mind, it looks like it is allowed.

brad112358:
Another thought... Can this same script be expanded to do the passthrough setup and perhaps the flash for esp devices as well?  Just need to hold the boot button on power up.  Perhaps detect esp device automatically somehow and choose appropriate flashing tool (by noticing reboot shutdown is not supported?)

olliw42:
I was thinking about it too
I just don't see how this would be reasonably possibly with the esp ... I mean setting ap in passthru, easy, but then??
if you hold down button at power up ... not sure if the boot would reliably survive until the ap is set into passthru, I would think any e.g. heartbeat from the fc would confuse it
to test the simplest would be to just add a -esp parameter

brad112358:
For some reason, the mavlink messages did not present a problem when I tested passthrough with programming the esp receiver I have integrated into my flight controllers.  If it works with manual config of passthrough via QGC, it should work with the script.

olliw42:
add an option to only do passthru to the script
for you to play with and test :)

olliw42:
you can give it a -esp option
or in the gui uncheck systemboot

tools/flash_rx_via_ap_passthru3.py
```diff
@@ -0,0 +1,428 @@
...
+        if int(msg.param_value) == 2: # prtocol = MAVLink2
+            serial_list.append('SERIAL'+str(i))
+            if i == int(serialx):
+                serialx_in_list = True
+
+    if not serialx_in_list:
+        printError('ERROR: SERIAL'+str(serialx)+' is not MAVLink2 protocol!')
+        exit(1)
+
+    baud = ardupilot_read_baud(link, 2)
```
brad112358, line 135: The hard coded 2 here should be serialx

olliw42:
closing.
there is a branch https://github.com/olliw42/mLRS/tree/dev-flashscript

PR #188, opened 2024-05-28
dev ESP32 ESP-IDF UART Driver

jlpoltrack:
@tmcadam Maybe give a test and share your thoughts - functionality at this point should be identical to current Arduino implmentation.


olliw42:
THX for having brought this to us anyway!! :)

PR #193, opened 2024-06-10
esp, lib, tdiy uart lib

olliw42:
@tmcadam or @jlpoltrack 

title says it

largely just nfc changes

Two changes relevant to EPS32:
- check of txbuf size, this is what is concluded in the tx module attempt branches
- no tx flush
These were concluded in the tx branches using the esp-idf. Since it's used lowlevel also here I think this shoulöd also apply here,

### DIFF
--- a/mLRS/Common/hal/esp/rx-hal-radiomaster-rp4td-2400-esp32.h
+++ b/mLRS/Common/hal/esp/rx-hal-radiomaster-rp4td-2400-esp32.h
@@ -23,7 +23,7 @@
 
 #define UART_USE_SERIAL1 
 #define UART_BAUD                 416666   // CRSF baud rate
-#define UART_USE_TX_IO            18       // t pad on the receiver
+#define UART_USE_TX_IO            IO_P18   // t pad on the receiver
 #define UART_USE_RX_IO            -1       // no Rx pin needed
 #define UART_TXBUFSIZE            256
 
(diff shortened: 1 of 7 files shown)


mLRS/Common/hal/esp/rx-hal-radiomaster-rp4td-2400-esp32.h
```diff
@@ -17,13 +17,15 @@
 
 
 //-- UARTS
+// UARTB = serial port
 // UART = output port, SBus or whatever
+// UARTC = debug port
 
 #define UART_USE_SERIAL1 
 #define UART_BAUD                 416666   // CRSF baud rate
 #define UART_USE_TX_IO            18       // t pad on the receiver
```
jlpoltrack, line 26: Probably want IO_P18 here

jlpoltrack:
Looks good to me - tested on the bench with SuperD:

- MAVFTP looks good both way (parameters and mission upload)
- Good in steady state conditions, 50 Hz with FLRC stream rates

mLRS/Common/hal/esp/rx-hal-radiomaster-rp4td-2400-esp32.h
olliw42: ups... corrected. THX!

olliw42:
many thx, guys!

jlpoltrack:
Will merge in a few hours unless someone chimes in with a reason not to.

PR #195, opened 2024-06-13
esp, platformio.ini and hal, tidy up a bit

olliw42:
- annotates the platformio.ini, and also rearranges a bit
- adds POWER_OVERLAY flag to make hals for iflight's 2400 simpler


platformio.ini
```diff
@@ -167,53 +169,53 @@ build_flags =
   -D RX_ELRS_GENERIC_2400_TD_PA_ESP32
 
 
-[env:rx-iflight-elrs-td-2400]
-extends = env:rx-generic-2400-td-pa
+;-- generic 2.4 GHz ELRS boards with overlays or other variations
+
+[env:rx-generic-2400-pa-dcdc]           ; like generic 2400 pa, but sx uses DCDC mode
```
jlpoltrack, line 174: May make since to remove as no hardware exists for this target.

jlpoltrack:
Looks good, especially the iFlight stuff - just one comment.

platformio.ini
olliw42: DONE. Thx

olliw42:
removed the dcdc target

did a bit more comments, rearrangment and tidying

did also add a target for iflight 900 ... it's equal to generic, but I really like this device so I thought we may want to give it it's own name, so a precompiled is available like for the ifflight 2400

jlpoltrack:
Looks good to me, will merge later today.

olliw42:
I think we should mark it with a "docu needed" label (to add the iflight900) ...

platformio.ini
```diff
@@ -167,53 +175,48 @@ build_flags =
...
-[env:rx-iflight-elrs-td-2400]
-extends = env:rx-generic-2400-td-pa
-build_flags =
-  ${env_common_rx_esp32.build_flags}
-  -D RX_ELRS_IFLIGHT_ELRS_TD_2400_ESP32
-
+;-- generic 2.4 GHz ELRS boards with overlays or other variations
 
-[env:rx-iflight-elrs-2400]
+[env:rx-iflight-elrs-2400]              ; like generic 2400 td pa, but provides 250 mW instead of 100 mW, has its own name
```
jlpoltrack, line 180: Comment should read

'like generic 2400 pa, but provides 500 mW instead of 100 mW, has its own name'

jlpoltrack:
Docs PR added - let me know if you're thinking the iFlight 900 True Diversity should go to the selected section (it sounds like you have this hardware on hand)

platformio.ini
olliw42: ups. THANKS! done.

olliw42:
> let me know if you're thinking the iFlight 900 True Diversity should go to the selected section (it sounds like you have this hardware on hand)

yo, I do have the 2400 and 900 TD now, arrived few days ago. I really like them, hence I thought the 900 should also get it's own target, but the implication was not to advance them to the selected area.

PR #197, opened 2024-06-16
esp, rx, iflight900, swap sx, adjust power

olliw42:
title says it

- swaps sx and sx2, so that antenna1 is on left ufl, like for the iflight2400 td 250mW
- lowers the power. gives proper values for 100 mW and 250 mW. For 500 mW setting it's lower by 50 mW, but it was low anyway. It appears the PA has reached already nonlinear region, and I don't think one should go too high into it

bench tested on iflight 900 500 mW receiver

I'm not sure we should keep the iflight900 in the docu list. Mine does not appear to work reliable. I cannot change settings without the thing going into continous reset upon next power up. No idea what this is. Also, the transmission appears to be flimsy, when I move or touch it the radio sometimes loses like 1/2 sec worth of packets. Could be the antenna or the connectors, or the hardware itself, don't know, but from the bench behavior I don't put any trust in mine. It also gets surprisingly warm at 100 mW setting ...

### DIFF
--- a/mLRS/Common/hal/esp/rx-hal-generic-900-td-pa-esp32.h
+++ b/mLRS/Common/hal/esp/rx-hal-generic-900-td-pa-esp32.h
@@ -31,17 +31,18 @@
 
 
 //-- SX1: SX12xx & SPI
+// antenna1 = left ufl
 
-#define SPI_CS_IO                 IO_P27
+#define SPI_CS_IO                 IO_P13
 #define SPI_MISO                  IO_P33
 #define SPI_MOSI                  IO_P32
 #define SPI_SCK                   IO_P25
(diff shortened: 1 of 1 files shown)


jlpoltrack:
One quick thought - does it make sense to limit to 250 mW (if it can't reach 500 anyways and runs hots?)

olliw42:
was wondering too

olliw42:
anything I should/need to do?

jlpoltrack:
Let's just decide what to do with the Docs:
- Leave as is
- Remove
- In notes, say only 350 mW power output

image

olliw42:
I guess I would prefer 3 too

PR #198, opened 2024-06-16
esp, rx, iflight2400 250mW, adjust power levels

olliw42:
title says it

bench tested on an iflught 2400 250mW


jlpoltrack:
Looks good, merging.

Issue #200, opened 2024-06-20
bug: esp32 diversity, allow antenna2


olliw42:
on the esp32 receivers with diversity, which use single spi, selecting antenna2 will make it crash. This is attributed to how the spi and sx are initialized.
Needs to be corrected however.

brad112358:
Is there any chance this is related to https://github.com/olliw42/mLRS/issues/245 ?

olliw42:
this one here - if I'm not mistaken - is understood. If I recall it is because with only sx2 the spi is not initialized. We would need an underlying level of structure which does the spi init either case, but not twice. We don't have that, and a straigth fiorward apporach would look ugly, in terms of arhcitecture.

jlpoltrack:
Yes, unrelated.  Further - antenna 2 is disabled for all of the ESP32 diversity hardware. https://github.com/olliw42/mLRS/pull/203

PR #202, opened 2024-06-25
common, fix antenna 2 on single spi

jlpoltrack:
To address: https://github.com/olliw42/mLRS/issues/200

Tested on SuperD

### DIFF
--- a/mLRS/Common/sx-drivers/sx128x_driver.h
+++ b/mLRS/Common/sx-drivers/sx128x_driver.h
@@ -559,7 +559,11 @@ class Sx128xDriver2 : public Sx128xDriverCommon
         spib_init();
         spib_setnop(0x00); // 0x00 = NOP
 #else
-        // spi init done already by driver1
+        // spi init only necessary when only antenna 2 selected, otherwise handled by driver1
+        if (Config.Diversity == DIVERSITY_ANTENNA2) {
+            spi_init();
+            spi_setnop(0x00); // 0x00 = NOP
+        }
 #endif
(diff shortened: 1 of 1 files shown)


olliw42:
THX!

I unfortunately don't like it all however. This breaks all abstraction, layering & architecture. The simplest apporach would be let Init have a paramter, like bool initialize_spi, and set that properly upon call in main. 
Structure-wise it's all so nice, but I don't have any good idea either :(

olliw42:
it is actually also surprising it works, since Config is populated in setup_init(); which is called only after sx2.Init(); ....

https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/mlrs-tx.cpp#L274-L277
https://github.com/olliw42/mLRS/blob/main/mLRS/CommonRx/mlrs-rx.cpp#L147-L150

jlpoltrack:
> it is actually also surprising it works

It doesn't - the ESP soft reboot after parameter save made it looked like it did.  Going to close.

jlpoltrack:
One last thought - for the time being should we change the diversity mask for single SPI so that it doesn't allow Antenna 2 to be selected?  Currently if you select Antenna 2 you'd get stuck and would have to wipe / reflash: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/setup.h#L87-L89

olliw42:
I had that thought too but kind of was hoping that the proper solution might found in not too distant future ... but if we don't I guess we should do exactly that, at least as momentry precaution.

PR #203, opened 2024-06-27
common, setup disable antenna 2 on single SPI

jlpoltrack:
Temporary solution to prevent getting stuck if one selects Antenna 2 on single SPI diversity hardware.


olliw42:
THX!!

PR #208, opened 2024-07-22
add matek devices, well tested

olliw42:
title says it
thx to the excellent hal templates of MatekSys, which this is massively baesd on :)


jlpoltrack:
Tx | MR900 works (flashed the pre-built).  Would prefer a lower default power output (currently 24 dBm).  Will test Tx | 2.4 shortly.

MATEKSYS:
LGTM. Thanks

PR #212, opened 2024-08-20
Check python version in run_setup.py + verbose option

tmcadam:
This PR updates `run_setup.py` to make it run more consistently regardless of platform or system setup:

- checks for python or python3
- changes run_setup.py sub-commands to use python (or python3) explicitly
- updates the script to use subprocess module instead of os.system
- adds a --verbose option to run with or without output

### DIFF
--- a/mLRS/CommonRx/mavlink_interface_rx.h
+++ b/mLRS/CommonRx/mavlink_interface_rx.h
@@ -1064,7 +1064,7 @@ void tRxAutoPilot::Do(void)
     uint32_t tnow_ms = millis32(); // we need to get fresh time, since a HEARTBEAT might be received in the main Do loop
 
     if (sysid && ((tnow_ms - heartbeat_tlast_ms) > 2500)) { // we lost connection to our fc
-dbg.puts("\nlost heartbeat");
+//dbg.puts("\nlost heartbeat");
         Init();
         return;
     }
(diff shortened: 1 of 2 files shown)


jlpoltrack:
Took a quick run and code compiles fine (VS Code + STM32 Cube IDE).  Only note - is the lack of print output expected?  I get this on a location where I've previously used the run_setup:

image

Compare to main branch:

image

On a fresh install the submodule update does print a verbose output but the st_drivers and fastmav_generate do not.

tmcadam:
@jlpoltrack 

The output can be enabled with `python .\run_setup.py --verbose`

jlpoltrack:
My bad - would help if I read the entirety of your original post :)

I guess I prefer it to print out all of these things - but that's just a preference.  Otherwise, I have no complaints and this looks good to me.

olliw42:
first, MANY THX
didn't had time yet to test, just looked at it

on first look I'm not so happy with the verbose thing
- could you have a look at the other py's. Here it's done at the end, and such info is passed in as parameters
- I guess I would like -silent better, as for the others, as it rather shoudl be the default to see something. I would expect verbose to show more than default ... 
- I'm not sure how it should come out, but some of the called pys do have a -silent, and it somehow also shoudl be affected by what is choosen for this py ...

maybe we have sevelar silent/verbose levels?

anyway, good first attempt! When I find time to play with it, I may merge this and have a quick follow up PR to "streamline" it.

tmcadam:
@olliw42 I'll have a look

olliw42:
MANYX THX!

PR #213, opened 2024-08-21
esp generic 2400 led logic swap

jlpoltrack:
Tested on RP2 receiver.  Previously, there was no LED when connected.


olliw42:
THX
lgtm

PR #214, opened 2024-08-26
A small bug fix and improve use on Linux

brad112358:
Allow for posix systems to install their own local copy of STM32_Cube_Prog
Don't time out passthrough to allow easier manual use of flash tool 
Fix hard coded serial in call to ardupilot_read_baud

### DIFF
--- a/tools/flash_rx_via_ap_passthru3.py
+++ b/tools/flash_rx_via_ap_passthru3.py
@@ -132,7 +132,7 @@ def ardupilot_connect(uart, serialx):
         printError('ERROR: SERIAL'+str(serialx)+' is not MAVLink2 protocol!')
         exit(1)
 
-    baud = ardupilot_read_baud(link, 2)
+    baud = ardupilot_read_baud(link, serialx)
     if baud != 115200:
         link.close()
         link = mavutil.mavlink_connection(uart, baud)
@@ -285,9 +285,14 @@ def mlrs_flash_firmware_file(uart, baud, file, full_erase_flag):
     arg += ' -w "' + file + '"'
     arg += ' -v'
     arg += ' -g'
-    res = os.system(os.path.join('zmodules','STM32CubeProgrammer','bin','STM32_Programmer_CLI.exe') + ' ' + arg)
+    st_path = os.path.join('zmodules','STM32CubeProgrammer','bin','STM32_Programmer_CLI.exe')
+    if os.name == 'posix': # Assume local install
+        st_path = os.path.join("~",'STMicroelectronics','STM32Cube','STM32CubeProgrammer','bin','STM32_Programmer_CLI')
+
+    res = os.system(st_path + ' ' + arg)
     if res > 0:
         print('ERROR: Flashing firmware failed!')
+        print('Please flash manually and then power cycle')
         exit(1)
     print('------------------------------------------------------------')
     print('firmware flashed')
@@ -334,6 +339,9 @@ if run_gui:
     comport_list = []
     for com in sorted(coms):
         comport_list.append(com.device)
+    if not comport_list:
+        comport_list.append('None Found')
+        print('ERROR: No serial ports found. Please connect your FC')
 
     class Application(Frame):
         def __init__(self, master=None):
@@ -373,7 +381,7 @@ if run_gui:
             else:
                 self.systemboot_value = BooleanVar()
                 self.systemboot_value.set(False)
-            if os.name != 'posix' and not args_esp:
+            if not args_esp:
                 row += 1
                 self.firmware_file = StringVar()
                 self.firmware_file.set('')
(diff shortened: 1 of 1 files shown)


olliw42:
many THX
did not test it and only looked at it briefly.
questions please: 
- it does not change the behavior as regards STM3CubeProgrammer for windows, right?
- this test `if not args_esp` is gone in line 376: this is intentional. i.e., you have tested this what it should be? (I kind of believe I did test it for esp and that's why it was there)
- I don't see what change releates to "Don't time out passthrough to allow easier manual use of flash tool". Could you give me a brief hint?

it lgtm, I just want to have a clearer undertsanding of what the situation is :)

brad112358:
Correct, it should not impact the use of STM32CubeProgrammer for windows.

Good catch on test for args_esp.  I'll fix that.

The timeout change from 30 to 0 is on line 409

olliw42:
> The timeout change from 30 to 0 is on line 409
ah, stupid me. THX.

you have seen in discord that MUSTARDTIGER may have a web app skeleton?

brad112358:
Yes, I hope he can flesh it out to cover all our use cases.  I wonder if he might be underestimating the work involved.  :-)

tools/flash_rx_via_ap_passthru3.py
```diff
@@ -285,9 +285,14 @@ def mlrs_flash_firmware_file(uart, baud, file, full_erase_flag):
...
     arg += ' -g'
-    res = os.system(os.path.join('zmodules','STM32CubeProgrammer','bin','STM32_Programmer_CLI.exe') + ' ' + arg)
+    st_path = os.path.join('zmodules','STM32CubeProgrammer','bin','STM32_Programmer_CLI.exe')
+    if os.name == 'posix': # Assume local install
+        st_path = os.path.join("~",'STMicroelectronics','STM32Cube','STM32CubeProgrammer','bin','STM32_Programmer_CLI')
+
+    res = os.system(st_path + ' ' + arg)
     if res > 0:
         print('ERROR: Flashing firmware failed!')
+        print('Please flash manually and then power cycle')
```
olliw42, line 295: what is this error message supposed to tell?
if it failed why would flashing "manually" work?

tools/flash_rx_via_ap_passthru3.py
```diff
@@ -334,6 +339,9 @@ def mlrs_flash_firmware_file(uart, baud, file, full_erase_flag):
     comport_list = []
     for com in sorted(coms):
         comport_list.append(com.device)
+    if not comport_list:
+        comport_list.append('None Found')
+        print('No serial ports found.  Please connect FC')
```
olliw42, line 344: "ERROR: No serial ports found. Please connect your FC."

brad112358: OK.  If you like that better.  The main thing is not to crash which is what it was doing...

brad112358:
I wonder if we should ask the ardupilot folks to have my passthrough improvement backported.  It's wasn't much work to fix the cherry-pick conflicts when I did it for my personal 4.5 branch.
If we don't have anything better when Ardupilot finally releases it, I think this should get merged into main even if it is just a temporary solution.  It makes flashing things like the mR900-22 and elrs receivers much easier.  If you agree, I could also add a call to esptool to further automate flashing elrs receivers which would be good since we don't support update via WiFi.

olliw42:
I think that's a good idea. One could mentioned that it helps with the matek receivers also.
Probably we want to rename zmodules to something less odd then, like 'bin' maybe

I think I did give esptools a try, but there was some problem with it which made me not add it in this run. Probably just some additional py packages which are not installed per default usually, which would add another layer of inconvenience to a user.

tools/flash_rx_via_ap_passthru3.py
olliw42: yeah, it's just nitpick. I try to be somewhat consistent in the output across all py scripts in the mlrs repo.

tools/flash_rx_via_ap_passthru3.py
brad112358: Because on Linux it may have failed due to not finding the tool in the expected location.

brad112358: On windows, I suppose a manual use of STM32CubeProgrammer may also fix an issue like write protection, etc.

olliw42: frankly, sounds speculative ... IMHO better just to tell what sort of issue occured instead of trying a suggeestion which may not help
this reads as ifthe user is supposed to know what to do and that it will be the unambigious resolution to the occuring issue. I think none is necessarily correct.
IMHO.

brad112358:
esptools worked for me.  Getting tools properly installed is always a sticking point, but for Linux at least, I prefer to just assume the user can deal with it.  I suppose if you want, for Windows, you could figure out a way to include whatever is required like you do for STM32CubeProgrammer.  But I'm probably not the best one to do that work.

olliw42:
the canonical Win user doesn't like to install and even worse to figure her/himself what to install ;)

brad112358:
I looked to see if there was a roadmap for when Ardupilot 4.6 might be released, but I couldn't find anything.  It has been a while since the last feature release and I did find a comment in developer meeting notes mentioning that some backports are getting harder, so perhaps they will bite the bullet soon?

olliw42:
I would think that the beta stage for 4.6 will start in a not too distant future ...

tools/flash_rx_via_ap_passthru3.py
olliw42: ah, So rather something like "ERROR: Flashing firmware failed. Receiver is however still in systemboot and trying to manually flash might work."

PR #217, opened 2024-09-25
matek, mr900-22 dc-dc

jlpoltrack:
Enables the DC-DC converter for the Matek mR900-22.  Tested and works fine - should save about 10mA based on previous tests with Wio E5 Mini.


olliw42:
very nice. MANY THX.

PR #222, opened 2024-10-31
Big Bandit Tx

jlpoltrack:
Attempt to address a few things when using Bandit Micro FW on Big Bandit:

- Flip the Display
- Enable two LEDs

First attempt was to use a separate HAL like RP4TD but redefine of LED functions was ugly.  Figured it may make more sense to keep all in one HAL file.

@tmcadam FYI

### DIFF
--- a/mLRS/Common/hal/esp/esp-hal.h
+++ b/mLRS/Common/hal/esp/esp-hal.h
@@ -96,6 +96,10 @@
 #include "tx-hal-radiomaster-bandit-micro-900-esp32.h"
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
+#include "tx-hal-radiomaster-bandit-micro-900-esp32.h"
+#endif
+
 
 // -- DIY
 
--- a/mLRS/Common/hal/esp/esp-hal.h
+++ b/mLRS/Common/hal/esp/esp-hal.h
@@ -93,11 +93,11 @@
 #endif
 
 #ifdef TX_ELRS_RADIOMASTER_BANDIT_MICRO_900_ESP32
-#include "tx-hal-radiomaster-bandit-micro-900-esp32.h"
+#include "tx-hal-radiomaster-bandit-series-900-esp32.h"
 #endif
 
 #ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
-#include "tx-hal-radiomaster-bandit-micro-900-esp32.h"
+#include "tx-hal-radiomaster-bandit-series-900-esp32.h"
 #endif
 
 
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
@@ -10,7 +10,9 @@
 
 /*
   Flashing ESP8285:
-  - put Tx module into FLASH_ESP mode
+  - change ser dest to serial2
+  - make sure baudrate is 115200 (default)
+  - put Tx module into FLASH_ESP mode from tools
   - Board: Generic ESP8266 Module
   - Reset Method: no dtr (aka ck)
 */
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
@@ -65,8 +65,14 @@
 //#define DEVICE_HAS_NO_COM
 #define DEVICE_HAS_NO_DEBUG
 
-#define DEVICE_HAS_SINGLE_LED
-#define DEVICE_HAS_I2C_DISPLAY_ROT180
+#ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
+    #define DEVICE_HAS_SINGLE_LED_RGB
+    #define DEVICE_HAS_I2C_DISPLAY
+#else
+    #define DEVICE_HAS_SINGLE_LED
+    #define DEVICE_HAS_I2C_DISPLAY_ROT180
+#endif
+
 #define DEVICE_HAS_FAN_ONOFF // board has a Fan, which can be set on or off
 #define DEVICE_HAS_ESP_WIFI_BRIDGE_ON_SERIAL2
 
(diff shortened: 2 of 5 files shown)


tmcadam:
Awesome, I have the big Bandit in 900 and 2400 :+1 . No issues with the RF side of things? Presumably it is quite similar to the micro in that respect.

jlpoltrack:
> No issues with the RF side of things? Presumably it is quite similar to the micro in that respect.

Yea, the designs are pretty similar and differences are primarily the housing and RGB leds.

mLRS/Common/hal/esp/esp-hal.h
```diff
@@ -96,6 +96,10 @@
 #include "tx-hal-radiomaster-bandit-micro-900-esp32.h"
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
+#include "tx-hal-radiomaster-bandit-micro-900-esp32.h"
```
olliw42, line 100: just a thought, should we maybe rename to "...-bandit-series-900-...."

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
```diff
@@ -10,7 +10,9 @@
 
 /*
   Flashing ESP8285:
-  - put Tx module into FLASH_ESP mode
+  - change ser dest to serial2
+  - make sure baudrate is 115200 (default)
```
olliw42, line 14: is this for the tx module, or the setting in Arduino?

I think in the tx module one does not have to use 115200. If right, and it's the ArduinoIDE, I would list it at the end with "Baudrate: 115200" (to make it more clear its's in the IDE

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
```diff
@@ -65,8 +67,14 @@
 //#define DEVICE_HAS_NO_COM
 #define DEVICE_HAS_NO_DEBUG
 
-#define DEVICE_HAS_SINGLE_LED
-#define DEVICE_HAS_I2C_DISPLAY_ROT180
+#ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
+    #define DEVICE_HAS_SINGLE_LED_RGB
```
olliw42, line 71: pl only two blanks for idented defines
also no is ok

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
```diff
@@ -158,16 +166,108 @@ IRAM_ATTR bool button_pressed(void) { return false; }
...
+
 #define LED_RED                   IO_P15
 
-void leds_init(void)
-{
-    gpio_init(LED_RED, IO_MODE_OUTPUT_PP_LOW);
-}
+#ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
+
+    #include <NeoPixelBus.h>
```
olliw42, line 175: pl remove all the identation, just like normal code

platformio.ini
```diff
@@ -285,6 +285,17 @@ build_flags =
...
+  ${env_common_tx_esp32.build_src_filter}
+  +<modules/sx12xx-lib/src/sx127x.cpp>
+  +<Common/thirdparty/gdisp.c>
+build_flags = 
+  ${env_common_tx_esp32.build_flags}
+  -D TX_ELRS_RADIOMASTER_BANDIT_MICRO_900_ESP32
+
+[env:tx-radiomaster-bandit-900] ; Unified_ESP32_900_TX
+extends = env_common_tx_esp32
+board = esp32dev
```
olliw42, line 298: the following lib_deps should be just two blanks ident

mLRS/Common/hal/esp/esp-hal.h
jlpoltrack: Done

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
jlpoltrack: Have changed to 'Upload Speed: 115200' to align with the menu item in the Arduino IDE.

PS, news to me that the Tx module doesn't need this setting to any specific value?

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
jlpoltrack: Removed idents

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
jlpoltrack: Done, cleaned up white space at the top too

platformio.ini
jlpoltrack: Addressed

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-micro-900-esp32.h
olliw42: the com port (when via UART) is always at 115200
doesn't the esp do autobauding when in flash mode?

the baudrate of the serial/serial2 and that coded into the wireless bridge code need to match, true, but I think that's not for flashing ... but better not trust my word but check LOL

jlpoltrack: > but I think that's not for flashing ... but better not trust my word but check LOL

This seems to not be the case.  If I set to tx ser baud to 9600 then I can no longer flash.

olliw42: hm ... maybe 9600 is to slow for the IDE runnig at 115200 to get the timing right ... what woukd happen with 57600, 230400?
it would be sad if those who want to use 230400 on the serial essentially couldn't anymore ...

jlpoltrack: For me, the tx ser baudrate needs to align with the upload speed in the IDE.  That being said the ESP8266 offers interesting options - maybe we allow 921600 baud?

image

olliw42: > For me, the tx ser baudrate needs to align with the upload speed in the IDE to upload

I'm happy to accept this as a fact :)

so we should add a lin like you had before (maybe just making it clearer it's in the tx)

olliw42:
fantastic
looks perfect to me
will merge when I'm on that branch agan later :)

PR #223, opened 2024-10-31
Bandit rx

jlpoltrack:
For consideration, enables the Big Bandit and Bandit Micro to be used as receivers.  Tested out on Bandit with the JST connections on the bottom of the housing.

@tmcadam FYI


olliw42:
THX
but I think we should wait with this to after the next beta (i.e. wait for v1.3.05)
I'm a bit worried about the many changes/additions we are going to have for v1.3.04

mLRS/Common/hal/esp/esp-device_conf.h
```diff
@@ -157,6 +157,30 @@
   #define FREQUENCY_BAND_915_MHZ_FCC
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32
```
olliw42, line 160: this looks like not branched from latest

jlpoltrack:
> I think we should wait with this to after the next beta (i.e. wait for v1.3.05)

Sounds good, am going to close this in the meantime and will re-submit later.

PR #224, opened 2024-11-02
mR900 Power Tweaks

jlpoltrack:
- Enables 50 mW, 100 mW power levels on mR900-30 gear
- Adds 3 dBm power level to bare SX127x to allow for symmetric power output with minimum R9M power level
- Comment tidy

### DIFF
--- a/mLRS/Common/hal/hal-power-pa.h
+++ b/mLRS/Common/hal/hal-power-pa.h
@@ -109,7 +109,6 @@ const rfpower_t rfpower_list[] = {
 
 const rfpower_t rfpower_list[] = {
     { .dbm = POWER_0_DBM, .mW = 1 },
-    { .dbm = POWER_3_DBM, .mW = 2 },
     { .dbm = POWER_10_DBM, .mW = 10 },
     { .dbm = POWER_17_DBM, .mW = 50 },
 };
@@ -142,6 +141,7 @@ const rfpower_t rfpower_list[] = {
 // PA max 30 dBm, PAin max 10 dBm, PA gain 26 dB, LNA gain 16 dB
 #if defined POWER_PA_SE2435L || defined POWER_PA_MATEK_MR900_30 // Matek mR900-30
 #define POWER_PA_DEFINED
+#define POWER_USE_PA_CONFIG_10_DBM
 
 // SX126X power setting can vary from -9 .. 22 for -9 dBm ... 22 dBm
 #include "../setup_types.h"
--- a/mLRS/Common/hal/hal-power-pa.h
+++ b/mLRS/Common/hal/hal-power-pa.h
@@ -109,6 +109,7 @@ const rfpower_t rfpower_list[] = {
 
 const rfpower_t rfpower_list[] = {
     { .dbm = POWER_0_DBM, .mW = 1 },
+    { .dbm = POWER_3_DBM, .mW = 2 },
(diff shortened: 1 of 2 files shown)


olliw42:
> Adds 3 dBm power level to bare SX127x to allow for symmetric power output with minimum R9M power level

hm ... I'm not sure this is worth it
note that doing this change means that the power setting will unnoticedly suddenly be different for all users which upgrade their firmware ... with potetially not so good effects 
we would have to add some dedicated conversion code to catch this particular situation
so, unless there is really strong argument I woud not consider it worthwhile

mLRS/Common/hal/hal-power-pa.h
```diff
@@ -109,6 +109,7 @@ const rfpower_t rfpower_list[] = {
 
 const rfpower_t rfpower_list[] = {
     { .dbm = POWER_0_DBM, .mW = 1 },
+    { .dbm = POWER_3_DBM, .mW = 2 },
```
olliw42, line 112: see comment in main thread

mLRS/Common/sx-drivers/sx126x_driver.h
```diff
@@ -250,7 +250,11 @@ class Sx126xDriverCommon : public Sx126xDriverBase
         SetRxGain(SX126X_RX_GAIN_BOOSTED_GAIN);
         SetOverCurrentProtection(SX126X_OCP_CONFIGURATION_140_MA); // default for SX1262 according to data sheet, but can't hurt
 
+        #if defined POWER_PA_SE2435L || defined POWER_PA_MATEK_MR900_30
```
olliw42, line 253: I guess I would rather like a generic define such as "POWER_USE_PA_CONFIG_10_DBM"
to not spill the generic drivers with device specifics

mLRS/Common/hal/hal-power-pa.h
jlpoltrack: Removed

mLRS/Common/sx-drivers/sx126x_driver.h
jlpoltrack: Fair enough - new define + moved to hal-power

mLRS/Common/hal/hal-power-pa.h
olliw42: thx

mLRS/Common/sx-drivers/sx126x_driver.h
```diff
@@ -250,7 +250,11 @@ class Sx126xDriverCommon : public Sx126xDriverBase
         SetRxGain(SX126X_RX_GAIN_BOOSTED_GAIN);
         SetOverCurrentProtection(SX126X_OCP_CONFIGURATION_140_MA); // default for SX1262 according to data sheet, but can't hurt
 
+        #if defined POWER_USE_PA_CONFIG_10_DBM
```
olliw42, line 253: nitpick
pl remove the blanks(ident of the präprozessor lines
pl use #ifdef
see example https://github.com/olliw42/mLRS/blob/1e5581e04cdf120512c0264d7ad2ec733b5c1d34/mLRS/Common/sx-drivers/sx126x_driver.h#L443-L447

jlpoltrack: Done

olliw42:
thx so much 👍 
will merge tomorrow morning

PR #227, opened 2024-11-08
python3 fix

jlpoltrack:
Was needed when using an install of Python which used 'python3' for the binary.

### DIFF
--- a/run_setup.py
+++ b/run_setup.py
@@ -53,7 +53,7 @@ def check_python():
     global python_cmd
     if _check_python_version('python') == 3:
         python_cmd = 'python'
-    elif check_python('python3') == 3:
+    elif _check_python_version('python3') == 3:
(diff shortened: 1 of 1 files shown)


olliw42:
ähm ... this is interesting ... how did it come that the check for python used a different function than the check for python3?

is there maybe some win/linux thing going on too?

jlpoltrack:
Not sure if OS specific but the error made sense, this function call https://github.com/olliw42/mLRS/blob/main/run_setup.py#L56 was trying to call the 'parent' function https://github.com/olliw42/mLRS/blob/main/run_setup.py#L51 and there's an argument mismatch (it was trying to pass 1 argument and 'check_python' takes none)

olliw42:
ah
so this is clear bug
THX

PR #228, opened 2024-11-11
Fix issue with run_setup and drone can

tmcadam:
This is very rough, but demonstrates a fix for the issue with dronecan failing in run_setup.py

  - Creates a python virtual environment in the mLRS/common/dronecan folder.
  - Installs the dependencies
  - Runs the dronecan steps with the python executable from virtual env

ONLY TESTED ON LINUX!

### DIFF
--- a/mLRS/Common/dronecan/dronecan_generate_c_library.py
+++ b/mLRS/Common/dronecan/dronecan_generate_c_library.py
@@ -13,7 +13,22 @@ import os
 import shutil
 import re
 import sys
+import venv
+import pathlib
+import subprocess
 
+script_folder = pathlib.Path(__file__).parent.resolve()
+venv_folder = os.path.join(script_folder, 'venv')
+venv.create(venv_folder, with_pip=True)
+python_path = os.path.join(venv_folder, "bin", "python")
+
+def pip_install(package):
+    res = subprocess.call([python_path, "-m", "pip",  "install",  package])
+
+pip_install("setuptools")
+pip_install("empy==3.3.4")
+pip_install("pexpect")
+pip_install("dronecan")
 
 mLRSProjectdirectory = os.path.dirname(os.path.abspath(__file__))
 mLRSdirectory = os.path.join(mLRSProjectdirectory,'mLRS')
@@ -33,7 +48,7 @@ def kill_outdir():
 
 
 def os_system(arg):
-    res = os.system(arg)
+    res = os.system(f"{python_path} {arg}")
     if res != 0:
         print('# ERROR (errno =',res,') DONE #')
         os.system("pause")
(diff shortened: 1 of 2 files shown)


jlpoltrack:
Brilliant - confirming it worked for me on MacOS.

olliw42:
oh my gosh ... that's really above my head ... and since I guess I have to maintain that scrupt "scares" me ...
we really need that?
I mean, it's super elegangt and professional, but ... over the top of my head ...

tmcadam:
Hmmm, this is actually an unusual way of doing it, but it keeps it all in a python script, and keeps it cross-platform. I guess other option is to replace to run_setup.py with run_setup.bat(windows) and run_setup.sh(linux mac). These shell scripts can install the dronecan dependencies and run the other python scripts, that would be the more common way of doing it.

Last option is to just document the installation steps for the 4 python dependencies before running run_setup.py and let devs do it themselves. I think you still need to setup python virtual environment for this, the more recent versions of python seem to increasingly severe warnings on system-wide dependency installation.

olliw42:
@tmcadam 
I'm looking at this again as the inderlying issue really needs a solution and your approach looks quite nice (and I start to feel more comfortable as venv's become more familiar to me :))

One questin though I find confusing:
When I run a test code with improt venv venv.create(xxx) I do get a 'Scripts' folder but not a 'bin' folder?
Why is that, why do we get different folders hosting the python.exe?
Is that an os thing, or what?

tmcadam:
@olliw42 

https://stackoverflow.com/questions/43826134/why-is-the-bin-directory-named-differently-scripts-on-windows

It looks like a windows mac/linux thing. I can't remember but likely I was using linux when I created this PR.

mLRS/Common/dronecan/dronecan_generate_c_library.py
```diff
@@ -13,7 +13,22 @@
...
 import re
 import sys
+import venv
+import pathlib
+import subprocess
 
+script_folder = pathlib.Path(__file__).parent.resolve()
+venv_folder = os.path.join(script_folder, 'venv')
+venv.create(venv_folder, with_pip=True)
+python_path = os.path.join(venv_folder, "bin", "python")
```
Hyperion218, line 23: To resolve the issue of differents folder name containing python between Unix/MAC and Windows OS, same-ish code as in the original virtualenv commit ([here](https://github.com/pypa/virtualenv/commit/993ba1316a83b760370f5a3872b3f5ef4dd904c1)) can be used.
```suggestion
bin_dir = "bin"
if sys.platform == 'win32':
    bin_dir = "Scripts"
python_path = os.path.join(venv_folder, bin_dir, "python")
```

moosss:
I just did a fresh `run_setup.py` on Ubuntu in WSL2. I already created 1 PR to fix a python/python3 issue, I also ran into 3 other packages which needed installed manually:
 * empy
 * dronecan
 * pexpect

I ended up installing them with `pip3 install dronecan --break-system-packages`.

I'm not sure if this fix would fix those 3 package issues or not, just wanted to bring it up. I can investigate if needed.

PR #229, opened 2024-11-14
Add multiple page support to bw lua to support all parameters

### LINKED ISSUE #89: Lua for BW screens: Support more than 5 parameters, and other improvements
follow up on https://github.com/olliw42/mLRS/issues/74.

- The current lua script for bw screens supports only 5 parameters, for reasons of limited ram on radios with bw screen. It is however possible to read parameters also one by one, by using the parameter-by-index command, which should allow extended functionality on bw screen radios. 
- The handled parameters are hardcoded by their index. If the setup layout changes the index changes, breaking the functioning.

brad112358:
Some versions of this have been tested on X9D and X9D plus by other users.
Also added README.md with workaround in case of "not enough memory" error.
resolves #89


lua/README.md
```diff
@@ -0,0 +1,11 @@
+# mLRS - lua #
```
olliw42, line 1: the inofficial docs style guide is to write things with their correct way, here it should be Lua, nothing else :)
dito few times below
:)

lua/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This folder contains lua scripts which may be used to change the Tx module and connected receiver settings when the Tx module is installed in a OpenTx or EdgeTx radio.
```
olliw42, line 3: ... in an EdgeTx or OpenTx radio ...

lua/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+[mLRS-bw.lua](mLRS-bw.lua) is for radios with black and white screens
```
olliw42, line 5: I woud suggest to change the order of sequence

lua/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+[mLRS.lua](mLRS.lua) is for radios with color screens
+
+Please see the [documentation page](https://github.com/olliw42/mLRS-docu/blob/master/docs/LUA.md#mlrs-documentation-mlrs-lua-script) for install and usage details
+
+Note:  If you receive a "not enough memory" error, you can try reducing the file size by running the code through a lua minifier tool such as the one [here](https://mothereff.in/lua-minifier) and install the minified version as a workaround.  Please also report this to the mLRS developers.
```
olliw42, line 11: pl make the Note bold and italics

olliw42:
F A N T A S T I C

just tiny style nit picks

could you pl keep the -test lua script? I do not yet want to lose it as source of inspiration (at some point I also want to convert the color lua to a index based download)

brad112358:
> could you pl keep the -test lua script? I do not yet want to lose it as source of inspiration (at some point I also want to convert the color lua to a index based download)

If you insist, but I think the bw script is now a better example of index based download.  The test script may have been useful at one point to exercise the message, but it downloads all of the parameters in a loop, not a subset or a page at a time like the bw script does and it has also not been maintained and it doesn't work with modern versions of  mLRS.  I think it's better to hide it to avoid user or new developer confusion.  Nothing is really removed in git.  You can easily retrieve it from an old branch or tag if you want to reference it or resurrect it later.

lua/README.md
brad112358: The entire note text?

lua/README.md
brad112358: You mean the order the files appear in?  I matched the order on github, but OK.  I guess the color screens are more important?

olliw42:
pl just keep it for now

lua/README.md
olliw42: yes :)

lua/README.md
olliw42: yo, just the note

olliw42:
THX !!!!

PR #230, opened 2024-11-15
Fix crash when editing bind phrase

brad112358:
When shortening variable names to save memory, I missed a couple which affected only editing the bind phrase


olliw42:
THX !!

PR #234, opened 2024-11-28
DIY F1 E28 DC-DC Updates

jlpoltrack:
Updates for the F1 E28 Dual board:

- Increase SPI speed to 18 MHz, helps with jitter
- Enable DC-DC converter on E28, saves about 20 mA (2 x 10 mA)

Have been running this configuration myself for long time.

### DIFF
--- a/mLRS/Common/hal/stm32/rx-hal-diy-e28dual-board02-f103cb.h
+++ b/mLRS/Common/hal/stm32/rx-hal-diy-e28dual-board02-f103cb.h
@@ -75,7 +75,7 @@
 #define SPI_USE_SPI1 // PA5, PA6, PA7
 #define SPI_CS_IO                 IO_PB0
 #define SPI_USE_CLK_LOW_1EDGE     // datasheet says CPHA = 0  CPOL = 0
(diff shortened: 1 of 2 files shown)


olliw42:
I'm ok with the DC-DC
I'm not so sure with the 18 MHz ... that's a lot, and I believe to remember we had a "spi freq" discussion in the past

I guess I would rather like to understand why it would improve jitter ... in theory it should be pretty constant time ... and actually should not show up in your jitter measurements via the tx/rx enable pins. I guess it's not the real reasons.

jlpoltrack:
Fair enough, now just enables the DC-DC and keeps 9 MHz SPI.

olliw42:
just looked into the süecs, and 18 MHz is at least the allowed range
can you recall a past "spi freq" discussion?

jlpoltrack:
I vaguely recall some discussion when working on the ESP32 SPI stuff - you were doing some measurements on F1 and I think I pointed out that F1 + SX128x could run at 18 MHz but the decision at that time was to not update as it worked fine for long time.

PR #235, opened 2024-11-30
Dev txjitter discussion

brad112358:
I have also included the code I'm currently using to gather execution time statistics and a couple of other trivial things in my local copy of your dev-txjitter branch.  Feel free to comment on these additions and perhaps I'll do actual PRs to add these later.


olliw42:
is this based of the latest txjitter? 
seems not so, could you please do? it's a bit hard to see with all these legacy changes

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -762,7 +778,28 @@ INITCONTROLLER_END
...
+//            crsf.TelemetryStart();
+//            whileTransmit.Trigger();
+        }
+
+        link_task_tick_ms();
+
+        if (!doPreTransmit) doSysTask2++;
+    }
+    if (doSysTask2) {
+        doSysTask2--;
```
brad112358, line 800: These 4 lines are a bit confusing to me as they appear to be an obscure way to say just "if (!doPreTransmit) {"
Since doSysTask2 is not referenced anywhere else that I can see.

olliw42: that's true, I found it slimmer though, didn't like the identing

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -762,7 +778,28 @@ INITCONTROLLER_END
...
+
+        if (tx_tick == 2) {
+            isInTimeGuard = true; // prevent extra work
+        }
+        if (tx_tick == 1) {
+            doPreTransmit = true; // trigger next cycle
+            pretransmit_tstamp_us = micros16();
+        }
+        if (tx_tick == 0) {
+//            crsf.TelemetryStart();
```
brad112358, line 791: Why are we no longer doing anything here?  Shouldn't we still be syncing the actual transmit with doSysTask?

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -811,16 +839,26 @@ INITCONTROLLER_END
...
 
     case LINK_STATE_TRANSMIT:
+        do_transmit_prepare(tdiversity.Antenna());
+        link_state = LINK_STATE_TRANSMIT_SEND;
+        DBG_MAIN_SLIM(dbg.puts("\nt");)
+        break;
+
+    case LINK_STATE_TRANSMIT_SEND: {
+        uint16_t dt = micros16() - pretransmit_tstamp_us;
+        if (dt < 500) break;
```
brad112358, line 849: It appears you want to send at least 500 us after tx_tick == 1.  How is this supposed to reduce jitter?  Why aren't we syncing to doSysTask here?

olliw42: we are not synching to tx anymore becasue when we did the gap had been 1 ms ... now it can be shorter, here 500 us

mLRS/CommonTx/mlrs-tx.cpp
olliw42: no, we sync on the doPretransmit tick, and do the transmit 500 us later

the tick==0 might be removed eventually and the other ticks deccrease by one. I was testing with when to better start the crsf and while thing, moved it now to below into the case TRANSMIT. Not sure if that is needed however, if not I would prefer it to be here.

brad112358:
> is this based of the latest txjitter? seems not so, could you please do? it's a bit hard to see with all these legacy changes

This is the changes between your main and your dev-txjitter

mLRS/CommonTx/mlrs-tx.cpp
brad112358: I'd much prefer the indenting!

mLRS/CommonTx/mlrs-tx.cpp
brad112358: OK.  So, should I test 500 or 750 which JLP says was better?

olliw42: I guess I would first go with 500
it is not yet clear to me why 500 was a regression for JLP
if you also see 500 us isn't good but 750 is, I guess it's maybe not worth the 250us and should undo the last commit

PR #236, opened 2024-12-01
rx, tx dynamic fan

jlpoltrack:
When using switchable power + fan on/off, fan is only configured during boot - this updates the fan to use the current power level in the main loop.

When using 500 mW + switchable power:

Before - fan on all the time
After - fan on only at 500 mW

### DIFF
--- a/mLRS/Common/fan.h
+++ b/mLRS/Common/fan.h
@@ -35,7 +35,7 @@ class tFan
     void Init(void);
 
     void SetPower(int8_t power_dbm);
-    void Tick_ms(void);
+    void Tick_ms(int8_t power_dbm);
 
   private:
     bool initialized;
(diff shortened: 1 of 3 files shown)


olliw42:
nice catch !
wouldn't it be easier and work too to just add a `fan.SetPower(sx.RfPower_dbm());` before `fan.Tick_ms();` ?
would also do anything only if the power level has changed

jlpoltrack:
> wouldn't it be easier and work too to just add a `fan.SetPower(sx.RfPower_dbm());` before `fan.Tick_ms();` ?

Done

olliw42:
and it works?

jlpoltrack:
> and it works?

Yea, tested on Matek 2G4 with 500 mW

PR #237, opened 2024-12-03
Fix race when tasks take longer than doSysTask tick

brad112358:
Since the "--" operator is not generally atomic in RISC architectures and, I believe, specifically not on STM32 or ESP32 where the decrement is implemented as load, modify, store, the change from clearing doSysTask to decrementing it introduced a race.

I am not aware that this race has ever caused an observed problem as it should be fairly rare, but, consider what would happen.
In the case where the ISR executes between the load and store; the increment in HAL_IncTick() is effectively lost which can make the Tx transmit 1 ms late which could cause the problem on the Rx which we are now all too familiar with.

I believe the method implemented here is the most efficient way to avoid this race.

### DIFF
--- a/mLRS/Common/hal/esp-timer.h
+++ b/mLRS/Common/hal/esp-timer.h
@@ -17,17 +17,28 @@
 #define SYSTICK_DELAY_MS(x)       (uint16_t)(((uint32_t)(x)*(uint32_t)1000)/SYSTICK_TIMESTEP)
 
 
-// Two variables are required to avoid race with HAL_IncTick() when tasks may take longer than one tick
-volatile uint32_t doSysTask = 0; // Only changed in ISR; incremented to signal tick
-uint32_t doneSysTask = 0; // Never changed in ISR; incremented when (doSysTask != doneSysTask)
+// Two variables allow to avoid race with HAL_IncTick() when tasks may take longer than one tick
+uint32_t doSysTask_done = 0; // Never changed in ISR; incremented when (uwTick != doSysTask_done)
+volatile uint32_t uwTick = 0; // Only changed in ISR; incremented to signal tick
 
-volatile uint32_t uwTick = 0;
+void resetSysTask()
+{
+    doSysTask_done = uwTick;
+}
 
+volatile bool doSysTask()
+{
+    if (uwTick != doSysTask_done)
+        {
+            doSysTask_done++;
+            return true;
+        }
+    return false;
+}
 
 IRAM_ATTR void HAL_IncTick(void)
 {
-    uwTick += 1;
-    doSysTask++;
+    uwTick++;
 }
 
 
@@ -111,11 +122,10 @@ uint16_t micros16(void)
 
 void timer_init(void)
 {
-    doSysTask = 0;
-    doneSysTask = 0;
     uwTick = 0;
     systick_millis_init();
     micros_init();
+    resetSysTask();
 }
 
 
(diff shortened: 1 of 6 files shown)


olliw42:
interesting.

the canonical approach would be to disable/enable the isrs. In stdstm we do have the ATOMIC() macro to do this. 
so ATOMIC(doSysTask--);

I bleieve this is only needed in the main loop's doSysTask. Note that in esp.h and hc04.h doSysTask is set to 0, which should be atomic.

brad112358:
Thanks for helping make my point.  The two flavors of ESP can't even provide the same macro to accomplish what we need here.
I think I greatly prefer the portable code I've provided here.

Are you interested in combining uwTick and doSysTask into one variable?

olliw42:
personally I would not mind having some ATOMIC for esp8285 and esp32 also (I don't see why it's a point of criticism that the two flavors don't have the same ... they are two different mcus)
I do see the point though. 
However, I think this then really needs to be abstracted, like so it's ugly.
I guess one could introduce functions resetDoSysTask(), doDoSysTask(), updateDoSysTask(), the two variables could be called doSysTask_pending, doSysTask_done. 
If all hidden behind an API, doSysTask_pending could go and uwTick be used.

mLRS/Common/hal/esp-timer.h
```diff
@@ -17,17 +17,28 @@
...
-volatile uint32_t uwTick = 0;
+void resetSysTask()
+{
+    doSysTask_done = uwTick;
+}
 
+volatile bool doSysTask()
+{
+    if (uwTick != doSysTask_done)
+        {
```
olliw42, line 32: formatting pl, 
if () {
}
and 4 spaces ident
dito below
:)

brad112358: I think I followed the existing formatting in the file which I strongly thinks is what we want to do.  I think a separate PR should be used if you want to reformat existing code or it gets hard to see what changed later.

Can you point out the specific area where I didn't use 4 spaces?

olliw42: I'm not sure where you have seen that formatting of if clauses in the code
if there are exaples somewhere Ive missedthem, but that's not the normal
pl change it

brad112358: Sorry.  I got hung up on the 4 spaces thing and missed that you were talking about just the if statement

brad112358: I didn't look at what you typed carefully enough and thought you meant the function declaration.  BTW,  Why aren't you consistent between if and functions?

brad112358: I think the dito below part also contributed to my misunderstanding.  Where below?

mLRS/Common/hal/timer.h
```diff
@@ -17,14 +17,27 @@
...
-volatile uint32_t doSysTask = 0;
+void resetSysTask()
+{
+    doSysTask_done = uwTick;
+}
 
+volatile bool doSysTask()
+{
+    if (uwTick != doSysTask_done)
+        {
```
olliw42, line 31: here?
it seems I'm saying something very difficult :)

brad112358: Ahh.  OK.  I thought you meant below in the same file.

brad112358:
Hopefully this is better...

olliw42:
👍 
LGTM
when you have tested it and found to work I would merge

PR #240, opened 2024-12-05
R9 binary output

jlpoltrack:
Enables binary output (.bin) for all R9 hardware that has ELRS bootloader support - makes it easier to generate .ELRS files without using the tools script.


olliw42:
I'm not sure I get this right: what this PR does is to simply enable the checkbox to also create a bin file when compiling?

jlpoltrack:
> I'm not sure I get this right: what this PR does is to simply enable the checkbox to also create a bin file when compiling?

Correct, this option is stored in the .cproject files (so avoids me having to do)

olliw42:
THX

PR #242, opened 2024-12-10
Add warning to bw lua when Rx and Tx setup layout versions don't match

brad112358:
I also found some code which could be removed, so hopefully, this won't expand memory use much.

This has been tested only on my RM pocket


olliw42:
super. THX.

PR #243, opened 2024-12-11
Bandit Rx 2

jlpoltrack:
Trying this again now that the latest release is out :)


olliw42:
may I somewhat critically ask, is there a real use case, i.e., is there somoeone who actually wants to use it, or is it more like "can be done"?

jlpoltrack:
Good question, while a bit of a stretch - IMO it offers a much better experience than R9M as receiver - as it has no inverters to deal with and offers JST-GH connectors on the bottom of the module:

image

olliw42:
ah, that's a good point
probably valid only for the big bandit though
the micro is just this usual jrbay thing and needs real tinkering for making it useable, I think
but it's not like you or someone would have specifically asked for it, right?

jlpoltrack:
> probably valid only for the big bandit though

FWIW, Bandit Micro is also touted as being a receiver on the RM product page: 
image

> but it's not like you or someone would have specifically asked for it, right?

Mo did ask about using Bandit for both Tx and Rx but agree, probably not a lot of users.

PR #244, opened 2024-12-12
Add requirements.txt to track required python modules

ecarjat:
Creates the file requirements.txt to store the required python modules necessary to run `run_setup.py`.
This file needs to be updated every time a new module is necessary to run this script. If one follows the instructions in the associated mLRS-docu repo (https://github.com/olliw42/mLRS-docu/pull/238), the file can always be created with the following command `pip freeze > requirements.txt`


olliw42:
I'm closing this.

adding a requirements.tx would indeed be a useful thing, but AFAICS it involves substantial manual labor. Running pip freeze > requirements.txt as suggested does not look as a viable solution to me as it gives a zillion of dependencies and I'm really not in a possition to sort out which few should reasonably go into the file (nor would I be willing to do so).

Many thx for suggesting it however. Much appreciated.

PR #246, opened 2024-12-14
Add simple time statistics functions

brad112358:
for measuring min/max/ave time spent in one or more code paths. This may be useful for debugging and optimizing time critical operations.

### DIFF
--- a/mLRS/Common/common.h
+++ b/mLRS/Common/common.h
@@ -24,7 +24,6 @@
 #include "buzzer.h"
 #include "fan.h"
 #include "leds.h"
-//#include "time_stats.h" // Uncomment for time measurement functions
 
 
 //-------------------------------------------------------
--- a/mLRS/Common/common.h
+++ b/mLRS/Common/common.h
@@ -24,6 +24,7 @@
 #include "buzzer.h"
 #include "fan.h"
 #include "leds.h"
+//#include "time_stats.h" // Uncomment for time measurement functions
 
 
 //-------------------------------------------------------
--- a/mLRS/Common/time_stats.h
+++ b/mLRS/Common/time_stats.h
@@ -1,0 +1,83 @@
+//*******************************************************
+// Copyright (c) MLRS project
+// GPL3
+// https://www.gnu.org/licenses/gpl-3.0.de.html
+// OlliW @ www.olliw.eu
+//*******************************************************
+// time_stats
+//*******************************************************
+#ifndef TIME_STATS_H
+#define TIME_STATS_H
+#pragma once
+
+// Time Statistics for up to PM_NUM_BLOCK pairs of measurement points
+#ifdef USE_DEBUG
+#define TS_NUM_BLOCK 10
+
+struct TS_store {
+    uint32_t start;
+    uint32_t last_report;
+    uint32_t total;
+    uint32_t min;
+    uint32_t max;
+    uint32_t count;
+    uint32_t overflow;
+    uint16_t last_cnt;
+};
+static struct TS_store TS_data[TS_NUM_BLOCK];
+
+inline uint32_t TS_MICROS32(int block) {
+
+    // Separate overflow and last_cnt for each block allows either ISR/normal use, but not both
+    uint16_t cnt = MICROS_TIMx->CNT;
+    if (cnt < TS_data[block].last_cnt) {
+        TS_data[block].overflow += 0x10000;
+    }
+    TS_data[block].last_cnt = cnt;
+    return TS_data[block].overflow + cnt;
+}
+
+inline void TS_START(int block) {
+    TS_data[block].start = TS_MICROS32(block);
+}
+
+inline void TS_END(int block, uint32_t report_period = 10000000, bool cont = false) {
+    if (!cont && TS_data[block].start == 0) return; // no end without start
+    uint32_t time = TS_MICROS32(block);
+    uint32_t elapsed = time - TS_data[block].start;
+    if (TS_data[block].count == 0) {
+        TS_data[block].total = TS_data[block].max = TS_data[block].min = elapsed;
+    } else {
+        TS_data[block].total += elapsed;
+        if (elapsed < TS_data[block].min) TS_data[block].min = elapsed;
+        if (elapsed > TS_data[block].max) TS_data[block].max = elapsed;
+    }    
+    TS_data[block].count ++;
+    if (time - TS_data[block].last_report >= report_period) {
+        dbg.puts("[");
+        dbg.puts(u8toBCD_s(block)); dbg.puts("] ");
+        dbg.puts(u32toBCD_s(TS_data[block].total));
+        dbg.puts("/");
+        dbg.puts(u32toBCD_s(TS_data[block].count));
+        dbg.puts(" ");
+        dbg.puts(u16toBCD_s(TS_data[block].min));
+        dbg.puts(" ");
+        dbg.puts(u16toBCD_s(TS_data[block].max));
+        dbg.puts("\r\n");
+        TS_data[block].count = 0;
+        TS_data[block].last_report = time;
+    }
+    if (cont)
+        TS_data[block].start = TS_MICROS32(block); // continuous accumulation
+    else
+        TS_data[block].start = 0; // no end without start
+}
+#else
+inline uint32_t TS_MICROS32(int block) {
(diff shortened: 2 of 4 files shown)


mLRS/Common/time_stats.h
```diff
@@ -0,0 +1,83 @@
...
+        TS_data[block].count = 0;
+        TS_data[block].last_report = time;
+    }
+    if (cont)
+        TS_data[block].start = TS_MICROS32(block); // continuous accumulation
+    else
+        TS_data[block].start = 0; // no end without start
+}
+#else
+inline uint32_t TS_MICROS32(int block) {
```
olliw42, line 76: I still don't understand the benefit over e.g. #define TS_MICROS32()
I understand the compiler will hopefully optimize away, but it looks more natural to me

mLRS/Common/common.h
```diff
@@ -24,6 +24,7 @@
 #include "buzzer.h"
 #include "fan.h"
 #include "leds.h"
+//#include "time_stats.h" // Uncomment for time measurement functions
```
olliw42, line 27: couldn't we just always have it included?
does it need to be included here, or could it also be included in main?

brad112358: This just seemed like a convenient place to include it.  I left it commented out because it might be used only by developers and I didn't want it taking up space if it's not used when USE_DEBUG is defined.  I'd suggest you decide where to include it.

mLRS/Common/time_stats.h
```diff
@@ -0,0 +1,83 @@
...
+    uint32_t total;
+    uint32_t min;
+    uint32_t max;
+    uint32_t count;
+    uint32_t overflow;
+    uint16_t last_cnt;
+};
+static struct TS_store TS_data[TS_NUM_BLOCK];
+
+inline uint32_t TS_MICROS32(int block) {
```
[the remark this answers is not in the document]
brad112358: Yes, good point.  It should be static inline.

mLRS/Common/time_stats.h
```diff
@@ -0,0 +1,83 @@
...
+
+    // Separate overflow and last_cnt for each block allows either ISR/normal use, but not both
+    uint16_t cnt = MICROS_TIMx->CNT;
```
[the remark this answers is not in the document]
brad112358: I wanted to avoid the extra function call overhead.  But you have reminded me I meant to make this work for ESP as well.  Can we mark micros16() as inline in both esp-timer.h and timer.h?

mLRS/Common/time_stats.h
```diff
@@ -0,0 +1,83 @@
...
+    uint32_t max;
+    uint32_t count;
+    uint32_t overflow;
+    uint16_t last_cnt;
+};
+static struct TS_store TS_data[TS_NUM_BLOCK];
+
+inline uint32_t TS_MICROS32(int block) {
+
+    // Separate overflow and last_cnt for each block allows either ISR/normal use, but not both
```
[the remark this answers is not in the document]
brad112358: You didn't comment on this in a timer.h, but, in fact, while micros16() is safe to call from an ISR, micros64() is not safe to call from an ISR.  If we want the ability to use TS functions in both user level and ISR level code, we need to avoid race conditions. If we have a single copy of overflow_cnt and last_cnt like you do in micros64(), it won't always count properly when called from both normal and ISR level because an interrupt can happen in the middle of the call to micros64() and a second call to micros64() can happen which might result in an incorrect overflow_cnt or last_cnt.  It's faster to have a separate set of variables for each code block than to disable interrupts in the critical section.  If you are wondering if the race actually shows up in the statistics, I can testify that it did when I tried to use my first version of this code to gather data on interrupt timing.  I was seeing output every 3 seconds instead of every 10 seconds.

mLRS/Common/time_stats.h
brad112358: There are lots of benefits of functions over macros.  I won't argue which ones apply here, I just prefer this approach in this case.

mLRS/Common/time_stats.h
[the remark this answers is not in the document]
brad112358: Note that with my approach, it is still not safe to split TS_START() and TS_END() between normal and ISR code for the same block.  Thus, we can't use these functions to measure the time between setting a device register at user level and the interrupt which results from this, for example.  Any given block must be referenced only in a single ISR or only in user level code.

mLRS/Common/common.h
brad112358: If we always include it, I can add an extra ifdef so it can be disabled independently from USE_DEBUG

mLRS/Common/time_stats.h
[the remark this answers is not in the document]
olliw42: thx for the explanation. 
I guess it would be useful to have a bit of comments

mLRS/Common/time_stats.h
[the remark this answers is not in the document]
olliw42: with inline or not the compiler will do anyway what it wants. I'm going more an more away from it.

mLRS/Common/time_stats.h
olliw42: for my education, would you maybe name one or two?

mLRS/Common/common.h
olliw42: ok.

olliw42: I guess I would suggest to completely remove the #else with content. That way code needs to be cleaned if not in debug.

mLRS/Common/time_stats.h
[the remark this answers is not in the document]
brad112358: I guess if we want to force the compiler to inline, we can also add __attribute__((always_inline)).  Is there any reason not to do this for micros16() in esp-timer.h and timer.h?

mLRS/Common/time_stats.h
brad112358: Google should find plenty of educational info, though I doubt you will see much you don't already know or couldn't figure out for yourself.  Side effects from evaluating the arguments multiple times in the #define case are one example.  Google found these for me:
https://icarus.cs.weber.edu/~dab/cs1410/textbook/6.Functions/inline.html
https://chunminchang.gitbooks.io/cplusplus-learning-note/content/Appendix/preprocessor_macros_vs_inline_functions.html

mLRS/Common/common.h
brad112358: OK, but I'll remove the empty functions with "#if 0" in case we decide we want to be able to leave some use of these in production.

mLRS/Common/time_stats.h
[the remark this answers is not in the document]
olliw42: my point rather was I don't want to force to inline. I tended to rather go away form that since the actual effect is usually like zero.

mLRS/Common/common.h
brad112358: OK.  It's easy enough to put back later if we decide it's handy to measure something whenever debug is on.

olliw42:
do you see ANY real-world benefit of the forced inline?
the code looks overdone frankly

TS_END really should be forced inline?

why just one static inline an dnot the others too?

it should not be micros() for ESP, but it should use the same as for stm32

I will merge and then tidy :)

brad112358:
I admit I didn't even try to measure any benefit that inline provides so it is only my best guess that it should help reduce the overhead of using these functions.  I think this guess will be correct in environments without a significant instruction cache such as our STM mcus.  But, perhaps I should remove the always_inline and just let the compiler decide.  I did read somewhere that GCC won't inline without being forced when optimization is turned off though.

Regarding the use of static here:  For a time, I was mistakenly thinking of this code like it was part of a .cpp file instead of the header file where it actually lives which is why I changed TS_MICROS32() to static at your prompting since it is not intended to be called from any other file.
But I think, in the general case of header files, I got this wrong.
I think TS_START() and TS_END() probably should not be marked static especially without always_inline since if the compiler chooses not to inline these functions, we would end up with a separate copy for each compilation unit where the file is included.  Now that I say this, I think I probably have static in a couple of other places where things should probably have external linkage as well.

In practice though, I guess it doesn't matter since we only include this file in one compilation unit.  But, if you are going to tidy this up, I suggest removing static rather than adding it.

PR #250, opened 2025-01-11
Don't use broadcast to avoid lost messages

brad112358:
esp8266 assigns ip+99 instead of ip+1 for first DHCP client

### DIFF
--- a/esp/mlrs-wireless-bridge-esp8266/mlrs-wireless-bridge-esp8266.ino
+++ b/esp/mlrs-wireless-bridge-esp8266/mlrs-wireless-bridge-esp8266.ino
@@ -131,7 +131,7 @@ void led_off(void) {}
 
 #if (WIRELESS_PROTOCOL <= 1) // WiFi TCP, UDP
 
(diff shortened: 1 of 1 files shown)


brad112358:
This works much better for me!

jlpoltrack:
Looks much better but still see small single digit packet loss per Mission Planner (46/1731 = 2.6%) ?  Do you see this?  

image

jlpoltrack:
Updated - I tried again this morning with WiFi Channel = 13 and WiFi Power = 10 and know it looks perfect.  Will attribute the issues above to environment issues.

brad112358:
I've let it run a couple of hours at 1mw Tx power and default WiFi Power (What is default?).  QGC reported about 0.1% message lost in that test.

olliw42:
could you pl change the comment to

// the first DHCP client/MissionPlanner gets assigned +99

seems like we want to merge :)

brad112358:
By MissionPlanner, you actually mean any ground station, but OK.

olliw42:
it's for dummies like me to knoow what we are talking about :)

MANY THX Brad!!!

PR #251, opened 2025-01-12
Add support for many Radiomaster internal 2400 ELRS modules

brad112358:
No provision yet for backpack configuration or flashing without using ELRS tools.

See the comments in Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h for some hints on flashing.


lua/mLRS.lua
```diff
@@ -1374,7 +1374,7 @@ local function scriptRun(event)
         return 2
     end
     if mbridge == nil or not mbridge.enabled() then
-        if model.getModule(1).Type ~= 5 then
+        if model.getModule(0).Type ~= 5 and model.getModule(0).Type ~= 5 then
```
olliw42, line 1377: this looks like a bug, right? one should be (1), shouldn'tt it

mLRS/Common/hal/esp/esp-device_conf.h
```diff
@@ -149,6 +149,13 @@
   #define FREQUENCY_BAND_2P4_GHZ
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_TX16S_2400_ESP32
```
olliw42, line 152: I guess I would prefer having an indication that it's an internal in the def name

mLRS/Common/hal/esp/esp-device_conf.h
```diff
@@ -149,6 +149,13 @@
...
+  #define DEVICE_NAME "RM TX16S int 2.4G"
```
olliw42, line 153: maybe "... Int ..." or "... INT ..."

general question: are all these internal different, and need extra handling by us? Or are there some generics possible?

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
```diff
@@ -0,0 +1,214 @@
...
+#define UARTB_USE_SERIAL // serial, is on P16/P17
+#define UARTB_BAUD                TX_SERIAL_BAUDRATE
+#define UARTB_USE_TX_IO           IO_P17
+#define UARTB_USE_RX_IO           IO_P16
+#define UARTB_TXBUFSIZE           1024 // TX_SERIAL_TXBUFSIZE
+#define UARTB_RXBUFSIZE           TX_SERIAL_RXBUFSIZE
+
+#define UART_USE_SERIAL1 // full duplex CRSF/MBridge (JR pin5)
+//#define UART_BAUD                 400000
+#define UART_BAUD                 1843200 // Internal module
```
jlpoltrack, line 84: This is the 1.87 Mbaud option in EdgeTx?

mLRS/Common/hal/esp/esp-hal.h
```diff
@@ -88,6 +88,10 @@
 #include "tx-hal-radiomaster-rp4td-2400-esp32.h"
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_TX16S_2400_ESP32
+#include "tx-hal-radiomaster-tx16s-2400-esp32.h"
```
olliw42, line 92: I guess I'd like some indication of int also in file name

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
```diff
@@ -0,0 +1,214 @@
...
+    "debug_backpack_baud": 460800,
+    "debug_backpack_rx": 16,
+    "debug_backpack_tx": 17,
+    "backpack_boot": 15,
+    "backpack_en": 25
+*/
+
+
+#define DEVICE_HAS_JRPIN5
+#define JRPIN5_INTERNAL // Full duplex CRSF interface for Internal modules
```
olliw42, line 63: usually we only have DEVICE_HAS_xxx at the top

(and these usually can be changed without the things breaking apart). A specifier like JRPIN5_INTERNAL should go somewhere below.

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
```diff
@@ -0,0 +1,214 @@
...
+
+#define UARTB_USE_SERIAL // serial, is on P16/P17
+#define UARTB_BAUD                TX_SERIAL_BAUDRATE
+#define UARTB_USE_TX_IO           IO_P17
+#define UARTB_USE_RX_IO           IO_P16
+#define UARTB_TXBUFSIZE           1024 // TX_SERIAL_TXBUFSIZE
+#define UARTB_RXBUFSIZE           TX_SERIAL_RXBUFSIZE
+
+#define UART_USE_SERIAL1 // full duplex CRSF/MBridge (JR pin5)
+//#define UART_BAUD                 400000
```
olliw42, line 83: this can then go

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
```diff
@@ -0,0 +1,214 @@
...
+#define UART_BAUD                 1843200 // Internal module
+#define UART_USE_TX_IO           IO_P1
```
olliw42, line 85: something went worng with tab/spaces I guess

jlpoltrack:
Simply lovely, tested on TX16S + SuperD, quick test looks good:

image

mLRS/CommonTx/esp.h
```diff
@@ -71,6 +71,9 @@ bool com_is_full(void)
 {
 #ifdef DEVICE_HAS_COM_ON_USB
     return usb_tx_full();
+#elif defined DEVICE_HAS_NO_COM
+    //return !uart_tx_notfull();
```
olliw42, line 75: is this a comment you've added and wasn't in the original code

I strugge a bit

jlpoltrack:
Should update date in Lua here: https://github.com/olliw42/mLRS/blob/main/lua/mLRS.lua#L13

mLRS/CommonTx/esp.h
```diff
@@ -286,8 +289,10 @@ void tTxEspWifiBridge::EnterFlash(void)
 {
     if (!passthrough) return;
 
+#ifdef USE_DISPLAY
```
olliw42, line 292: good catch
isn't only for this internal PR, but should go into main independently, right?
if so might be nice to have an extra PR, as this PR here may need a bit longer

mLRS/CommonTx/jr_pin5_interface.h
```diff
@@ -105,7 +108,8 @@ class tPin5BridgeBase
 
     // interface to the uart hardware peripheral used for the bridge, called in isr context
     void pin5_tx_start(void) { uart_tx_start(); }
-    void pin5_putc(char c) { uart_tx_putc_totxbuf(c); }
+    // This is clumsy.  Can we avoid directly manipulating txbuf?
+    void pin5_putbuf(uint8_t* buf, uint16_t len) { while (len--) {uart_tx_putc_totxbuf(*(buf++)); }}
```
olliw42, line 112: is this only for your changes, and would break e.g. STM32 targets

if so, pl figure out how to not break STM32. If not, consider doing an extra PR. This PR here may take longer.

Btw: I don't like the len-- buf++ construct, For pointers I generally like to go to a function protoype ... uint8_t* const buf ... (see a number of the recent changes)(helps the compiler to detect certain flaws)(we had at least one where it triggered :)).

mLRS/CommonTx/mbridge_interface.h
```diff
@@ -135,26 +135,24 @@ bool tMBridge::transmit_start(void)
...
-// we can assume that there is at least one byte available
+// send in one chunk to help ensure it is transmitted with no gaps
 uint8_t tMBridge::send_serial(void)
 {
-    pin5_putc(0x00); // we can send anything we want which is not a command, send 0x00 so it is easy to recognize
-    for (uint8_t i = 0; i < MBRIDGE_M2R_SERIAL_PAYLOAD_LEN_MAX; i++) {
-        uint8_t c = serial_getc();
-        pin5_putc(c);
-        if (!serial_rx_available()) break;
+    uint8_t buf[MBRIDGE_M2R_SERIAL_PAYLOAD_LEN_MAX+1];
```
olliw42, line 141: same question/point ass the one before

mLRS/CommonTx/esp-jr_pin5_interface.h
```diff
@@ -0,0 +1,147 @@
...
+// Check and rescue
+// a good place to call it could be ChannelsUpdated()
+// Note: For the FRM303 it was observed that the TC callback may be missed in the uart isr, basically when
+// the jrpin5's uart isr priority is too low. This caused the jrpin5 loop to get stuck in STATE_TRANSMITING,
+// and not even channel data would be received anymore (= very catastrophic). This code avoids this.
+// With proper isr priorities, the issue is mainly gone, but the code remains, as safety net.
+//
+// For full duplex CRSF (internal module), this is used to poll for received bytes and pace telemetry
+
+void tPin5BridgeBase::CheckAndRescue(void)
```
olliw42, line 131: ok, this seems to be the key part

it seems you need to link into the code at this place and the call of this function appeared to be at the right place
not nice as you can guess, given the name of the function.
consider adding an extra function/method for what you need to do, with a proper name.

If CheckAndREscue is empty so it be. But this here doesn't really anything like that,

mLRS/CommonTx/esp-jr_pin5_interface.h
```diff
@@ -0,0 +1,147 @@
...
+{
+    while (uart_rx_available()) { // polling for now
+        parse_nextchar(uart_getc());
+    }
+
+    uint32_t tnow_ms = millis32();
+
+    // Send telemetry and mbridge at the same rate as external module to be sure lua can keep up
```
olliw42, line 139: from scratch this owuld worry me
doesn't it assume that the while loop before really stops at the end of a frame, so this transmit starts comes in sync?

generally I also try to not to do too much in a sub loop, to give the higher loop a chance to run. Am wondering also from this point if the while loop before shouldn't have a test for an end of a framre

olliw42:
FANTASTIC

I'm a bit jaleous I don't have a suitable radio, to be honest. (but I don't need one) 

can't do testing. So, just some quick comments pulled out of air. Mostly tidy and nitpick. It seems to me it is not yet non-breaking other targets. I guess would be nice to work towards that.

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
brad112358: Hmm.  Oops.  I guess I just don't know what the standard rates are in this range.  Before I realized this was configurable from the hardware menu, I assumed it would be one of the same speeds available for external modules, but it didn't work so I measured it with the scope and it looked like it was very close to  4 times 460800, but later when I discovered this was configurable in EdgeTx, I didn't notice the discrepancy. 1.84 is not 1.87, but it's close enough to work.  I will fix this.  I had assumed that all radios would ship with a common default speed, but now I see that the pocket was set to 5.25M even though the TX16S was set to 1.87.  If we can't depend on a common setting, we may as well use 400K and reduce the overhead of the messages coming from the radio.  We are not super latency sensitive for RC anyway.  Is 1.87M really exactly 18700000?  Or is the displayed number rounded?  Another reason to stick with 400K?

lua/mLRS.lua
brad112358: Yup, JLP noticed it didn't work on external modules and I fixed it.  Thanks!

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
jlpoltrack: Here are the rates: https://github.com/EdgeTX/edgetx/blob/d6f24d1ea543f5e3338e1b9a9694584b8717ee43/radio/src/telemetry/crossfire.h#L122-L128

I suppose I would go with 400K if that works - mostly to align with the external modules.

mLRS/Common/hal/esp/esp-device_conf.h
brad112358: You choose this name in your owpr-esp-tx-internal branch and I changed to match what you chose when I looked at your branch in the middle of getting this going.  grumble grumble

mLRS/Common/hal/esp/esp-device_conf.h
brad112358: It seems that they don't even use the same pins between two different RM boards!

mLRS/Common/hal/esp/esp-hal.h
brad112358: You also chose this name.  Does the "I guess" indicate I don't need to change it? :-)

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
brad112358: I haven't used this yet anywhere.  But we will probably need something like this eventually.

mLRS/CommonTx/esp.h
brad112358: I added this and it needs to be cleaned up, and, depending on baud rates, uart_tx_notfull needs a real implementation.

mLRS/CommonTx/esp.h
brad112358: Yes, it is more general than this pr, but doesn't seem worth a separate pr to me.  Since JLP is already developing on top of this code, hopefully I can get this cleaned up and you can merge it fairly soon.

mLRS/CommonTx/jr_pin5_interface.h
brad112358: How does this break STM32?

mLRS/CommonTx/mbridge_interface.h
brad112358: This is part of how I didn't break stm32

mLRS/CommonTx/esp-jr_pin5_interface.h
brad112358: Could we just rename it to something that covers both rescue and poll concepts?  They are a bit similar in function and I'd rather not introduce another.  How about CheckTransmit?

mLRS/CommonTx/esp-jr_pin5_interface.h
brad112358: I believe there is no need to be in sync as this is a full duplex serial connection.  We are doing two independent things here:  Gathering up all the received data and sending if it's time.

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
```diff
@@ -0,0 +1,214 @@
...
+//-- UARTS
+// UARTB = serial port to backpack WiFi bridge
+// UART  = full duplex CRSF serial connection to radio - code still calls it JR bay pin5
+// UARTF = debug port
+
+#define UARTB_USE_SERIAL // serial, is on P16/P17
+#define UARTB_BAUD                TX_SERIAL_BAUDRATE
+#define UARTB_USE_TX_IO           IO_P17
+#define UARTB_USE_RX_IO           IO_P16
+#define UARTB_TXBUFSIZE           1024 // TX_SERIAL_TXBUFSIZE
```
jlpoltrack, line 79: Did you find 1024 needed here?  Could this be switched to the TX_SERIAL_TXBUFSIZE define which is 512?

mLRS/Common/hal/esp/esp-device_conf.h
olliw42: me having choosen something in some initial thing long ago isn't a good argument :D:D:D

I guess TX_ELRS_INTERNAL_xxx or TX_ELRS_INT_xxx would give some easy to catch naming scheme which also would do some good ordering

mLRS/Common/hal/esp/esp-hal.h
olliw42: you don't have to, but I think we want to think about the naming ... for the define due to the ELRS in it I find it easier than for the hal name

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
olliw42: then please at least leave a 

#ifndef JRPIN5_INTERNAL #error #endif in the esp jrpin5 file

mLRS/CommonTx/jr_pin5_interface.h
olliw42: it doesn't ? I was thinking so since it's a change in the main h file.

if not when all good (except the len-- buf++ thing)

mLRS/CommonTx/esp-jr_pin5_interface.h
olliw42: I'm the opposite opinion. Better to disentangle. The two things do totally different things for the two mcus

mLRS/CommonTx/esp-jr_pin5_interface.h
olliw42: would this be even true if in the parse loop two frames would be parsed?

mLRS/CommonTx/esp-jr_pin5_interface.h
brad112358: It really feels stupid to me to have two mutually exclusive entry points called from the same spot which both manage serial communication with the radio.  But I guess there is no point arguing.  So, should both be called and one is empty in each case?  Or should there be another ugly ifdef where they are called so we don't waste time calling an empty function?

mLRS/CommonTx/esp-jr_pin5_interface.h
brad112358: Why wouldn't it be true? I admit I haven't looked at the EdgeTx side to see if they made any stupid assumptions, but why would it matter?  This is full duplex.  We can send our telemetry to the radio whenever we like.  I choose to send it with about 4ms between each message to give the Lua script the same amount of time to process a message as it had in the half duplex case.

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
brad112358: The size needed probably depends mostly on protocol used to flash the ESP8286 and the difference in baud rate on the two ends of the passthrough.  I took this number from the serial port used for the backpack on the bandit.   Do you think it can be smaller?

mLRS/CommonTx/jr_pin5_interface.h
brad112358: Can you point me to the recent changes you reference?  Or, better, elaborate on what the problem is with the construct I've used and how specifying const can matter in this case and how you would prefer to write this function?

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
jlpoltrack: Am currently using 512 and seems fine - although flashing the ESP8266 isn't something that I had really considered, so I think fine to leave as-is.

mLRS/CommonTx/jr_pin5_interface.h
olliw42: the problem of the construct is that you can't write unit8_t* const buf in the function header. That's however the pattern I've went to in the recent times for the reason I've mentioned.

mLRS/CommonTx/esp-jr_pin5_interface.h
olliw42: for a god number of frames there will be a response, but if it parses two messages and call transmit only once a response can be missed (if I recall the code doesn't buffer responses)

mLRS/Common/hal/esp/esp-device_conf.h
brad112358: OK.  There seems to be more target sharing possible than I expected which raises the issue of naming again.  What should this be called if it supports all or most of the Radiomaster 2.4 internal modules?

mLRS/Common/hal/esp/tx-hal-radiomaster-tx16s-2400-esp32.h
brad112358: I think I'll just remove this define for now.

mLRS/CommonTx/esp-jr_pin5_interface.h
brad112358: OK.  I'm not sure if that actually ever happens, but it's a good point and another good reason to stick with 400K baud.
I'll rework this.

brad112358:
Hopefully I accounted for all the comments so far in the version I just pushed.

olliw42:
@brad112358 can you pl do me a favor and remove/undo the changes in the wireless .ino script from this PR

PR #253, opened 2025-01-16
Jumper ELRS Internal 2G4

jlpoltrack:
- Adds support for Jumper Internal 2G4 module.
- Small whitespace cleanup in esp-device_conf and esp-hal


mLRS/Common/hal/esp/tx-hal-jumper-int-2400-esp32.h
```diff
@@ -0,0 +1,175 @@
...
+#define DEVICE_HAS_JRPIN5_FULL_DUPLEX
+#define DEVICE_HAS_SINGLE_LED
+#define DEVICE_HAS_NO_COM
+#define DEVICE_HAS_NO_DEBUG
+#define DEVICE_HAS_FAN_ONOFF
+#define DEVICE_HAS_ESP_WIFI_BRIDGE_ON_SERIAL
+
+
+//-- UARTS
+// UARTB = serial port to backpack WiFi bridge
```
olliw42, line 25: pl add the complete list, see e.g here: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h#L72-L77

mLRS/Common/hal/esp/tx-hal-jumper-int-2400-esp32.h
```diff
@@ -0,0 +1,175 @@
...
+
+//-------------------------------------------------------
+// ESP32, Jumper T20 Internal ELRS
+//-------------------------------------------------------
+
+// https://github.com/ExpressLRS/targets/blob/master/TX/Jumper%20T-20%202400.json
+// Added fan on GPIO 33
+
+#define DEVICE_HAS_JRPIN5_FULL_DUPLEX
+#define DEVICE_HAS_SINGLE_LED
```
olliw42, line 17: it doesn't seem to have led ?? (see lines 109ff)

mLRS/Common/hal/esp/tx-hal-jumper-int-2400-esp32.h
```diff
@@ -0,0 +1,175 @@
...
+IRAM_ATTR void esp_reset_low(void) { gpio_low(ESP_RESET); }
+
+IRAM_ATTR void esp_gpio0_high(void) { gpio_low(ESP_GPIO0); }
+IRAM_ATTR void esp_gpio0_low(void) { gpio_high(ESP_GPIO0); }
+
+#endif
+
+
+//-- POWER
+#define POWER_GAIN_DBM            28 // gain of a PA stage if present
```
olliw42, line 162: is this a known PA, which could be used from https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/hal-power-pa.h ?

olliw42:
cool

just minor comments

mLRS/Common/hal/esp/tx-hal-jumper-int-2400-esp32.h
jlpoltrack: I found this and the empty functions to be the 'best' way to prevents the leds.h throwing a bunch of error as there's no define for DEVICE _HAS_NO_LED. If you want this to be handled someway else let me know.

olliw42: ah, ok
all good as it is then
we will have to add a no led at some point

PR #254, opened 2025-01-21
Setup to flash backpack via standard tools

brad112358:
after configuring passthrough using EdgeTx CLI

Define esp_boot0() for internal Tx modules

This has been tested with 2.4 GHz internal Tx modules on Radiomaster Pocket and Jumper T15 using ETXinitPassthrough.py which will be coming in a separate PR.

It also works with the ELRS version of esptool.

### DIFF
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h
@@ -10,10 +10,9 @@
   Flashing ESP8285 backpack WiFi bridge:
   - Board: Generic ESP8266 Module, define LED_IO 16
   - First flash can be via web browser/ELRS WiFi
-  - Flash backpack before erasing ELRS or flash ELRS first using configurator via EdgeTx passthrough
   - Need ELRS version of esptool from https://github.com/ExpressLRS/Backpack
   - Power up radio, plug in USB, select VCP
-  - On Linux, run something like "Backpack/python/external/esptool/esptool.py --passthrough --port /dev/ttyACM0 --baud 460800 --before etx --after hard_reset write_flash 0x0000 ~/Arduino/build/mlrs-wireless-bridge-esp8266.ino.bin
+  - On Linux, run something like "Backpack/python/external/esptool/esptool.py --passthrough --port /dev/ttyACM0 --baud 115200 --before etx --after hard_reset write_flash 0x0000 ~/Arduino/build/mlrs-wireless-bridge-esp8266.ino.bin
 
   Flashing ESP32 module:
   - Need ELRS python folder from https://github.com/ExpressLRS/ExpressLRS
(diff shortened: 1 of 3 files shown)


mLRS/Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h
```diff
@@ -184,10 +183,16 @@ IRAM_ATTR void led_red_toggle(void) { gpio_toggle(LED_RED); }
...
 
+uint8_t esp_boot0()
+{
+    return gpio_read_activelow(ESP_BOOT0);
+}
+
+
 void esp_init(void)
 {
     gpio_init(ESP_GPIO0, IO_MODE_OUTPUT_PP_LOW); // high -> esp will start in bootloader mode
```
olliw42, line 198: no gpio_init(ESP_BOOT0, IO_MODE_INPUT_PU); needed ???

mLRS/CommonTx/esp.h
```diff
@@ -227,12 +228,12 @@ void tTxEspWifiBridge::passthrough_do_flashing(void)
     delay_ms(100); // 10 ms is too short, ESP8285 needs more time
     esp_gpio0_high();
     delay_ms(10);
+    com->SetBaudRate(baudrate); // Standard tools should specify 115200 to avoid baudrate change
 #endif
 
+    ser->SetBaudRate(baudrate);
```
olliw42, line 234: why does it have to come before leds.Initxxx() ?

mLRS/CommonTx/esp.h
```diff
@@ -227,12 +228,12 @@ void tTxEspWifiBridge::passthrough_do_flashing(void)
     delay_ms(100); // 10 ms is too short, ESP8285 needs more time
     esp_gpio0_high();
     delay_ms(10);
+    com->SetBaudRate(baudrate); // Standard tools should specify 115200 to avoid baudrate change
```
olliw42, line 231: the dependency on the if is likely incorrect
I think com baudrate it is always 115200 ... except when it is on the internal module ... so the dependency should be likely the ame a that which selects which comprot is used

why does it have to be 115200? Can't it be the CRSF baudrate?

olliw42: I guess the simplest to do is to just do it always, i.e., to bring it outside of the define

mLRS/Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h
brad112358: Since this pin is also used to enter the bootloader at power on, this has to be already taken care of.  I assume via a pull-up resistor on the board, but I haven't buzzed it out.

mLRS/CommonTx/esp.h
brad112358: I just wanted it near the previous line so it would be more clear we are setting the speed of both serial ports.

mLRS/CommonTx/esp.h
brad112358: > the dependency on the if is likely incorrect I think com baudrate it is always 115200 ... except when it is on the internal module ... so the dependency should be likely the ame a that which selects which comprot is used

Good point.  

> why does it have to be 115200? Can't it be the CRSF baudrate?

The double passthrough makes things a little complicated.  The standard esptool likes to change the backpack baud rate on the fly unless 115200 is specified on the command line  On the other hand, the ELRS version can be told not to change it.  The bottom line is if we want it to work with both the ELRS version of esptool and the combination of ETXinitPassthrough.py as it is today and the standard tools like the Arduino IDE, it's much easier i we use 115200 pretty much everywhere.

brad112358:
Is this better?

mLRS/Common/hal/esp/tx-hal-radiomaster-int-2400-esp32.h
olliw42: I'm not sure I understand that. I mean, we did do it when we wanted to use it as button, right ... from that it would make sense to me to also do it now

would it hurt to add it in the esp_init(), in the sense it would make it hot work? 
one could add a comment like // not mandatory, but play it safe

brad112358: No, it would not make it not work.  It would just make the code longer and harder to understand with 0 benefit.

The default configuration of IO_P0 is already Pull-up as documented in the datasheet.  And, it has to be this way or it wouldn't boot properly in simple designs where this pin may be left unconnected.  Further, because the internal pull-up is weak, I'm quite sure that all of these boards which actually connect a switch and wire to this pin include an external pull-up resistor or otherwise arrange to carefully control the state of this pin without depending on the internal pull-up.

In the case of the button: On ESP32, **if** we know the button is on IO_P0, there is also no need to configure this pin, but, since a button could be on a different pin for some boards, I can see justification to configure it by habit as part of a code pattern in case the code gets copied to another hal.

But, In the case of the internal module, where the wire from EdgeTx will always use IO_P0 because it is also used to program the ESP32, we can be quite sure no pin configuration will ever be needed.

Since it seems like it might make you happy, I will add some comments.

olliw42: I'm not sure you notice the inconsistency in the reasoning,but I'm indeed most happy with the comments as alternative. MANY THX.

PR #255, opened 2025-01-21
Add a slightly expanded version of ETXinitPassthrough.py

brad112358:
from the ELRS project for use with internal Tx modules.

Run without --backpack option to prepare to flash the esp32 
Run with --backpack option to prepare to flash the backpack

Tested with Radiomaster pocket and Jumper T15


brad112358:
I've combined the 2 files now.

olliw42:
btw, I've started working on setting up a GUI py script. It will be "empty" and just a scelleton at first, but we can then fill in the functuinality. Would allow us to do this here and also the receiver passthrough trick. Hope I can show you guys something by the end of the week.

olliw42:
we should have something similar now in the mLRS_Flasher repo. So am closing.

MANY THX :)

PR #256, opened 2025-01-21
Radiomaster XR4 Receiver

jlpoltrack:
- Add RM XR4 receiver target (untested)
- Add Generic ESP32C3 target (untested, but needed update to esp-peripherals for GPIO)
- Add RM XR1 receiver target (untested)
- Updated LR driver to allow negative gain, DBR4 front-end measured to lose 2 dBm from sx setting
- Simplified power levels
- Updated LR driver to allow custom RfDIO define
- Updated ESP32 RGB LED Library to latest


olliw42:
many thx, and in good time as I got my xr :+1:
I would very much appreciate having this in smaller junks however
I guess I would want a first PR for the XR4 only

jlpoltrack:
Fair enough, now just XR4 - I actually stuffed up and had removed the internal ELRS modules so needed to update anyways.

mLRS/Common/sx-drivers/lr11xx_driver.h
```diff
@@ -194,7 +194,13 @@ class Lr11xxDriverCommon : public Lr11xxDriverBase
         // Order from User Manual: SetPacketType, SetModulationParams, SetPacketParams, SetPAConfig, SetTxParams
         SetRxTxFallbackMode(LR11XX_RX_TX_FALLBACK_MODE_FS);
         SetRxBoosted(LR11XX_RX_GAIN_BOOSTED_GAIN);
-        SetDioAsRfSwitch(0b00001111, 0, 0b00000100, 0b00001000,  0b00001000, 0b00000010, 0, 0b00000001);  // ELRS hardware specific
+#ifndef RADIO_RFSW_CTRL
+        SetDioAsRfSwitch(15, 0, 4, 8, 8, 2, 0, 1);  // Default ELRS selection
+#else
+        uint8_t radioRfswCtrl[] = RADIO_RFSW_CTRL;
```
olliw42, line 200: pl use [8]

I guess I need to check what defines we have otherwise to see if RADIO_xxx fits in the pattern LOL

olliw42: the pattern is POWER_USE_xxx or SX_USE_xxx
so here it should be SX_USE_RFSW_CTRL
also these defined should be reported at the top of the file!

jlpoltrack: > also these defined should be reported at the top of the file!

Do you mean you want the SX_USE_RFSW_CTRL to be at the top of the Hal?  If not, please elaborate :)

olliw42: have a look at the other drivers, they're listed as comments in the top

jlpoltrack: Got it - should be addressed

jlpoltrack:
Got my XR4 - tested working and looks good:
- Binding works
- Telemetry works
- CRSF out on Tx2 pad works

PR #260, opened 2025-01-26
Esp jrpin5 external

brad112358:
Half duplex JRpin5 for esp32
    Leverage esp32 serial MODE_RS485_HALF_DUPLEX and GPIO_DRIVE_CAP_0 to work around the missing TC interrupt for JRpin5.

### DIFF
--- a/mLRS/Common/esp-lib/esp-uart-template.h
+++ b/mLRS/Common/esp-lib/esp-uart-template.h
@@ -59,8 +59,8 @@ typedef enum {
   #if (UART$_TXBUFSIZE > 0) && (UART$_TXBUFSIZE < 256)
     #error UART$_TXBUFSIZE must be 0 or >= 256
   #endif
-  #if (UART_RXBUFSIZE > 0) && (UART_RXBUFSIZE < 256)
-    #error UART_TXBUFSIZE must be 0 or >= 256
+  #if (UART$_RXBUFSIZE > 0) && (UART$_RXBUFSIZE < 256)
+    #error UART$_TXBUFSIZE must be 0 or >= 256
   #endif
 #endif
 
--- a/mLRS/Common/esp-lib/esp-uart-template.h
+++ b/mLRS/Common/esp-lib/esp-uart-template.h
@@ -59,8 +59,8 @@ typedef enum {
   #if (UART$_TXBUFSIZE > 0) && (UART$_TXBUFSIZE < 256)
     #error UART$_TXBUFSIZE must be 0 or >= 256
   #endif
-  #if (UART$_RXBUFSIZE < 256)
-    #error UART$_RXBUFSIZE must be >= 256
+  #if (UART_RXBUFSIZE > 0) && (UART_RXBUFSIZE < 256)
+    #error UART_TXBUFSIZE must be 0 or >= 256
   #endif
 #endif
 
--- a/mLRS/Common/hal/hal.h
+++ b/mLRS/Common/hal/hal.h
@@ -39,8 +39,8 @@ following for the tx-hal and rx-hal files.
 In tx-hal files:
 
 #define DEVICE_HAS_DIVERSITY        // board supports diversity
-#define DEVICE_HAS_JRPIN5           // board has a pin for JR bay Pin5/SPort
-#define DEVICE_HAS_JRPIN5_FULL_DUPLEX   // board has full duplex uart for JR bay Pin5 function
+#define DEVICE_HAS_JRPIN5           // board has a pin for JR bay Pin5/SPort with tc interrupt
+#define DEVICE_HAS_JRPIN5_NO_TC     // board has half or full duplex uart for JR bay Pin5 function but no tc interrupt
 #define DEVICE_HAS_IN               // board has an IN port, which supports both normal and inverted UART signals
 #define DEVICE_HAS_IN_NORMAL        // board has an IN port, which supports only normal UART signals
 #define DEVICE_HAS_IN_INVERTED      // board has an IN port, which supports only inverted UART signals
@@ -323,7 +323,7 @@ Note: Some "high-level" features are set for each device in the device_conf.h fi
 #endif // DEVICE_IS_TRANSMITTER
 
 
-#if defined DEVICE_HAS_JRPIN5 || defined DEVICE_HAS_JRPIN5_FULL_DUPLEX
+#if defined DEVICE_HAS_JRPIN5 || defined DEVICE_HAS_JRPIN5_NO_TC
    #define USE_JRPIN5
 #endif
 
--- a/mLRS/Common/hal/hal.h
+++ b/mLRS/Common/hal/hal.h
@@ -40,7 +40,6 @@ In tx-hal files:
 
 #define DEVICE_HAS_DIVERSITY        // board supports diversity
 #define DEVICE_HAS_JRPIN5           // board has a pin for JR bay Pin5/SPort with tc interrupt
-#define DEVICE_HAS_JRPIN5_NO_TC     // board has half or full duplex uart for JR bay Pin5 function but no tc interrupt
 #define DEVICE_HAS_IN               // board has an IN port, which supports both normal and inverted UART signals
 #define DEVICE_HAS_IN_NORMAL        // board has an IN port, which supports only normal UART signals
 #define DEVICE_HAS_IN_INVERTED      // board has an IN port, which supports only inverted UART signals
@@ -323,7 +322,7 @@ Note: Some "high-level" features are set for each device in the device_conf.h fi
 #endif // DEVICE_IS_TRANSMITTER
 
 
-#if defined DEVICE_HAS_JRPIN5 || defined DEVICE_HAS_JRPIN5_NO_TC
+#if defined DEVICE_HAS_JRPIN5
    #define USE_JRPIN5
 #endif
 
--- a/mLRS/CommonTx/crsf_interface_tx.h
+++ b/mLRS/CommonTx/crsf_interface_tx.h
@@ -172,8 +172,8 @@ bool tTxCrsf::transmit_start(void)
 
 // a frame is sent every 4 ms, frame length is max 64 bytes
 // a byte is 25 us
-// gaps between frames are 1 ms or so
-#define CRSF_PARSE_NEXTCHAR_TMO_US  500
+// gaps between frames are 3 ms or so
+#define CRSF_PARSE_NEXTCHAR_TMO_US  2200 // Lots of jitter when polled.
 
 
 // CRSF frame format:
@@ -200,6 +200,17 @@ void tTxCrsf::parse_nextchar(uint8_t c)
             cnt = 0;
             frame[cnt++] = c;
             state = STATE_RECEIVE_CRSF_LEN;
+            if (discarded) {
+                if (discarded > 1) {
+                    dbg.puts(u16toBCD_s(discarded));
+                    dbg.puts(" bytes lost!\n");
+                }
+                discarded = 0;
+            }
+        }
+        else {
+            // Detect discarded bytes
+            discarded++;
         }
         break;
 
@@ -325,8 +336,6 @@ bool tTxCrsf::ChannelsUpdated(tRcData* const rc)
 {
     if (!enabled) return false;
 
-    pin5_do();
-    
     CheckAndRescue();
 
     if (!channels_received) return false;
--- a/mLRS/CommonTx/crsf_interface_tx.h
+++ b/mLRS/CommonTx/crsf_interface_tx.h
@@ -173,7 +173,7 @@ bool tTxCrsf::transmit_start(void)
 // a frame is sent every 4 ms, frame length is max 64 bytes
 // a byte is 25 us
 // gaps between frames are 3 ms or so
-#define CRSF_PARSE_NEXTCHAR_TMO_US  2200 // Lots of jitter when polled.
+#define CRSF_PARSE_NEXTCHAR_TMO_US  500
 
 
 // CRSF frame format:
@@ -200,6 +200,7 @@ void tTxCrsf::parse_nextchar(uint8_t c)
             cnt = 0;
             frame[cnt++] = c;
             state = STATE_RECEIVE_CRSF_LEN;
+#ifdef USE_DEBUG
             if (discarded) {
                 if (discarded > 1) {
                     dbg.puts(u16toBCD_s(discarded));
@@ -211,6 +212,7 @@ void tTxCrsf::parse_nextchar(uint8_t c)
         else {
             // Detect discarded bytes
             discarded++;
+#endif
         }
         break;
 
--- a/mLRS/CommonTx/crsf_interface_tx.h
+++ b/mLRS/CommonTx/crsf_interface_tx.h
@@ -172,7 +172,7 @@ bool tTxCrsf::transmit_start(void)
 
 // a frame is sent every 4 ms, frame length is max 64 bytes
 // a byte is 25 us
-// gaps between frames are 3 ms or so
+// The gaps between received frames are at least 2.4 ms
 #define CRSF_PARSE_NEXTCHAR_TMO_US  500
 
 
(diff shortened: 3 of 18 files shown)


VohminV:
maybe the code example will help you

https://github.com/AlessandroAU/ESP32-CRSF-Packet-Gen/blob/master/CRSF_testcode/CRSF.ino

jlpoltrack:
More of an FYI, I measure 2.7 volts for the STM32 > ESP32 direction at the Big Bandit JST connector on the scope:

image

ESP32 > STM32 is 3.3 volts.

brad112358:
> I haven't checked it in flight, but happy module es900tx + rx_elrs_generic_900_td_pa_esp32, they work and I can control rx settings from lua script via es900tx

Thanks.  What radio handset did you test with?

brad112358:
This latest push moves to interrupt driven Rx for both internal and external modules.  It has received more extensive testing on both internal and external modules and other than the question of resistor values on the various radios, I'm quite happy with how this is working.  @olliw42 and @jlpoltrack , please let me know what you think.

mLRS/Common/esp-lib/esp-uart-template.h
```diff
@@ -59,8 +59,8 @@ typedef enum {
   #if (UART$_TXBUFSIZE > 0) && (UART$_TXBUFSIZE < 256)
     #error UART$_TXBUFSIZE must be 0 or >= 256
   #endif
-  #if (UART$_RXBUFSIZE < 256)
-    #error UART$_RXBUFSIZE must be >= 256
+  #if (UART_RXBUFSIZE > 0) && (UART_RXBUFSIZE < 256)
```
olliw42, line 62: in the template you must wirte UART_$_xxxx
the other files are then generated from the tempate 
they are now incorrect becasue of the missing $

mLRS/CommonTx/crsf_interface_tx.h
```diff
@@ -172,8 +172,8 @@ bool tTxCrsf::transmit_start(void)
 
 // a frame is sent every 4 ms, frame length is max 64 bytes
 // a byte is 25 us
-// gaps between frames are 1 ms or so
-#define CRSF_PARSE_NEXTCHAR_TMO_US  500
+// gaps between frames are 3 ms or so
+#define CRSF_PARSE_NEXTCHAR_TMO_US  2200 // Lots of jitter when polled.
```
olliw42, line 176: this makes me nervous
what is this needed for??

mLRS/CommonTx/crsf_interface_tx.h
```diff
@@ -200,6 +200,17 @@ void tTxCrsf::parse_nextchar(uint8_t c)
             cnt = 0;
             frame[cnt++] = c;
             state = STATE_RECEIVE_CRSF_LEN;
+            if (discarded) {
```
olliw42, line 203: this is only for taking statistics, right?

if so I would disable it with no debug

also, this code piece appears in few placed, can't it be a function or a macro?

mLRS/Common/hal/hal.h
```diff
@@ -39,8 +39,8 @@ following for the tx-hal and rx-hal files.
 In tx-hal files:
 
 #define DEVICE_HAS_DIVERSITY        // board supports diversity
-#define DEVICE_HAS_JRPIN5           // board has a pin for JR bay Pin5/SPort
-#define DEVICE_HAS_JRPIN5_FULL_DUPLEX   // board has full duplex uart for JR bay Pin5 function
+#define DEVICE_HAS_JRPIN5           // board has a pin for JR bay Pin5/SPort with tc interrupt
+#define DEVICE_HAS_JRPIN5_NO_TC     // board has half or full duplex uart for JR bay Pin5 function but no tc interrupt
```
olliw42, line 43: I understand your intention, but I find it cryptic
as far as I can see, this is pretty specific for ESP32 ... at least for the moment
so one could just use the old DEVICE_HAS_JRPIN5 and distunguish by ESP in the jrpin5 lib

I must say I also struggle a bit to understand how/why NO_TC can cover both full duplex and single wire, I would have thought there must be some pin managment which is not need for full duplex. But that's certainly my limitation LOL.

so, I would remove an use the old define and spilt out by esp32

mLRS/CommonTx/jr_pin5_interface.h
```diff
@@ -63,8 +63,8 @@ void (*uart_tc_callback_ptr)(void) = &uart_tc_callback_dummy;
 #define UART_RX_CALLBACK_FULL(c)    (*uart_rx_callback_ptr)(c)
 #define UART_TC_CALLBACK()          (*uart_tc_callback_ptr)()
 
-#ifdef DEVICE_HAS_JRPIN5_FULL_DUPLEX
-#include "jr_pin5_interface_full_duplex.h"
+#ifdef DEVICE_HAS_JRPIN5_NO_TC
```
olliw42, line 66: splitting out by esp should do it, IMHO
the h file could then be named also _esp.h

mLRS/CommonTx/jr_pin5_interface_no_tc.h
```diff
@@ -91,11 +88,42 @@ void tPin5BridgeBase::Init(void)
...
+    UART_SERIAL_NO.setMode(MODE_RS485_HALF_DUPLEX);
+
+    gpio_matrix_in((gpio_num_t)UART_USE_TX_IO, U1RXD_IN_IDX, true);
+    gpio_pulldown_en((gpio_num_t)UART_USE_TX_IO);
+    gpio_pullup_dis((gpio_num_t)UART_USE_TX_IO);
+
+    gpio_set_level((gpio_num_t)UART_USE_TX_IO, 0);
+    gpio_set_direction((gpio_num_t)UART_USE_TX_IO, GPIO_MODE_INPUT_OUTPUT);
+    gpio_matrix_out((gpio_num_t)UART_USE_TX_IO, U1TXD_OUT_IDX, true, false);
+    gpio_set_drive_capability((gpio_num_t)UART_USE_TX_IO, GPIO_DRIVE_CAP_0);
```
olliw42, line 125: it probably would usefull to have a comment here for why

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -752,7 +752,7 @@ RESTARTCONTROLLER
     sx_serial.Init(&serial, &mbridge, &serial2); // ports selected by SerialDestination, ChannelsSource
     cli.Init(&comport);
 #ifdef USE_ESP_WIFI_BRIDGE
-  #ifdef DEVICE_HAS_JRPIN5_FULL_DUPLEX
+  #if defined DEVICE_HAS_JRPIN5_NO_TC && UART_USE_TX_IO != UART_USE_RX_IO
```
olliw42, line 755: aha ... here there is a difference
I don't like that at all, super cryptic and things like the UART should really be used only low level. Just think about what happens when it's not the UART but a different UARTX ... no.
either:
- option a: this part is done in the hal.h and a respective USE define is set and used everywhere higher up
- option b: the DEVICE_JRPIN5_FULL_DUPLEX is kept
I guess I would prefer b as it is just easy to comprehend

olliw42:
some coments which are really just code structure, not substantial
generally it really looks quite clean and thus looks like this is how it should look like. VERY nice.

mLRS/CommonTx/jr_pin5_interface_no_tc.h
```diff
@@ -91,11 +88,42 @@ void tPin5BridgeBase::Init(void)
...
     telemetry_start_next_tick = false;
     telemetry_state = 0;
 
     pin5_init();
+
+    UART_SERIAL_NO.setRxTimeout(1);
+    // A small value (like 4) for RxFIFOFull has a higher impact on main loop time,
+    // but reduces the delay between JRpin5 receive and transmit.  Choose large for now.
+    // The documented default is 128, but seems to actually be smaller than 64 when RXBUFSIZE = 0
+    UART_SERIAL_NO.setRxFIFOFull(64);
```
jlpoltrack, line 102: Why these two lines?  They are already set in the UART lib to 1 / 8: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/esp-lib/esp-uart-template.h#L162-L163 (so at least first line could be omitted)

mLRS/CommonTx/crsf_interface_tx.h
brad112358: Oops.  It made me nervous too which is why I switched both full and half duplex to use the receive isr.  This can go back to 500.

mLRS/CommonTx/jr_pin5_interface_no_tc.h
brad112358: I had missed that they were set there.  That explains why the default didn't seem to be correct.  64 seemed to have less impact on the maximum time spent in the main loop than 8 did.

brad112358:
I think I've accounted for review comments here, but I won't be testing this until tomorrow.

mLRS/CommonTx/crsf_interface_tx.h
```diff
@@ -172,7 +172,7 @@ bool tTxCrsf::transmit_start(void)
 
 // a frame is sent every 4 ms, frame length is max 64 bytes
 // a byte is 25 us
-// gaps between frames are 1 ms or so
+// gaps between frames are 3 ms or so
```
olliw42, line 175: the comment is true?
with 4 ms slices and 2x64 bytes at 400000 where can't be 3 ms space in general. Is this referring to the time from end of a received frame to start of the next receiving frame? If so maybe ... // gaps between incomming frames ....

olliw42:
great. I left few comments which are really nitpick
I guess only removing the DEBUG is relevant, so we have it nicely in the PR
I will when tidy up the other pieces. It's easier for me to see how it looks when I have it in the CubeIDE :)

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -91,11 +83,43 @@ void tPin5BridgeBase::Init(void)
...
+  #define UART_SERIAL_NO       Serial1
+#elif defined UART_USE_SERIAL2
+  #define UART_SERIAL_NO       Serial2
+#else
+  #error UART_SERIAL_NO must be defined!
+#endif
+    UART_SERIAL_NO.setMode(MODE_RS485_HALF_DUPLEX);
+
+    gpio_matrix_in((gpio_num_t)UART_USE_TX_IO, U1RXD_IN_IDX, true);
+    gpio_pulldown_en((gpio_num_t)UART_USE_TX_IO);
```
jlpoltrack, line 112: Did you find this pulldown necessary? A quick test disabling it shows no difference and as the ESP32 side is always actively driven this ~50K wouldn't have any impact one way or the other.

olliw42: I would strongly suggest to have it enabled
it really can depend on the hardware, and just because it has no effect when disabled is not a reason to disable it. I would argue the other way around. Electrically there just should be a pulldown somewhere.

olliw42: oh ... you mean why on TX ??
there should be on RX ...

I guess I also would do first the _dis() and then the _en()

is this code part really for JR_PIN5_FULL_DUPLEX ??? isn't it for #ifndef JR_PIN5_FULL_DUPLEX ??

if so, I guess I would first dis all pulls, when configure the TX, and when enable the pulldown

olliw42: so I guess acomment after the #ifdef saying what this code is for could be usefull LOL

VohminV:
> > I haven't checked it in flight, but happy module es900tx + rx_elrs_generic_900_td_pa_esp32, they work and I can control rx settings from lua script via es900tx
> 
> Thanks. What radio handset did you test with?

Radiomaster tx12, but everything worked in mspx

mLRS/CommonTx/jr_pin5_interface_esp.h
jlpoltrack: This is just in half-duplex - based on Line 109.

Was thinking that the single ESP32 pin is set as an output and should always be actively driven, so the internal 50K pull-down shouldn't really matter.  But since this has been working, agree that it is probably best not to modify at this point.

olliw42: yes, the code looks like so ... but it is wrapped in #ifdef JR_PIN5_FULL_DUPLEX :)

mLRS/CommonTx/mbridge_interface.h
```diff
@@ -174,7 +174,22 @@ void tMBridge::parse_nextchar(uint8_t c)
 
     switch (state) {
     case STATE_IDLE:
-        if (c == MBRIDGE_STX1) state = STATE_RECEIVE_MBRIDGE_STX2;
+        if (c == MBRIDGE_STX1) {
+            state = STATE_RECEIVE_MBRIDGE_STX2;
+#ifdef USE_DEBUG
```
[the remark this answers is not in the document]
brad112358: It looks OK to me.  This is just about detecting if the driver isn't discarding echoed bytes in half duplex as it should or if we enable receive too late in STM or if we step on receive with transmit.  I don't care about an accurate count, so If we don't get MBRIDGE_STX1 as the first character after entering the idle state, we discard a byte and we want to record that and then, report when we do finally receive STX1.  Am I missing something?  Where do you think it should be?

mLRS/CommonTx/crsf_interface_tx.h
brad112358: The original comment seemed wrong to me.  The gaps between send and receive are ~0 or ~3 depending on if you mean receive -> send or send->receive.  There were no 1ms gaps to be seen on the scope.  How about "The gaps between received frames are at least 2.4 ms"

mLRS/CommonTx/jr_pin5_interface_esp.h
brad112358: Oops.  I'm glad I mentioned I didn't test this!  It should be "#ifndef JR_PIN5_FULL_DUPLEX"

As for the pulldown, I think I'll remove it.  I had intended to experiment with pullup instead to see if I could shift the voltage range up just a bit, but I think both are too weak to make much difference.

mLRS/CommonTx/crsf_interface_tx.h
```diff
@@ -200,6 +200,19 @@ void tTxCrsf::parse_nextchar(uint8_t c)
             cnt = 0;
             frame[cnt++] = c;
             state = STATE_RECEIVE_CRSF_LEN;
+#ifdef USE_DEBUG
```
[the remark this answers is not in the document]
olliw42: YES, I know, but I haven't anticipated enough how it would look like :)
I'm ok with having it in for a while

my suggestion was to remove it in this PR so the PR is cleaner

mLRS/CommonTx/crsf_interface_tx.h
olliw42: I think the fact you don't see 1ms gaps on the scope is ONLY because there are only short messages!
as said, at 400000 2x64 bytes are 3.2 ms which need to fit in the 4 ms slot, so technically it are only 0.8 ms !!

mLRS/CommonTx/mbridge_interface.h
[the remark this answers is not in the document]
olliw42: have you tried with debug disabled?
from teh indenting it seems to me the #endif can't be in the right place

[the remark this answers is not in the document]
brad112358: I temporarily changed the ifdef to ifndef and yes, it works.  Imagine the entire section gone, that's what the original code was before I added this check.  If I broke it up into two ifdef sections, it would probably look better to you because a "}" gets moved to between the ifdef blocks.  I guess I'll go ahead and do that to improve readability and end this discussion.

brad112358:
Hopefully this looks better.

brad112358:
I did do some testing on the latest version today on the Bandit and the internal Jumper.   It still looks good here.

olliw42:
MANY THX @brad112358. This is a really big thing.

Issue #258, opened 2025-01-24
Fail on Big Bandit with CLI + Help Command


jlpoltrack:
On current Main branch when using the CLI on Big Bandit, executing the help command causes the Tx to fail with rapid LED flash and become unresponsive.

PR #261, opened 2025-01-26
tx, cli delay tweak

jlpoltrack:
To address: https://github.com/olliw42/mLRS/issues/258

Need to stay under ESP32 Tx FIFO buffer of 128, the help command is > 128 and this delay was therefore ineffective.

### DIFF
--- a/mLRS/CommonTx/cli.h
+++ b/mLRS/CommonTx/cli.h
@@ -313,7 +313,7 @@ class tTxCli
 #elif !defined DEVICE_HAS_COM_ON_USB && (UARTC_TXBUFSIZE >= 2048)
     void delay(void) {}
 #else
-    void delay(void) { if (put_cnt > 192) { delay_ms(15); put_cnt -= 128; } } // 115200 -> 128 bytes = 11 ms, usb txbuf is small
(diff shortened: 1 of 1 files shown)


olliw42:
great
we need to check this for all other tx modules too, also the diy ones ... over time I did lot's of changes here like kind of back and forth since there always were issues here or where coming up. If we can't find acoomon algorithm which works for all, we may have to separate for ESP and STM32. I would hate it of we would have to distinuguish by target.

olliw42:
it could makes sense to make it dependent on UARTC_TXBUFSIZE like a line before
that is
#elif UARTC_TXBUFSIZE <= 128
here your change
#else
here the old line

pl nevertheless check/explain why put_cnt = 0 should be correct

olliw42:
do we understand why does this hit only the bandit, but not other esp tx modules (which havethe same uart fifo) ?

olliw42:
I am wondering now if the if shouldn't be replaced by a while (put_cnt > xxx) ...

jlpoltrack:
Fix I suggested appears to be no good while doing a 'pl' with a receiver connected, so will close and spend some time on.

PR #265, opened 2025-01-29
Split help and pl output accross multiple calls to cli.do()

### LINKED ISSUE #258: Fail on Big Bandit with CLI + Help Command

brad112358:
Fixes #258

### DIFF
--- a/mLRS/CommonTx/cli.h
+++ b/mLRS/CommonTx/cli.h
@@ -288,13 +288,10 @@ class tTxCli
     typedef enum {
         CLI_STATE_NORMAL = 0,
         CLI_STATE_STATS,
+        CLI_STATE_HELP,
+        CLI_STATE_PL,
+        CLI_STATE_FREQS,
     } CLI_STATE_ENUM;
-
-    typedef enum {
-        CLI_OUTPUT_NONE = 0,
-        CLI_OUTPUT_HELP,
-        CLI_OUTPUT_PL,
-    } CLI_OUTPUT_ENUM;
 
     void addc(uint8_t c);
     void clear(void);
@@ -310,21 +307,9 @@ class tTxCli
     bool is_cmd_param_set(char* const name, char* const svalue);
     bool is_cmd_set_value(const char* const cmd, int32_t* const value);
 
-    uint16_t put_cnt;
-    void delay_off(void) { put_cnt = 0; }
-    void delay_clear(void) { put_cnt = 1; }
-//    void delay(void) { if (put_cnt > 768) { delay_ms(40); put_cnt -= 512; } } // 115200 -> 512 bytes = 44 ms
-#if defined DEVICE_HAS_COM_ON_USB && (USB_TXBUFSIZE >= 2048)
-    void delay(void) {}
-#elif !defined DEVICE_HAS_COM_ON_USB && (UARTC_TXBUFSIZE >= 2048)
-    void delay(void) {}
-#else
-    void delay(void) { if (put_cnt > 192) { delay_ms(15); put_cnt -= 128; } } // 115200 -> 128 bytes = 11 ms, usb txbuf is small
-#endif
-
-    void putc(char c) { com->putc(c); if (put_cnt) put_cnt++; delay(); }
-    void puts(const char* s) { com->puts(s); if (put_cnt) put_cnt += strlen(s); delay(); }
-    void putsn(const char* s) { com->puts(s); com->puts(CLI_LINEND); if (put_cnt) put_cnt += strlen(s)+strlen(CLI_LINEND); delay(); }
+    void putc(char c) { com->putc(c); }
+    void puts(const char* s) { com->puts(s); }
+    void putsn(const char* s) { com->puts(s); com->puts(CLI_LINEND); }
 
     void print_layout_version_warning(void);
     void print_config_id(void);
@@ -340,11 +325,9 @@ class tTxCli
     uint8_t task_pending;
     int32_t task_value;
 
-    CLI_OUTPUT_ENUM output_pending;
-    int16_t output_index;
-    int8_t output_flag;
-
     uint8_t state;
+    int8_t print_index;
+    int8_t print_pl_flag;
 };
 
 
@@ -361,13 +344,9 @@ void tTxCli::Init(tSerialBase* const _comport)
     task_pending = TX_TASK_NONE;
     task_value = 0;
 
-    output_pending = CLI_OUTPUT_NONE;
-    output_index = 0;
-    output_flag = 0;
-
     state = CLI_STATE_NORMAL;
-
-    put_cnt = 0;
+    print_index = 0;
+    print_pl_flag = 0;
 }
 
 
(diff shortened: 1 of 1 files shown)


jlpoltrack:
Quick test on Big Bandit - pl command no longer causes fail state.

olliw42:
nice, so you suggest to follow the idea to output in smaller chunks
like that
on first look my first impulse is that the two fields for managing the output, output_pending and output_index, may go into one, by adding "substructure" to the enum. There thus would e.g. be two entries for the pl-tx chunck, one which just does it, and one as part of the sequence of pl. It's then more like a simple state machine. I might be wrong but I think worth a look.

mLRS/CommonTx/cli.h
```diff
@@ -290,11 +290,17 @@ class tTxCli
         CLI_STATE_STATS,
     } CLI_STATE_ENUM;
 
+    typedef enum {
+        CLI_OUTPUT_NONE = 0,
+        CLI_OUTPUT_HELP,
+        CLI_OUTPUT_PL,
+    } CLI_OUTPUT_ENUM;
```
olliw42, line 297: may be PRINT instead of OUTPUT, as the functions which it manages are also called print

mLRS/CommonTx/cli.h
```diff
@@ -334,6 +340,10 @@ class tTxCli
     uint8_t task_pending;
     int32_t task_value;
 
+    CLI_OUTPUT_ENUM output_pending;
```
olliw42, line 343: maybe PRINT/print instead of OUTPUT/output

mLRS/CommonTx/cli.h
```diff
@@ -334,6 +340,10 @@ class tTxCli
...
+    int16_t output_index;
```
olliw42, line 344: maybe print_chunk_idx, more descriptie IMHO and not so easily confused with the param idx, it's also a bit more along what we use in mavlink

mLRS/CommonTx/cli.h
```diff
@@ -334,6 +340,10 @@ class tTxCli
...
+    int8_t output_flag;
```
olliw42, line 345: maybe print_pl_flag (as it's only for pl AFAICS)

mLRS/CommonTx/cli.h
```diff
@@ -351,6 +361,10 @@ void tTxCli::Init(tSerialBase* const _comport)
     task_pending = TX_TASK_NONE;
     task_value = 0;
 
+    output_pending = CLI_OUTPUT_NONE;
+    output_index = 0;
+    output_flag = 0;
```
olliw42, line 366: while a long due leftover, I think we could use the opportunity to also add an enum for the flags ... was ugly to read so far

mLRS/CommonTx/cli.h
```diff
@@ -543,22 +557,29 @@ void tTxCli::print_param(uint8_t idx)
...
+    output_pending = CLI_OUTPUT_PL;
+    if (output_index == 0 && (output_flag == 0 || output_flag == 1 || output_flag == 3) && !connected()) {
         putsn("warn: receiver not connected");
     }
 
-    for (uint8_t idx = 0; idx < SETUP_PARAMETER_NUM; idx++) {
-        if ((flag == 1) && (setup_param_is_tx(idx) || setup_param_is_rx(idx))) continue;
-        if ((flag == 2) && !setup_param_is_tx(idx)) continue;
-        if ((flag == 3) && !setup_param_is_rx(idx)) continue;
+    for (uint8_t count = 0; output_index < SETUP_PARAMETER_NUM; output_index++) {
```
olliw42, line 567: no, pl don't do this kind of K&R style C weiredness, either a clear plain for or a clear plain while

mLRS/CommonTx/cli.h
```diff
@@ -543,22 +557,29 @@ void tTxCli::print_param(uint8_t idx)
...
+        if ((output_flag == 3) && !setup_param_is_rx(output_index)) continue;
 
-        if ((flag == 0 || flag == 3) && !connected() && setup_param_is_rx(idx)) continue;
+        if ((output_flag == 0 || output_flag == 3) && !connected() && setup_param_is_rx(output_index)) continue;
 
-        print_param(idx);
-        //delay_ms(10);
+        print_param(output_index);
+        count++;
+        if (count > 1) {
```
olliw42, line 576: see previous comment

mLRS/CommonTx/cli.h
```diff
@@ -639,36 +660,46 @@ char unit[32];
...
+        putsn("  pl          -> list all parameters");
+        putsn("  pl c        -> list common parameters");
+        putsn("  pl tx       -> list Tx parameters");
+        putsn("  pl rx       -> list Rx parameters");
+
+        putsn("  p name          -> get parameter value");
+        putsn("  p name = value  -> set parameter value");
+        putsn("  p name = ?      -> get parameter value and list of allowed values");
+        putsn("  pstore      -> store parameters");
+        output_index++; // Do next chunk next time
```
olliw42, line 677: this here would be a clear example where one could use just the enum with more subsructure to step through the pieces

mLRS/CommonTx/cli.h
```diff
@@ -683,6 +714,14 @@ bool rx_param_changed;
 
     //puts(".");
 
+    if (output_pending == CLI_OUTPUT_HELP) {
```
olliw42, line 717: maybe a switch

mLRS/CommonTx/cli.h
```diff
@@ -710,10 +749,10 @@ bool rx_param_changed;
         if (is_cmd("help"))  { print_help(); } else
         if (is_cmd("?"))     { print_help(); } else
         if (is_cmd("v"))     { print_device_version(); } else
-        if (is_cmd("pl"))    { print_config_id(); print_param_list(0); } else
-        if (is_cmd("pl c"))  { print_config_id(); print_param_list(1); } else
-        if (is_cmd("pl tx")) { print_config_id(); print_param_list(2); } else
-        if (is_cmd("pl rx")) { print_config_id(); print_param_list(3);
+        if (is_cmd("pl"))    { print_config_id(); output_flag = 0; print_param_list(); } else
```
olliw42, line 752: that's a bit weired from a logic point of view, as the output state machine is setup in mixed places, not so nice architecture wise. I guess I would separate into two functions, one to start the print and one to handle the prints. If one keeps on with the multiple fields
if all fields can be condensed into on enum, one would need just one handler, and could set the task here

for the print_help() we have the same logocal issue that the same function needs to do both start/trigger and handling. There it's even weirder as the output_index is used to start the state machine.

I'm not sure form scratch what the nicest solution would be but I think we could spend few efforts on making this look nice

olliw42:
if I'm understanding correctly, the pl is chunked into the sections c, tx, rx. I'm not sure I totally understand the timing aspect here. cli.Do() is done only in the while start part, that is each start of transmission. What would happen with FLRC? How would it interfere with the delay thing? I would think that with proper chunking and timing the delay stuff should go, i.e., any proper timing should be done now in the print-chunk handler. 

Has this been tested on anything else than the bandit?

jlpoltrack:
> Has this been tested on anything else than the bandit?

Tests on BetaFPV with pl command with receiver connected:

- FLRC: link drops temporarily
- 50 Hz: lq drop on Tx side (~80%)
- 31 Hz: lq drop on Tx side (~70%)
- 19 Hz: no observed LQ drop

brad112358:
> nice, so you suggest to follow the idea to output in smaller chunks like that on first look my first impulse is that the two fields for managing the output, output_pending and output_index, may go into one, by adding "substructure" to the enum. There thus would e.g. be two entries for the pl-tx chunck, one which just does it, and one as part of the sequence of pl. It's then more like a simple state machine. I might be wrong but I think worth a look.

I don't think so.  You may have missed that this would require about 24 additional states because we are only doing 2 lines of pl output on every call to do().  This is because even help which was only taking 2 or 3 ms can cause a problem so I wanted to keep each call under 2ms.

mLRS/CommonTx/cli.h
brad112358: In the case of help, it is indeed a chunk index, but in the case of pl, it actually is a parameter index!  In both cases, it's part of our print state and if we are short of RAM, I'd rather not use more of it.  I think I'll call it print index.

brad112358:
> Tests on BetaFPV with pl command with receiver connected:

Thanks for testing!

mLRS/CommonTx/cli.h
brad112358: OK.  Sometimes I think your code is weird too. :)

mLRS/CommonTx/cli.h
brad112358: Yes, maybe; especially if we had more states.

brad112358:
> I would think that with proper chunking and timing the delay stuff should go, i.e., any proper timing should be done now in the print-chunk handler.

That's an interesting question.  Currently, the put_cnt flow control hack is completely independent of this PR which seeks only to restrict the main loop delay.  But, I suppose we could break help up into many more pieces and eliminate put_cnt.

I also just realized I forgot to test listfreqs and it is currently causing a crash so we need the same tricks there.

jlpoltrack:
> I'm not worried about the link when we use the CLI, but without this PR, was the LQ drop worse?  Yes, LQ drop is worse on Main.

Tests on BetaFPV with pl command with receiver connected:

- FLRC: 
    - This branch: link drops temporarily
    - Main: link drops temporarily
 - 50 Hz: 
     - This branch: lq drop on Tx side (~80%)
     - Main: link drops temporarily
 - 31 Hz:
    - This branch: lq drop on Tx side (~70%)
    - Main: lq drop on Tx side (~80%)
 - 19 Hz: 
     - This branch: no observed LQ drop
     - Main: lq drop on Tx side (~70%)

Still curious why the fail state seems to only occur with SX127x, perhaps there is better state handling with SX128x.

mLRS/CommonTx/cli.h
brad112358: Is the latest commit better?  It now follows the pattern you had established with CLI_STATE_STATS

brad112358:
> I would think that with proper chunking and timing the delay stuff should go, i.e., any proper timing should be done now in the print-chunk handler.

I did end up subsuming the put delay hacks into the print chunk code and I tested it with TXBUFSIZE 0 (128 bytes).  cli.Do() takes under 500 uS now in my tests.

mLRS/CommonTx/cli.h
brad112358: Since this is unrelated and you didn't mention what you would want to call the enums, I think I'll pass on this and let you fix it later as you see fit, lest this require another round of changes due to my likely poorly chosen names.

brad112358:
> Still curious why the fail state seems to only occur with SX127x, perhaps there is better state handling with SX128x.

I suspect it's the same reason we saw crashes only with the SX127x when we had the jitter issue.

I suspect my latest revision my have further improved the LQ drops, but I haven't tested anything but 19 Hz.

brad112358:
I fixed a rather significant typo in my comment two up.  The time for Cli.Do() is way down, measuring at less than 50 uS in my tests.  Is there anything holding up merging this now?

Issue #245, opened 2024-12-13
Crash on DIY E28 dual board when diversity is disabled.


brad112358:
When I use mLRS-bw.lua to switch from diversity to antenna 1 or antenna 2 and power cycle, it seems to be crashing very quickly after power-up, leaving the red LED on solid.  Changing other settings and power cycling seems fine.  The SWUART debug port works fine  in diversity mode and I see a "." printed every second, but when I power cycle after selecting antenna 1 or antenna 2, I'm not getting anything at all on the debug port, just the solid red LED.  Note also that if I disconnect JRPIN5 it powers up OK and the red LED flashes as it should.

This issue does not appear when DEVICE_HAS_DIVERSITY is not defined.

brad112358:
It works in branch v1.3

brad112358:
I ran a git bisect.  It took about 8 steps.
~/git/mlrs$ git bisect bad
1ebfbc6fd384a7ae071ef95d8e4959ff22b04bf7 is the first bad commit
commit 1ebfbc6fd384a7ae071ef95d8e4959ff22b04bf7
Author: olliw42 <waldmanns@gmx.de>
Date:   Sun Sep 22 15:09:12 2024 +0200

    tx,rx, make rf power changeable without save/store

 lua/mLRS.lua                           |  9 ++++-
 mLRS/Common/rf_power.h                 | 61 ++++++++++++++++++++++++++++++++++
 mLRS/Common/sx-drivers/sx126x_driver.h |  6 ++++
 mLRS/Common/sx-drivers/sx127x_driver.h |  6 ++++
 mLRS/Common/sx-drivers/sx128x_driver.h |  6 ++++
 mLRS/Common/sx-drivers/sx12xx_driver.h |  1 +
 mLRS/CommonRx/mlrs-rx.cpp              |  4 +++
 mLRS/CommonTx/mlrs-tx.cpp              |  6 +++-
 8 files changed, 97 insertions(+), 2 deletions(-)
 create mode 100644 mLRS/Common/rf_power.h

jlpoltrack:
On E28 Dual, if I set to Antenna1 to end up in a crash state, then comment out the sx2.UpdateRfPower line here and flash things seem to work: https://github.com/olliw42/mLRS/commit/1ebfbc6fd384a7ae071ef95d8e4959ff22b04bf7#diff-d442ec9fc39334fee02e97533bc34e173409b31ffbd4f4e056bfd97b76cbb18cR50-R51

[...]

PR #266, opened 2025-01-29
Fix single antenna usage on diversity setups

jlpoltrack:
To address https://github.com/olliw42/mLRS/issues/245


jlpoltrack:
Closing, this as it will be handled here https://github.com/olliw42/mLRS/pull/302 (and there are merge conflicts :) )

PR #267, opened 2025-01-30
Esp32 RGB LED Library

jlpoltrack:
To replace Neopixelbus library, primary motivations:

- Neopixelbus only allows one to use the RMT peripheral on ESP32C3, using I2S peripheral should generate fewer interrupts.
- Allows more flexibility as uses ESP-IDF I2S driver.

Tested working on:
- Plain ESP32 (Radiomaster DBR4) 
- ESP32C3 (Radiomaster XR1)
- ESP32S3 (Devboard)


jlpoltrack:
- Does not work with Flash Freq = 80 MHz + Flash Mode = QIO which decrease Tx jitter
- Shows slightly worse jitter than NeoPixelBus

PR #270, opened 2025-01-31
generic ghz to reg

jlpoltrack:
Replaces SX1280_FREQ_GHZ_TO_REG with SX12XX_FREQ_GHZ_TO_REG to make 2.4 on LR11xx a little nicer.  Will do corresponding PR for sx12xx library.


olliw42:
I guess SX12xx is a bit confusing as it also includes sx127x and sx126x ... which I guess is not the intention here

olliw42:
reopening
I'm sorry, I completedly missunderstood the PR.

I in fact think this is a very good idea, as it aligns the sx1280 handling in the fhss lib with what we do for the other sx.

What I don't understand though is why it needs a submodule change? (I guess this has tripped me off)

jlpoltrack:
> What I don't understand though is why it needs a submodule change? (I guess this has tripped me off)

The submodule update is needed as that's where the function is defined.  e.g. https://github.com/olliw42/sx12xx-lib/pull/19/files#diff-5760f0bd9b174c85e75b1601566f8acaa822ccb72e23bd584dd25214849ebbaaR22

olliw42:
ah, I see, there are two things going on ... it's also to prepare for lr 2.4 ... 

but: this submodule udate is nopt yet in, i.e., merging would likely leave a mess, right? Even if one would merge the submodule PR it probably gives a mess since the update here won't point to the correct head, right?
i.e., I have feeling that this as is gives troubles 

I just have merged your PR to the sx lib submodule, and will bring it in to here in a moment. That is, I think you'll have to update this PR here.

jlpoltrack:
> That is, I think you'll have to update this PR here.

Done.  Seems to be just a two-step process, once the submodule gets merged this one needs a final update.

olliw42:
ähm ... this PR should not need a submodule update anymore ... right?

jlpoltrack:
No idea why, they have the same commit hash and I tried running git submodule update --init --recursive too
image

jlpoltrack:
Okay there we go, I merged main back into this and the submodule stuff goes away.

PR #272, opened 2025-02-03
nomad tx

jlpoltrack:
Title

### DIFF
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
@@ -23,9 +23,9 @@
 
 
 //-- UARTS
-// UARTB = serial port BT/ESP port
+// UARTB = serial port 
 // UARTC = COM (CLI)
-// UARTD = serial2
+// UARTD = serial2 BT/ESP port
 // UART  = JR bay pin5
 // UARTE = in port, SBus or whatever
 // UARTF = debug port
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
@@ -1,0 +1,352 @@
+//*******************************************************
+// Copyright (c) MLRS project
+// GPL3
+// https://www.gnu.org/licenses/gpl-3.0.de.html
+//*******************************************************
+// hal
+//*******************************************************
+
+//-------------------------------------------------------
+// ESP32, Radiomaster Tx Nomad
+//-------------------------------------------------------
+
+// https://github.com/ExpressLRS/targets/blob/master/TX/Radiomaster%20Nomad.json
+
+#define DEVICE_HAS_JRPIN5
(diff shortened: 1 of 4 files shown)


mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
```diff
@@ -0,0 +1,352 @@
...
+#define DEVICE_HAS_DIVERSITY_SINGLE_SPI
+#define DEVICE_HAS_SINGLE_LED_RGB
+#define DEVICE_HAS_SERIAL_OR_COM
+#define DEVICE_HAS_FAN_ONOFF
+#define DEVICE_HAS_ESP_WIFI_BRIDGE_ON_SERIAL2
+#define DEVICE_HAS_ESP_WIFI_BRIDGE_CONFIGURE  // requires backpack firmware to have AT mode enabled
+
+
+//-- UARTS
+// UARTB = serial port BT/ESP port
```
brad112358, line 26: UARTB isn't BT/ESP port here.  Should this e changed?

mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
```diff
@@ -0,0 +1,352 @@
...
+#define UARTB_RXBUFSIZE           TX_SERIAL_RXBUFSIZE
+
+#define UARTC_USE_SERIAL // COM (CLI), is connected to USB-C via USB<>UART
+#define UARTC_BAUD                115200
+#define UARTC_USE_TX_IO           IO_P1
+#define UARTC_USE_RX_IO           IO_P3
+#define UARTC_TXBUFSIZE           0  // TX FIFO = 128
+#define UARTC_RXBUFSIZE           TX_COM_RXBUFSIZE
+
+#define UART_USE_SERIAL1 // JR bay pin5
```
brad112358, line 47: Do you want the UART order here to be the same as the comment above?

mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
jlpoltrack: Good catch, moved to UARTD Serial2

mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-900-esp32.h
jlpoltrack: Well all other HALs use this order with UART after UARTB and UARTC.

olliw42: pl use the order/sequence as in all hals
it makes it easier for us to be consistent

olliw42:
MANY thx.
I'd frankly like to keep that a bit on hold however. We have some other tx-es to work on, and I like to let the tx things settle a bit, lots of stuff were going here recently. And the lr11xx support isn't full yet. No 19Hz7x, and we also should allow switching to 2.4 (dual band we can keep for later). This we can work out better on the rx.

jlpoltrack:
Sounds good, will put this on the back burner.

PR #273, opened 2025-02-03
Bridge esp32 c3 ifdef

jlpoltrack:
Bridge on ESP32C3 requires core 2.0.17 otherwise the WiFi name and channel aren't able to be set correctly.

### DIFF
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
@@ -151,16 +151,14 @@ String bluetooth_device_name = ""; // "mLRS BT"; // Bluetooth device name, "" re
 //-------------------------------------------------------
 
 #if CONFIG_IDF_TARGET_ESP32C3
-#if ESP_ARDUINO_VERSION != ESP_ARDUINO_VERSION_VAL(2, 0, 17)
+  #if ESP_ARDUINO_VERSION != ESP_ARDUINO_VERSION_VAL(2, 0, 17)
     #error ESP32C3 requires ESP Arduino Core 2.0.17 !
-#endif
-#else
-#if ESP_ARDUINO_VERSION < ESP_ARDUINO_VERSION_VAL(3, 0, 0)
+  #endif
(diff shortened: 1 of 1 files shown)


olliw42:
so what this means on needs an OLD version of Arduino??
doesn't this give a conflict with ones setup? I mean, can one have both at the same time without hazzles?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -150,11 +150,17 @@ String bluetooth_device_name = ""; // "mLRS BT"; // Bluetooth device name, "" re
 // Includes
 //-------------------------------------------------------
 
+#if CONFIG_IDF_TARGET_ESP32C3
+#if ESP_ARDUINO_VERSION != ESP_ARDUINO_VERSION_VAL(2, 0, 17)
```
olliw42, line 154: indent of two spaces

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -150,11 +150,17 @@ String bluetooth_device_name = ""; // "mLRS BT"; // Bluetooth device name, "" re
...
+    #error ESP32C3 requires ESP Arduino Core 2.0.17 !
+#endif
+#else
```
olliw42, line 157: can't this be an #elif, and not need the last #endif?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
jlpoltrack: Done

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
jlpoltrack: Yes, updated.

olliw42:
could you pl respond to https://github.com/olliw42/mLRS/pull/273#issuecomment-2632607416 :)

jlpoltrack:
> doesn't this give a conflict with ones setup? I mean, can one have both at the same time without hazzles?

Switching between the two versions takes a few minutes to download and install as the IDE doesn't install them side by side.

PR #274, opened 2025-02-03
Bandit: Disable debug port and restore com/serial

brad112358:
Also Somewhat allign UARTs setup with JLP's nomad PR


brad112358:
FYI, I did test this PR on big Bandit before I pushed, but I think most of the internal modules are probably currently broken in main and I don't plan to test again there until the DEVICE_HAS_ESP_WIFI_BRIDGE... stuff is resolved.

PR #280, opened 2025-02-07
Add ArduPilot tool, link statistics Lua script (thx to twistedwings)

olliw42:
Adds a somehwat tidied version of twistedwings Lua script for ArduPilot, to log the link statistics data.

I could not give it any testing myself. So, needs to be tested if my changes have screwed it up :)

@jlpoltrack

### DIFF
--- a/ardupilot_tools/mlrs_mavlink_link_stats.lua
+++ b/ardupilot_tools/mlrs_mavlink_link_stats.lua
@@ -64,7 +64,10 @@ function decode_header(msg)
     -- build up a map of the header
     local header = {}
 
-    local pos = 3 -- TODO: why that?
+    -- mavlink.receive_chan() returns as first result a mavlink_message_t structure
+    -- which holds the message crc in the first two bytes, the magic STX and rest of the message are following
+    -- the STX is thus found in the third byte
+    local pos = 3
 
     -- Magic packet start marker, can be used to identify the MAVLink version
     header.stx, pos = string.unpack("<B", msg, pos)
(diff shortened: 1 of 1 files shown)


ardupilot_tools/mlrs_mavlink_link_stats.lua
```diff
@@ -0,0 +1,208 @@
...
+            --gcs:send_text(6, string.format("Received TUNNEL from %d %d", header.sysid, header.compid))
+            if header.sysid == RADIO_LINK_SYSTEM_ID and header.compid == MAV_COMP_ID_TELEMETRY_RADIO then
+                handle_tunnel(msg, pos)
+            end
+        end
+
+        return update() -- give it an immediate second try
+    end
+
+    return update(), 10 -- run at 100 Hz
```
jlpoltrack, line 204: Should be 'return update, 10' (no parentheses)

olliw42: hm...
the () in line 204 were also in twistedwings version which did work for him. The () in line 201 are due to me. I think it doesn't matter.
you have evidence that without the () it works for you but not with the ()?

jlpoltrack: Yes, it works for me with this change otherwise it doesn't.  More details here: https://discord.com/channels/1005096100572700794/1163909355234656267/1337424229800345611

olliw42:
corrected. THX.

did the instructions in the header work for you, or anything one should add?

jlpoltrack:
> did the instructions in the header work for you, or anything one should add?

My only suggestion would be to have something like 'mLRS Receiver Requirements' instead of 'Configuration of mLRS receiver' to indicate that the two settings are mandatory.

PR #282, opened 2025-02-08
Fewer Sx Calls

jlpoltrack:
Remove a few unnecessary Sx calls:

- ReadBuffer used instead of ReadFrame to avoid the call to GetRxBufferStatus, same thing done in ISR: https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/mlrs-tx.cpp#L322
- Sx that is not transmitting is already in idle before a transmit, for the Tx side this is done in doPreTransmit: https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/mlrs-tx.cpp#L945-L946 and for the Rx side this is done in doPostReceive: https://github.com/olliw42/mLRS/blob/main/mLRS/CommonRx/mlrs-rx.cpp#L839-L840

### DIFF
--- a/mLRS/Common/common.h
+++ b/mLRS/Common/common.h
@@ -233,9 +233,9 @@ tLEDs leds;
 void sxReadFrame(uint8_t antenna, void* const data, void* const data2, uint8_t len)
 {
     if (antenna == ANTENNA_1) {
-        sx.ReadFrame((uint8_t*)data, len);
+        sx.ReadBuffer(0, (uint8_t*)data, len);
     } else {
-        sx2.ReadFrame((uint8_t*)data2, len);
+        sx2.ReadBuffer(0, (uint8_t*)data2, len);
     }
 }
 
@@ -245,10 +245,8 @@ void sxSendFrame(uint8_t antenna, void* const data, uint8_t len, uint16_t tmo_ms
 #if !defined DEVICE_HAS_DUAL_SX126x_SX128x && !defined DEVICE_HAS_DUAL_SX126x_SX126x // SINGLE BAND
     if (antenna == ANTENNA_1) {
         sx.SendFrame((uint8_t*)data, len, tmo_ms);
-        sx2.SetToIdle();
     } else {
         sx2.SendFrame((uint8_t*)data, len, tmo_ms);
-        sx.SetToIdle();
     }
 #else
     sx.SendFrame((uint8_t*)data, len, tmo_ms);
--- a/mLRS/Common/sx-drivers/sx12xx_driver.h
+++ b/mLRS/Common/sx-drivers/sx12xx_driver.h
@@ -127,6 +127,7 @@ class SxDriverDummy
     void GetPacketStatus(int8_t* const RssiSync, int8_t* const Snr) {}
     void SendFrame(uint8_t* const data, uint8_t len, uint16_t tmo_ms) {}
     void ReadFrame(uint8_t* const data, uint8_t len) {}
+    void ReadBuffer(uint8_t offset, uint8_t* const data, uint8_t len) {}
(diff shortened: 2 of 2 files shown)


olliw42:
- ReadBuffer used instead of ReadFrame

That's indeed an interesting one. I'm actually not sure why I did use ReadBuffer(), I mean it's not totally clear why it actually works  ... I'm sure the ReadFrame follows the datasheet more closely ... interesting, but I think we should carefully investigate. Could be also sx chip dependent.

- Sx that is not transmitting is already in idle before a transmit, for the Tx side this is done in doPreTransmit

I'm extremely hesitant to remove Idles ... I recall that iot went through a number of iterations before it didn't ran into errors even in the edge cases (FRM303 e.g. was such a case which was difficult to get right). What we could do is to store the current mode in a variable and not do the spi if it is in the mode already.

olliw42:
reopened it
I think we may want to look into the ReadBuffer thing. If we can use that generally it would be useful.

jlpoltrack:
> I think we may want to look at least into the ReadBuffer thing. If we can use that generally it would be useful.

One can get away with using ReadBuffer as the baseAddress is set to 0 for Rx during initialization here, e.g. for SX128x: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/sx-drivers/sx128x_driver.h#L209. The other note is that ReadFrame also returns the payload length from the BufferStatus command but as mLRS uses a fixed payload length and this length is also passed into either the ReadBuffer and ReadFrame functions am not sure of the purpose.

> on the longer run we also may want to critically review the sx handling, e.g. as regards the Idle. I e.g. always stumple over this line https://github.com/olliw42/mLRS/blob/main/mLRS/CommonRx/mlrs-rx.cpp#L837-L841, as I don't understand it anymore. It's really done empirically and not by deep analysis and understanding.

I would guess these lines are perhaps for SX127x which doesn't seem to have an AutoFS / Fallback mode after a Tx or Rx has been completed...

olliw42:
I was thinking to try to simplify the sx readFrame call call as you suggest here and not use GetRxBufferStatus(&rxPayloadLength, &rxStartBufferPointer);
However, for the LR11xx there is no SetBufferBaseAddress(0, 0); as there is for the sx128x/6x/7x ... not sure if it thus would work reliably ???

jlpoltrack:
> However, for the LR11xx there is no SetBufferBaseAddress(0, 0);

Curiously, the LR1121 User Manual says this command exists but doesn't provide a corresponding op code for how to call it.  That being said, the defaults at 0x00 anyways.  I've been running this for long time on the bench with my LR1121 hardware, so no issues with it working.

image

> question is also if we actually want to do. I mean, it looks like a very good idea for sure, and is for now, but if we would look into going to header Lora mode we will need to call it anyway.

It's not strictly needed and this can all be revisited later IMO as it is not a critical thing, so okay if it makes sense to close.

olliw42:
> Curiously, the LR1121 User Manual says this command exists but doesn't provide a corresponding op code for how to call it. That being said, the defaults at 0x00 anyways. I've been running this for long time on the bench with my LR1121 hardware, so no issues with it working.

I understand it might be the default, but I think we either want to add such a function and call it or leave a comment in the driver that it should be set at startup but isn't as there is no docs for it and seems not to matter ...

the "problem" is that we actually should change the ISR to call ReadFrame, to have that in and being tested/used to ensure that we do not get unexpected side effects from here when we would change to header mode ... this mixed state situation currently is "unfortunate" ... I'm really not sure what the best path here is, which makes me kind to be in don't-do-aynthing mood LOL

jlpoltrack:
> I'm really not sure what the best path here is, which makes me kind to be in don't-do-aynthing mood LOL

Yea, no problem.  I will close.

PR #284, opened 2025-02-08
lua, enable internal modules on bw radios

tmcadam:
This change allows internal modules to be used on the RM Boxer, or other B&W radios that have ELRS internal modules (?). It replicates the color lua functionality in this regard.

### DIFF
--- a/lua/mLRS-bw.lua
+++ b/lua/mLRS-bw.lua
@@ -567,7 +567,7 @@ local function doParmLoop()
         end
         cmd = nil
     end --for
-    
+
     if DEV_ITEM_TX ~= nil and DEV_ITEM_RX ~= nil and connected and warned == 0 and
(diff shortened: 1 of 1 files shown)


olliw42:
MANY thx
we would need however also the compiled .luac. Please add it to this PR (you can get it form your radio sd card). I can't generate it since I don't have a bw radio :)

tmcadam:
I wasn't sure if these would compile the same for all bw radios. If they didn't it would be better to leave as is.

olliw42:
@brad112358 as regards the .luac, any bw radio produces the same?

brad112358:
> @brad112358 as regards the .luac, any bw radio produces the same?

Yes, I believe it was Risto who reassured us that the compiled form would be portable to other radios with the same processor architecture.

PR #285, opened 2025-02-08
ESP32 Rx Clock on Core 0

jlpoltrack:
Interrupts on ESP32 take something like 1-2 us, so a 10 us interrupt (100 KHz) means something like 100-200ms per second was being used to service this interrupt.

Previously FSK on DBR4 / XR4 would show bad link drops when running 115200 baud.  Now the LQ is rock solid at 100% when running FSK with 230400 baud.  ESP32C3 can use the same code even though it is single core.

Tested on:
- RadioMaster XR4
- RadioMaster XR1
- BetaFPV SuperD

### DIFF
--- a/mLRS/Common/hal/esp-rxclock.h
+++ b/mLRS/Common/hal/esp-rxclock.h
@@ -28,8 +28,7 @@
 #endif
 
 
-volatile bool doPostReceive = false;
-volatile bool doPostReceiveESP32 = false;
+volatile bool doPostReceive;
 
 uint16_t CLOCK_PERIOD_10US; // does not change while isr is enabled, so no need for volatile
 
@@ -65,7 +64,7 @@ void CLOCK10US_IRQHandler(void)
 
     // this is 1 ms after RX was or was supposed to be received
     if (CNT_10us == CCR3) {
-        doPostReceiveESP32 = true;
+        doPostReceive = true;
     }
 
     taskEXIT_CRITICAL_ISR(&esp32_spinlock);
@@ -116,6 +115,7 @@ class tRxClock
 void tRxClock::Init(uint16_t period_ms)
 {
     CLOCK_PERIOD_10US = period_ms * 100; // frame rate in units of 10us
+    doPostReceive = false;
 
     CNT_10us = 0;
     CCR1 = CLOCK_PERIOD_10US;
@@ -174,10 +174,10 @@ IRAM_ATTR void tRxClock::Reset(void)
 #endif
     CCR1 = CNT_10us + CLOCK_PERIOD_10US;
     CCR3 = CNT_10us + CLOCK_SHIFT_10US;
+    MS_C = CNT_10us + CLOCK_CNT_1MS;  // MS_C only used on ESP8266
 #ifdef ESP32
     taskEXIT_CRITICAL(&esp32_spinlock);
 #elif defined ESP8266
-    MS_C = CNT_10us + CLOCK_CNT_1MS;  // MS_C only used on ESP8266
     interrupts();
 #endif
 }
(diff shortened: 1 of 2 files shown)


jlpoltrack:
Updated performance numbers with this fix: https://github.com/olliw42/mLRS/pull/291

Baseline loop metrics @ 1 Hz shows 44 kHz:

image

With the fix + this branch, shows 77 kHz:

image

mLRS/CommonRx/mlrs-rx.cpp
```diff
@@ -723,6 +723,15 @@ IF_SX2(
     }//end of if(irq2_status)
 );
 
+#ifdef ESP32
```
olliw42, line 726: I don't like this
at minimum you put this into a fucntion which is called here
I'm not convinced however such a thing is really needed, it could be worthwhile to think about how this could be avoided

jlpoltrack: > I don't like this

That's fair - would a function like rxclock.CheckDoPostReceive be an acceptable name?  Would update (STM) rxclock.h as well to have this.

jlpoltrack:
I looked into this some more and I think an option is to simply not have this critical section in the rx loop (so mlrs-rx.cpp would be unmodified).

- On ESP32, a single uint32_t read / write is atomic.  So if Core 0 is setting doPostReceive to false at the exact time that Core 1 is reading doPostReceive there won't be a crash / halt or corruption.  Only a chance that Core 1 will read the stale value and will have to go through the loop one more time.
- Further, once Core 0 sets doPostReceive to true, it won't be modifying the value again any time soon (relatively) and therefore Core 1 is safe to do the read / modify / write on doPostReceive.

References: 
- https://esp32.com/viewtopic.php?t=18771#p74667
- https://esp32.com/viewtopic.php?t=18771#p69801

I have this running on the bench for a few hours and haven't noticed anything odd. Bonus is that loop rate improves 10%.

mLRS/Common/hal/esp-rxclock.h
```diff
@@ -19,7 +28,8 @@
 #endif
 
 
-volatile bool doPostReceive;
+volatile bool doPostReceive = false;
```
olliw42, line 31: no, we don't initialize like that, we generally and systematically do it by a Init() function :)
so pl remove the inits here and put them into the Init()

mLRS/Common/hal/esp-rxclock.h
```diff
@@ -119,10 +174,10 @@ IRAM_ATTR void tRxClock::Reset(void)
 #endif
     CCR1 = CNT_10us + CLOCK_PERIOD_10US;
     CCR3 = CNT_10us + CLOCK_SHIFT_10US;
-    MS_C = CNT_10us + CLOCK_CNT_1MS;
 #ifdef ESP32
     taskEXIT_CRITICAL(&esp32_spinlock);
 #elif defined ESP8266
+    MS_C = CNT_10us + CLOCK_CNT_1MS;  // MS_C only used on ESP8266
```
olliw42, line 180: I would prefer to set it as it was before
if one wants to outcomment to safe one shiould IMHO do it systematically, i.e., everywhere. Like so it at first looks like a bug. Outcommenting everythere saves really just a tiny bit, but makes teh code more clunky. Hence the preference.

mLRS/Common/hal/esp-rxclock.h
olliw42: to give you just one reasons: when we restart, these initis are not done anymore, but we want to avoid false states

jlpoltrack: Reverted

mLRS/Common/hal/esp-rxclock.h
jlpoltrack: Reverted but left the comment

olliw42:
very nice! THX!

PR #287, opened 2025-02-08
lua, update luac for bw lua script

tmcadam:
File taken from RM Boxer


olliw42:
@brad112358 you think this is ok?

tmcadam:
I have no idea tbh. But the PR is here for when we decide. I'll convert to draft to avoid confusion.

tmcadam:
Convert to draft, to avoid confusion. I am not sure this okay or not. It's here if we need it.

brad112358:
> @brad112358 you think this is ok?

I can't think of any reason it wouldn't be.

olliw42:
THX!

PR #292, opened 2025-02-12
Bandit, In Functionality

jlpoltrack:
Title

### DIFF
--- a/mLRS/Common/esp-lib/esp-uart.h
+++ b/mLRS/Common/esp-lib/esp-uart.h
@@ -196,6 +196,11 @@ void uart_init(void)
     // isr is enabled !
 }
 
+void uart_rx_enableisr(FunctionalState flag)
+{
+    // not supported on ESP, allows in functionality
+}
+
 
 //-------------------------------------------------------
 // System bootloader
--- a/mLRS/Common/esp-lib/esp-uart.h
+++ b/mLRS/Common/esp-lib/esp-uart.h
@@ -198,7 +198,7 @@ void uart_init(void)
 
 void uart_rx_enableisr(FunctionalState flag)
 {
-    // not supported on ESP, allows in functionality
+    // not supported on ESP, allows in functionality without lots of ifdefs
 }
 
 
--- a/mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
@@ -130,9 +130,9 @@ void sx_dio_exti_isr_clearflag(void) {}
 
 void in_init_gpio(void) {}
 
-void in_set_normal(void) { gpio_matrix_in((gpio_num_t)UART_USE_RX_IO, U1RXD_IN_IDX, false); }
+void in_set_normal(void) { gpio_matrix_in((gpio_num_t)UARTE_USE_RX_IO, U1RXD_IN_IDX, false); }
 
-void in_set_inverted(void) { gpio_matrix_in((gpio_num_t)UART_USE_RX_IO, U1RXD_IN_IDX, true); }
+void in_set_inverted(void) { gpio_matrix_in((gpio_num_t)UARTE_USE_RX_IO, U1RXD_IN_IDX, true); }
 
 
 //-- Button
--- a/mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
@@ -25,6 +25,7 @@
 
 #define DEVICE_HAS_JRPIN5
 #define DEVICE_HAS_SERIAL_OR_COM // hold 5-way in down direction at boot to enable CLI
+#define DEVICE_HAS_IN
 #define DEVICE_HAS_NO_DEBUG
 #define DEVICE_HAS_SINGLE_LED_RGB
 #define DEVICE_HAS_I2C_DISPLAY_ROT180
@@ -63,6 +64,12 @@
 #define UART_USE_RX_IO            IO_P13
 #define UART_TXBUFSIZE            0  // TX FIFO = 128
 #define UART_RXBUFSIZE            0  // RX FIFO = 128 + 1
+
+#define UARTE_USE_SERIAL1 // in port, uses JRPin5
+#define UARTE_BAUD                 100000
+#define UARTE_USE_TX_IO            -1
+#define UARTE_USE_RX_IO            IO_P13
+#define UARTE_RXBUFSIZE            0 // RX FIFO = 128 + 1
 
 
 //-- SX1: SX12xx & SPI
@@ -117,6 +124,15 @@ void sx_dio_init_exti_isroff(void)
 }
 
 void sx_dio_exti_isr_clearflag(void) {}
+
+
+//-- In port
+
+void in_init_gpio(void) {}
+
+void in_set_normal(void) { gpio_matrix_in((gpio_num_t)UART_USE_RX_IO, U1RXD_IN_IDX, false); }
+
+void in_set_inverted(void) { gpio_matrix_in((gpio_num_t)UART_USE_RX_IO, U1RXD_IN_IDX, true); }
 
 
 //-- Button
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-series-900-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-series-900-esp32.h
@@ -26,6 +26,8 @@
 #define DEVICE_HAS_SERIAL_OR_COM // hold 5-way in down direction at boot to enable CLI
 #define DEVICE_HAS_IN
 #define DEVICE_HAS_NO_DEBUG
+//#define DEVICE_HAS_NO_SERIAL
+//#define DEVICE_HAS_NO_COM
 
 #ifdef TX_ELRS_RADIOMASTER_BANDIT_900_ESP32 // Bandit, "big" Bandit
 // Bandit, "big" Bandit have RGB LEDs, so we use our usual red/green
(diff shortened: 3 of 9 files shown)


mLRS/Common/esp-lib/esp-uart.h
```diff
@@ -196,6 +196,11 @@ void uart_init(void)
     // isr is enabled !
 }
 
+void uart_rx_enableisr(FunctionalState flag)
```
olliw42, line 199: you need to run creatig from jtemplate again, seems you changed it later on :)

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-series-900-esp32.h
```diff
@@ -24,8 +24,7 @@
 
 #define DEVICE_HAS_JRPIN5
 #define DEVICE_HAS_SERIAL_OR_COM // hold 5-way in down direction at boot to enable CLI
-//#define DEVICE_HAS_NO_SERIAL
-//#define DEVICE_HAS_NO_COM
+#define DEVICE_HAS_IN
```
olliw42, line 27: I like to keep the outcommented pieces since they make it easier to enable debug

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-series-900-esp32.h
```diff
@@ -81,6 +80,12 @@
 #define UART_TXBUFSIZE            0 // TX FIFO = 128
 #define UART_RXBUFSIZE            0 // RX FIFO = 128 + 1
 
+#define UARTE_USE_SERIAL1 // JR bay pin5
```
olliw42, line 83: the comment here is probably left over from copying, shoudl be related to IN, pl copy from another hal

olliw42:
nice!
just few nitpicks
shouldn't/couldn't there be similar for betafpv1w?

mLRS/Common/esp-lib/esp-uart.h
jlpoltrack: Whoops, done

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-series-900-esp32.h
jlpoltrack: Put them back

mLRS/Common/hal/esp/tx-hal-radiomaster-bandit-series-900-esp32.h
jlpoltrack: Updated

jlpoltrack:
> shouldn't/couldn't there be similar for betafpv1w?

Don't think it makes much sense, but sure added anyways :)

olliw42:
why doesn't make it much sense?
i.e., why does it make sense for the bandit?

jlpoltrack:
> why doesn't make it much sense?

Personally would avoid a 2.4 + 2.4 relay for interference reasons.  2.4 + 900 makes a lot more sense IMO.

olliw42:
ah, so you think that's the main use. haven't thought that way. 
I guess folks will have other reasons too to want sbus. As it doesn't seem to hurt, I guess we rather provide it LOL

olliw42:
there is no setting needed saying that IN is on JRPIN5? i.e.  DEVICE_HAS_IN_ON_JRPIN5_RX ?

I find that a bit confusing ...

jlpoltrack:
> there is no setting needed saying that IN is on JRPIN5? i.e. DEVICE_HAS_IN_ON_JRPIN5_RX ?

I looked at doing this but there is a lot of STM32 specific stuff there, e.g. https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/in_interface.h#L134-L146 so just using 'plain' IN made more sense.

olliw42:
ah, ok, now I see what you're doing
you're replacing the set_normal/set_inverted by your own ...
hmm ...
that's kind of a completely new hal scheme ....
hmm ... need to think/look more carefully ...

jlpoltrack:
> that's kind of a completely new hal scheme ....

I think this has been done for long time, can see remnants in the E5 Mini Hal: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/stm32/tx-hal-WioE5-Mini-wle5jc.h#L166-L190

Due to nature of ESP GPIO, the pin used is neither a UART TX nor RX so perhaps a new define like 'DEVICE_HAS_IN_ON_JRPIN5 would be an alternative.

olliw42:
it seems to me we have a bit of a mess here wth the stm32s ...
looks like the functions in the hal are not used with in on jrpin5 ...
something is really odd here

olliw42:
ok, it seems I managed to confuse my self, all is good. The code pieces I found confusing are actually outcommented and only there if IN should be used. The could be also go I guess.

I think the difference between STM32 and ESP32 in_on_jrpin5 is that for stm32 we need to link in the driver hanlders, which is not something we use with esp32, so to be clean we would need an extra implementation for in_on_jrpin5 for the esp32 ...
In some sense I would prefer that since we do have already low-level stuff in the in_interface.h, and would free the hals, but maybe not needed now

mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
```diff
@@ -64,6 +65,12 @@
 #define UART_TXBUFSIZE            0  // TX FIFO = 128
 #define UART_RXBUFSIZE            0  // RX FIFO = 128 + 1
 
+#define UARTE_USE_SERIAL1 // in port, uses JRPin5
```
olliw42, line 68: why are these needed, given below you use UART_xxx and not UARTE_xxx ??

mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
```diff
@@ -119,6 +126,15 @@ void sx_dio_init_exti_isroff(void)
 void sx_dio_exti_isr_clearflag(void) {}
 
 
+//-- In port
+
+void in_init_gpio(void) {}
+
+void in_set_normal(void) { gpio_matrix_in((gpio_num_t)UART_USE_RX_IO, U1RXD_IN_IDX, false); }
```
olliw42, line 133: here you use UART_xxx and not UARTE_xxx, so is the above needed?

mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
jlpoltrack: Good catch, while they resolve to the same thing the IN functions should refer to UARTE_.

mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
jlpoltrack: Updated these to have UARTE.

olliw42: I'm sorry, I may not have been misleading
if no other info from the serial defines is needed, I rather would prefer the serial defines to go and UART_xxx be used here as you did
if however the above serial define section is needed, then it's better like so
:)

jlpoltrack: The UARTE_TX_IO and UARTE_RX_IO are needed to allow the serial to initialize properly - so should be kept as is.

olliw42: that's the part which confuses me ... so this code does initialize the same serial and the same pins twice?

jlpoltrack: It doesn't fully initialize the serial, just sets up the RX pin as input, where the signal should be routed internally. and whether or not it should be inverted.  It doesn't setup the baud rate, buffer, etc

olliw42: so these defines in the above are not needed?
so none of these defines are needed when UART_xxx is used in the set_xx() functions?

jlpoltrack: My thoughts :)

- Defines are needed to initialize the serial port.  The uarte_setprotocol function does this and needs both UARTE_USE_TX_IO and UARTE_USE_RX_IO.  If we try to initialize serial1 without providing pins there will a problem because (for some reason unknown to me) the default serial1 pins are connected to the ESP32 SPI flash and will probably result in a crash.
- gpio_matrix_in function is for the internal mux, which is used to set the inverted logic.  The pin, pin type and routing are already previously done by the serial begin but this seemed like the best way to set the inverted logic.

olliw42: so this serial is initialized twice, right, once as jrpin5, and once as in.
https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/in_interface.h#L43
https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/jr_pin5_interface_esp.h#L132

so
1. this doesn't make sense logically
2. what defines will win and determine the actual config of the uart depends on the order in which the two inits are called
3. since the defines are different, the uart will be configured incorrectly for one of them

if these points don't lead to malfunction, then only because of luck and coincidence, not because of proper code

I conclude we need to sort this out

jlpoltrack: > so this serial is initialized twice, right, once as jrpin5, and once as in.

Doesn't the 'Tx Ch Source' parameter choose one or the other? https://github.com/olliw42/mLRS/blob/fa196d76e9d35e6e6b4a9ab949fe6e75dd196bed/mLRS/Common/common_conf.h#L50 (maybe I misunderstand)

jlpoltrack: All good

olliw42:
MAN THX!

PR #294, opened 2025-02-15
PIO Tidy

jlpoltrack:
- pico32 and esp32dev are essentially the same define just named differently, so just have one.
- remove out of date comment

esp32dev json: https://github.com/platformio/platform-espressif32/blob/develop/boards/esp32dev.json
pico32 json: https://github.com/platformio/platform-espressif32/blob/develop/boards/pico32.json

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -144,7 +144,7 @@ build_flags =
 
 [env:rx-generic-900-td-pa]
 extends = env_common_rx_esp32
-board = pico32
+board = esp32dev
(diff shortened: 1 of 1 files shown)


olliw42:
hm. I see the argument but actually like pico32 better than esp32dev
1. it's clear what chip it actually is
2. esp32dev sounds like a dev version thing

so, I guess i woudl actually rather come t the opposite, namely to replace all esp32dev ...

PR #295, opened 2025-02-19
esp uart getbuf + jr pin read

jlpoltrack:
Add a getbuf function the ESP UART lib.

### DIFF
--- a/mLRS/CommonTx/jr_pin5_interface_esp.h
+++ b/mLRS/CommonTx/jr_pin5_interface_esp.h
@@ -91,9 +91,9 @@ void tPin5BridgeBase::Init(void)
 
     pin5_init();
 
-    // trigger the callback on a symbol timeout
-    // use a lambda, since the rx callback doesn't actually need the argument
-    UART_SERIAL_NO.onReceive([this]() { pin5_rx_callback(0); }, true);
+    // onReceive uses the pin5_rx_callback function
+    // true means trigger only on a symbol timeout
+    UART_SERIAL_NO.onReceive((void (*)(void)) uart_rx_callback_ptr, true);
     
 #ifndef JR_PIN5_FULL_DUPLEX
 
@@ -121,9 +121,9 @@ void tPin5BridgeBase::Init(void)
 //-------------------------------------------------------
 // Receive callback
 // callback is triggered once the radio has stopped transmitting
-// the buffer will contain complete CRSF message or potentially the end of a
-// message when the callback is first initialized
-// use a FIFO to play it safe
+// the buffer will contain a complete CRSF message or potentially 
+// the end of a message when the callback is first initialized
+// use a fifo to play it safe
 
 void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
 {
@@ -135,7 +135,7 @@ void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
     pin5_getbuf(buf, available);
     crsf_fifo.PutBuf(buf, available);
     
-    // parse for a crsf message
+    // parse for a CRSF message
     while (crsf_fifo.Available()) {
         if (state >= STATE_TRANSMIT_START) break; // read at most 1 message
         parse_nextchar(crsf_fifo.Get());
--- a/mLRS/CommonTx/jr_pin5_interface_esp.h
+++ b/mLRS/CommonTx/jr_pin5_interface_esp.h
@@ -92,7 +92,8 @@ void tPin5BridgeBase::Init(void)
     pin5_init();
 
     // trigger the callback on a symbol timeout
-    UART_SERIAL_NO.onReceive((void (*)(void)) uart_rx_callback_ptr, true);
+    // use a lambda, since the rx callback doesn't actually need the argument
+    UART_SERIAL_NO.onReceive([this]() { pin5_rx_callback(0); }, true);
     
 #ifndef JR_PIN5_FULL_DUPLEX
 
@@ -150,17 +151,20 @@ void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
 
 //-------------------------------------------------------
 // Pin5 Serial class
-// used for ESP passthrough flashing
+// used for ESP passthrough flashing, commented functions are unused
 
 class tJrPin5SerialPort : public tSerialBase
 {
   public:
+    // void Init(void) override { uart_init(); }
     void SetBaudRate(uint32_t baud) override { uart_setprotocol(baud, XUART_PARITY_NO, UART_STOPBIT_1); }
     bool full(void) { return !uart_tx_notfull(); }
     void putbuf(uint8_t* const buf, uint16_t len) override { uart_putbuf(buf, len); }
     bool available(void) override { return uart_rx_available(); }
     char getc(void) override { return uart_getc(); }
     void flush(void) override { uart_rx_flush(); uart_tx_flush(); }
+    // uint16_t bytes_available(void) override { return uart_rx_bytesavailable(); }
+    // bool has_systemboot(void) override { return uart_has_systemboot(); }
 };
 
 tJrPin5SerialPort jrpin5serial;
--- a/mLRS/CommonTx/jr_pin5_interface_esp.h
+++ b/mLRS/CommonTx/jr_pin5_interface_esp.h
@@ -150,8 +150,8 @@ void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
     uart_getbuf(buf, available);
     
     for (uint16_t i = 0; i < available; i++) {
-      if (state >= STATE_TRANSMIT_START) break; // read at most 1 message
-      parse_nextchar(buf[i]);
+        if (state >= STATE_TRANSMIT_START) break; // read at most 1 message
+        parse_nextchar(buf[i]);
     }
 
 
--- a/mLRS/CommonTx/jr_pin5_interface_esp.h
+++ b/mLRS/CommonTx/jr_pin5_interface_esp.h
@@ -139,11 +139,19 @@ void tPin5BridgeBase::pin5_tx_enable(bool enable_flag)
 }
 
 
-void tPin5BridgeBase::pin5_rx_callback(uint8_t c)
+void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
 {
     // poll uart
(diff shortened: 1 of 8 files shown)


olliw42:
it's not used anywhere ?!?

jlpoltrack:
Experimenting with stuff like this, so figured that I would include in the mean time:

image

jlpoltrack:
- Use getbuf in the JR rx callback as ESP prefers to use a buffer
- Add IRAM_ATTR to the function

Loop increases 9.5%, Max execution time decreases 13.5%

image

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -139,11 +139,19 @@ void tPin5BridgeBase::pin5_tx_enable(bool enable_flag)
...
 
-void tPin5BridgeBase::pin5_rx_callback(uint8_t c)
+void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
 {
     // poll uart
     while (uart_rx_available() && state != STATE_TRANSMIT_START) { // read at most 1 message
-        parse_nextchar(uart_getc());
+        uint16_t bytesAvailable = uart_rx_bytesavailable();
+        uint16_t i = 0;
+        char buffer[bytesAvailable];
```
olliw42, line 148: we don't do this in mLRS
I don't consider it a safe construct since it's stack usage is somewhat unpredictable
pl use a fixed size array, and it also needs to be checked that it can't be overflowed
.since this parses a crsf message, a buf size of 64 bytes, or maybe a bit larger, could be appropriate

olliw42:
interesting
I actually would have expected thi to have a significant effect at all

I don't like the code structure, so two comments

olliw42:
maybe something like (untested) this

```
    uint8_t buf[CRSF_FRAME_LEN_MAX + 16];
    uint16_t available = uart_rx_bytesavailable();
    if (available > CRSF_FRAME_LEN_MAX) available = CRSF_FRAME_LEN_MAX;   
    uart_getbuf(buf, available);
    for (uint16_t i = 0; i < available; i++) {
        if (state >= STATE_TRANSMIT_START) break; // read at most 1 message
        parse_nextchar(buf[i]);
    }
```

olliw42:
you have tested this works??

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -139,11 +139,19 @@ void tPin5BridgeBase::pin5_tx_enable(bool enable_flag)
 }
 
 
-void tPin5BridgeBase::pin5_rx_callback(uint8_t c)
+void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
 {
     // poll uart
     while (uart_rx_available() && state != STATE_TRANSMIT_START) { // read at most 1 message
-        parse_nextchar(uart_getc());
+        uint16_t bytesAvailable = uart_rx_bytesavailable();
```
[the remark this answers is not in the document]
jlpoltrack: Sure, addressed with latest.

mLRS/CommonTx/jr_pin5_interface_esp.h
jlpoltrack: Sure, addressed with latest.

jlpoltrack:
> you have tested this works??

It works but would like to do some more testing, so hold off on merging.

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -139,13 +140,21 @@ void tPin5BridgeBase::pin5_tx_enable(bool enable_flag)
...
-    while (uart_rx_available() && state != STATE_TRANSMIT_START) { // read at most 1 message
-        parse_nextchar(uart_getc());
+    char buf[CRSF_FRAME_LEN_MAX + 16];
+    uint16_t available = uart_rx_bytesavailable();
+    available = MIN(available, CRSF_FRAME_LEN_MAX);
+    
+    uart_getbuf(buf, available);
+    
+    for (uint16_t i = 0; i < available; i++) {
+      if (state >= STATE_TRANSMIT_START) break; // read at most 1 message
```
olliw42, line 153: nitpick
indend is 4 blanks, not two :D

jlpoltrack: Whoops...

olliw42:
lgtm for my side
tell when you consider it ready :)

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -139,13 +140,21 @@ void tPin5BridgeBase::pin5_tx_enable(bool enable_flag)
 }
 
 
-void tPin5BridgeBase::pin5_rx_callback(uint8_t c)
+void IRAM_ATTR tPin5BridgeBase::pin5_rx_callback(uint8_t c)
 {
     // poll uart
-    while (uart_rx_available() && state != STATE_TRANSMIT_START) { // read at most 1 message
-        parse_nextchar(uart_getc());
+    char buf[CRSF_FRAME_LEN_MAX + 16];
```
olliw42, line 146: I'm sorry since it'S due to me but this code doesn't work. It will lose the bytes which are comming after a change of state to TRANSMIT_START.
I think this issue may have also been in your previous code.

jlpoltrack: I think that's not an issue.  The callback is setup to trigger after the radio has stopped transmitting (64 bytes or 1 symbol timeout), so the buffer will only contain a complete message and can't contain the end of one message + start of another.  There is a chance for the buffer to contain a partial message when the callback is first initialized but this will get discarded / handled by parse_char when it can't find the address byte.

olliw42: ok, got you.
it's a marked difference as compared to the stm32 intended behavior though
I guess we want to document this aspect by some comment at some point

olliw42: a "simple" way out would be to read all bytes and put them into a fifo, such a class is available already and much used e.g. in mavlink code. The readout could be then again byte for byte.
Not sure if that wouldn't kill the performance benefit however.

jlpoltrack: Added the fifo and did a cleanup on the file.  Performance improvement is still there when using the FIFO.  @brad112358 could you take a look as well?  I did make one additional functional change which is to just have the callback trigger on a symbol timeout as opposed to a specific number of bytes and symbol timeout.  Tested on both internal and external modules.

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -154,36 +148,19 @@ void tPin5BridgeBase::pin5_rx_callback(uint8_t c)
...
 {
   public:
-    // void Init(void) override { uart_init(); }
     void SetBaudRate(uint32_t baud) override { uart_setprotocol(baud, XUART_PARITY_NO, UART_STOPBIT_1); }
     bool full(void) { return !uart_tx_notfull(); }
     void putbuf(uint8_t* const buf, uint16_t len) override { uart_putbuf(buf, len); }
     bool available(void) override { return uart_rx_available(); }
     char getc(void) override { return uart_getc(); }
     void flush(void) override { uart_rx_flush(); uart_tx_flush(); }
-    uint16_t bytes_available(void) override { return uart_rx_bytesavailable(); }
```
olliw42, line 185: are you sure?
without having looked up I'm pretty sure this function is used by the pt code ...

mLRS/CommonTx/jr_pin5_interface_esp.h
olliw42: looks good to me ...

mLRS/CommonTx/jr_pin5_interface_esp.h
olliw42: ok
I think I would define it nevertheless ... to have the api complete (I think this was also the reason the other two had been there but outcommented, to make it clear what is used)

jlpoltrack: Sounds good, I added a comment to clarify.

jlpoltrack:
Also replaced the conversion in the onReceive to use a lambda instead - seems cleaner IMO.

mLRS/CommonTx/jr_pin5_interface_esp.h
```diff
@@ -92,7 +92,8 @@ void tPin5BridgeBase::Init(void)
     pin5_init();
 
     // trigger the callback on a symbol timeout
-    UART_SERIAL_NO.onReceive((void (*)(void)) uart_rx_callback_ptr, true);
+    // use a lambda, since the rx callback doesn't actually need the argument
+    UART_SERIAL_NO.onReceive([this]() { pin5_rx_callback(0); }, true);
```
olliw42, line 96: is there any real world benefit of this?
this thing is really not my most prefered construct, which is maybe an understatement LOL
so, if it's just for the sake of it I rather would prefer it some old fashioned way
even the void*void type cast of before wasn't super nice, maybe one should define a proper function prototype in this lib?
if a proper type doesn't work well, ok, then let's go with that ...

olliw42: btw, with all that stuff here, am I be correct to assume that mBridge mode does not work anymore?

jlpoltrack: > is there any real world benefit of this?

Well the benefit is that you can see the function used in the callback without tracing back 'uart_rx_callback_ptr' .  That being said, sounds like there is a strong preference to keep as is so I will revert and just add a comment clarifying.

> even the void*void type cast of before wasn't super nice, maybe one should define a proper function prototype in this lib?

Well the reason that it isn't a function with no arguments is to play nice with the STM32 version (https://github.com/olliw42/mLRS/blob/main/mLRS/CommonTx/jr_pin5_interface.h#L328) otherwise you'd need to use defines in the crsf_interface_tx.h to account for the two different versions (uglier IMO).

> btw, with all that stuff here, am I be correct to assume that mBridge mode does not work anymore?

Without knowing much about how mBridge works or differs from CRSF, I'm not sure why any of the changes in this PR would impact it working.

olliw42: I would have to look more deeply again into the mbridge stuff, it just was that they needed different calbacks
the [this](){} construct I indeed do not like, it feels like going back to K&R typ C ...

olliw42:
lgtm
pl tell when ready for merge :)

jlpoltrack:
Good to merge

Issue #297, opened 2025-03-02
Bug: Channel Data broken on EdgeTx 2.11


jlpoltrack:
mLRS can no longer understand channel data over CRSF on EdgeTx 2.11 (tried RC2) whereas 2.10.6 works.  The Lua works fine but no channel data is sent to the receiver and switchable power also doesn't work.  Tested with F1 E28 Dual.

Guess is that somehow related to these PRs:

https://github.com/EdgeTX/edgetx/pull/5641
https://github.com/ExpressLRS/ExpressLRS/pull/3008

olliw42:
really? are they going nuts now?

can you in ad nutshell say what they do?

jlpoltrack:
There's now an optional byte in the RC Channels frame that ELRS uses for arming so you can get channel 5 back.  What I don't understand is that this should be backwards compatible and this extra byte should not be there unless you select it from the EdgeTx GUI.  The relevant code I think would be here: https://github.com/EdgeTX/edgetx/pull/5641/files#diff-b734ecd2cf7abf1d2157b23e4e59d166c2407b29afba5c4e493eb8e25324016f

On the scope, I checked the CRSF output and it was still 26 bytes so not 100% sure what is going on.

olliw42:
so, basically, they break the official CRSF spec here?

olliw42:
@rotorman Risto, any idea why this may break us?

olliw42:
a wild guess, maybe it's this thing https://github.com/EdgeTX/edgetx/blob/main/radio/src/telemetry/crossfire.cpp#L303-L306, i.e., that the fields are not properly reset/initialized if the tx module doesn't do device info.

jlpoltrack:
Okay, I have the scope hooked up with an mLRS module and it looks like the radio will just continually spam 0x28 'Parameter Ping Devices' https://github.com/tbs-fpv/tbs-crsf-spec/blob/main/crsf.md#0x28-parameter-ping-devices and not provide any channel data.

olliw42:
ah, so maybe it's not an RC thing at all but they have changed the behavior here ...

anyway, that's a ping to C8, i.e., the flight controller ... did edgetx not do it before?? I'm confused ...

jlpoltrack:
I think this PR: https://github.com/EdgeTX/edgetx/pull/5506

How I read the message:

- C8 is listed as 'Serial sync byte' in the CRSF spec
- 04 - Length
- 28 - Ping Devices
[...]

PR #298, opened 2025-03-03
tx, crsf, send device info upon ping device

olliw42:
attempt to account for https://github.com/olliw42/mLRS/issues/297

not tested, just did some code

@jlpoltrack maybe you find a chance to test this out :)

### DIFF
--- a/mLRS/Common/protocols/crsf_protocol.h
+++ b/mLRS/Common/protocols/crsf_protocol.h
@@ -259,6 +259,7 @@ typedef struct {
 
 
 /* 0x1C Link Statistics RX
+
   uint8_t rssi_db;        // RSSI (dBm * -1)
   uint8_t rssi_percent;   // RSSI in percent
   uint8_t link_quality;   // Package success rate / Link quality (%)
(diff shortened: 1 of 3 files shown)


jlpoltrack:
Not working at the moment.  This appears to send as a 'Broadcast Frame' but needs to be sent as a 'Extended Header Frame' with the addresses.

image

image

olliw42:
@jlpoltrack 
MANY thx for your PR :)
added it, and tidied it up
could you pl double check/test again?

It seems that EdgeTx cuts the name down to 15 chars max ... what do you see actually displayed now? 
how many chars? is there some version info?

olliw42:
I guess it would be nice if we also could get the bind button LOL

jlpoltrack:
> could you pl double check/test again?

Still working :)

> It seems that EdgeTx cuts the name down to 15 chars max ... what do you see actually displayed now?

In Settings > Version > Modules I see the string 'MLRS BetaFPV Mi V40.0.0'. For the version, the closet that I can get is to send '0x05030100' which results in V1.3.5' on the radio (but can't quite see an answer way to convert the VERSION to that).

olliw42:
thx
so, we robably want to not add the MLRS at the beginning of th ename, I guess
concerning the version, I was wondering if it is part of the name, or generated from the firmwareid field, looks like it's generated ...

jlpoltrack:
Well IMO probably doesn't need to be perfect, it's rather buried in the settings.  I had to look at the EdgeTx source to figure out where this would be...

olliw42:
@jlpoltrack 
added some code to - hopefully - do the version correctly. Could you pl test?

also did the serial_no ... EdgeTx seems to use it to identfy ELRS modules, so, we could use the same mechanism to - maybe at some point - make EdgeTx to identify mLRS, e.g., to also give us a bind button
https://github.com/EdgeTX/edgetx/blob/main/radio/src/telemetry/crossfire.cpp#L295-L296

jlpoltrack:
> added some code to - hopefully - do the version correctly.

Yea, looks good.

olliw42:
so, ready to merge?

jlpoltrack:
Yea, good to merge.

olliw42:
many thx, sir

PR #300, opened 2025-03-04
Device Info Destination Update

jlpoltrack:
Since the radio asked for the device information, should probably set the destination as the radio.

Tested on BetaFPV and E28 Dual with EdgeTx 2.11, no issues.


olliw42:
this field is ignored by EdgeTx, hence it works
I think I was following ELRS
not sure what's better ... hmhm

PR #301, opened 2025-03-05
mimic edgetx/elrs bind start

olliw42:
@jlpoltrack 
this might be something to test out
it makes mLRS tell EdgeTx it is ELRS, such that EdgeTx should show the bind button in the crsf setup page. Pressing the bind button should put the tx module into bind mode indeed

this is not for merge but to test if this is working, so we eventually could consider doing a change to EdgeTx to get a bind button also for mLRS. (mLRS does not have an unbind mechanism, so this would work differently)

### DIFF
--- a/mLRS/Common/protocols/crsf_protocol.h
+++ b/mLRS/Common/protocols/crsf_protocol.h
@@ -110,6 +110,8 @@ typedef enum {
 
 // 0x32.0x10 Crossfire command options
 typedef enum {
+    CRSF_COMMAND_SET_BIND_MODE          = 0x01, // command to enter bind mode
+    CRSF_COMMAND_CANCEL_BIND_MODE       = 0x02, // command to cancel bind mode
     CRSF_COMMAND_SET_MODEL_SELECTION    = 0x05, // command to select model/receiver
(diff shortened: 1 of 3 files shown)


jlpoltrack:
> it makes mLRS tell EdgeTx it is ELRS, such that EdgeTx should show the bind button in the crsf setup page.

Flashed BetaFPV Micro with this branch, don't see Bind button on the Model Setup > External RF Page.

jlpoltrack:
Would need to fake the version number as well - I see the check here is for version >= 3.4 https://github.com/EdgeTX/edgetx/blob/978532429289c5675bb396d68fae90d2faab6bfc/radio/src/gui/gui_common.cpp#L1262-L1265 to get the bind button to show up

olliw42:
great. THX.
the button should change to 'unbind' when pressed ... when you press it again the tx module should stop the bind mode ... does this work too?

jlpoltrack:
When I select the button to bind, the message on the button remains bind.  If I select the button again, then the module does a soft reboot.

olliw42:
yes, that's the behavior
so, I guess we can add the basic functionality, to be prepared for some addition to edgetx
THX!

PR #302, opened 2025-03-06
sx driver, bug, handle non initialized gconfig

olliw42:
this should account for the crash with diversity, https://github.com/olliw42/mLRS/issues/245

we had already two attempts to resolve, 
https://github.com/olliw42/mLRS/tree/owpr-rfpower-antenna
https://github.com/olliw42/mLRS/pull/266

however, I've determined that the underlying "true" issue is that gconfig in the sx drivers is not initialized due to these lines https://github.com/olliw42/mLRS/blob/main/mLRS/CommonRx/mlrs-rx.cpp#L549-L550, which make the Startup() may not be called and gconfig thus not be set. The result is an acces to an undefined location.
Interesting enough that the issue, which is clearly a bug, never came up before.

One could have changed code such that Startup() is always called but only the sx not initialzed. I've opted for allowing gconfig to be non-initzialized. Not sure which approach is better, bith seem to me to do the same at the end of the day.

### DIFF
--- a/mLRS/Common/sx-drivers/lr11xx_driver.h
+++ b/mLRS/Common/sx-drivers/lr11xx_driver.h
@@ -90,6 +90,7 @@ class Lr11xxDriverCommon : public Lr11xxDriverBase
 
     void Init(void)
     {
+        gconfig = nullptr;
         lora_configuration = nullptr;
         gfsk_configuration = nullptr;
     }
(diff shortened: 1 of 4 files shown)


jlpoltrack:
Test on F1 E28 Dual:

With Antenna1: No longer results in crash
With Antenna2: No longer results in crash, correct power and sensitivity reported

Looks good

olliw42:
THX

PR #306, opened 2025-03-11
tx,rx, transmit fhss index in ota frame

olliw42:
@jlpoltrack @brad112358 @tmcadam 

this is a first attempt at transmitting the fhss index in the ota frame. We had a discussion on this mechanism.

We need this
- to ensure that when we connect we do not connect to a false frame. This apperaed to happen especially on 1W systems, and in the 2.4GHz range, when in close range. This could lead to a "low LQ" connection wheras the vehicle is nearby. (= dangerous to fly)
- to ensure for dual band systems that when one connects buit has connection only to one band also the other band is set to the correct index so it can connect later on

The mechanism is to repurpose a hitherto unused 7 bit field to send in the Tx->Rx frame the fhss index, as well as an index for which band it is. We have backwards compatibility to the lucky fact that the respective field was set to UINT8_MAX before. 

It should fully work for single band, both single antenna and diversity and dual band. I have tested it for single band with single antenna and diversitywith a system in 900 and 2.4 GHz band, and it looks good to me.

Ergo, it could see critical review and testing :)


olliw42:
I've added what I think it needs for dual band.

I've also added some debug code which allows to mimic the situation that it connects first on one band, and when receives on the other band. It uses the 6-way radio switches on CH5 usually used for ArduPilot flight modes to set the cases.

To me it looks fine on the bench.

brad112358:
I admit I've only looked at a small part of this so far, so perhaps I'm just being daft, but this looks a bit different from the algorithms I remember from our discussion.  Can you describe how this works in a bit more detail?  Or, better, add a comment with an overview?

olliw42:
hey Brad
thx for also looking at it, very helpfull and very much needed
I'm a bit surpirsied by you not finding it close to the mehcanism we had discussed, to me it looks like I essentially did what you were suggesting. So I'm not sure what the point of confusion actually is and what to best say. So let me try something:

The basic idea is to transmit the frequency on which a packet is send along side with that packet.
 
In the case of a single band system (can be diversity) this is redundant information, but it is needed since the reciever may actually succesfully receive a packet on a different frequency than it is adjusted to (e.g. when the power is high). So the information in the packet allows the receiver to check if it has received that packet while being on that frequency, or not. In the latter case it wants to discard that packet. This avoids that it synchronizes incorrectly with the transmitter. We only need to do this in the synchronization phase (i.e. when the reciver wants to connect) since once it is correctly synchronized it will stay synchronized (or disconnects). In the single band case the mechanism really just helps with avoiding wrong connection in closer range/high power.

In the case of a dual band situation, the mehcanism is crucially required (call it a fundamental flaw in my dual band code). Let's imagine (for the moment) we are far distance, such that at this distance only the packets of one of the bands are received. So, teh receiver will see these packets, and synchronize with it's hoping sequence onto these packets. If the hopping sequences on both bands would be the same, the receiver could deterime from receiving the packets on one band what the frequency on teh other band would have to be, and all would be good. However, this is not the situation. Especially the 915 band is incommensurate to all other bands, since it's 25 hops can't be easily devided the other fhss lengths. So basically, the result would be that the receiver connects to the packets on one band, but cannot receiver the packets on the other bands. The advantage of dual band wiould be gone and we would fall back to single band.
So, what needs to happen is that when it receives a packet on one band, it needs to set the fhss for the other band to the correct frequency which the transmitter did use. 

Now, as we discussed, we do not want to send both frequencies in a packet becasue we would have to change the packet and would break backwards compatibility. So you had the suggestion to send these two frequencies alternately.

Does this help?

mLRS/Common/frames.h
```diff
@@ -428,4 +430,33 @@ tTxCmdFrameRxParams* rx_params = (tTxCmdFrameRxParams*)frame->payload;
...
+}
+
+
+uint8_t fhss_band_next(void)
+{
+    static uint8_t fhss_band = 0;
+    static uint8_t fhss_band_last = 0;
+
+    if (fhss_band == fhss_band_last) { // we had it two times, so toggle
+        fhss_band_last = fhss_band;
```
brad112358, line 452: If they are equal, why set fhss_band_last?

mLRS/CommonRx/mlrs-rx.cpp
```diff
@@ -774,18 +792,55 @@ dbg.puts(s8toBCD_s(stats.last_rssi2));*/
...
+dbg.puts(" set b1");
+                } else { // ok
+                    connect_fhss_index_band_seen |= 0x01; // 0 seen
+                }
+            } else {
+                if (fhss.GetCurrI2() != stats.received_fhss_index) { // somehow wrong frequency, so discard
+                    valid_frame_received = false;
+                    invalid_frame_received = true;
+dbg.puts(" discard");
+                    fhss.SetCurrI2(stats.received_fhss_index); // set band 2, should only happen for dual band
```
brad112358, line 819: I would prefer the comment to say "set second band index"...  ie received_fhs_index_band == 1 is second band, not band 2.  Same thing above, I would prefer comment say "first band" rather than "band 1" when received_fhs_index_band = 0.

mLRS/CommonRx/mlrs-rx.cpp
```diff
@@ -774,18 +792,55 @@ dbg.puts(s8toBCD_s(stats.last_rssi2));*/
...
                 connect_sync_cnt++;
                 uint8_t connect_sync_cnt_max = CONNECT_SYNC_CNT;
                 if (!connect_occured_once) {
                     connect_sync_cnt_max = Config.connect_sync_cnt_max;
                 }
+                if ((connect_sync_cnt >= connect_sync_cnt_max) && (connect_fhss_index_band_seen != 0x03)) {
+                    connect_sync_cnt = connect_sync_cnt_max - 1; // not yet
+dbg.puts(" delay");
+                }
                 if (connect_sync_cnt >= connect_sync_cnt_max) {
```
brad112358, line 844: I would suggest "if (connect_sync_cnt >= connect_sync_cnt_max && connect_fhss_index_band_seen == 0x03) {" and delete the 4 new lines above.

mLRS/Common/fhss.h
```diff
@@ -526,20 +525,24 @@ class tFhssBase
...
         bind_listen_i = 0;
     }
 
-    // only used by receiver, bool determines if it needs to switch back to LINK_STATE_RECEIVE
+    //-- only used by receiver
+
+    // Rx:
+    void SetCurrI(uint8_t i)
+    {
+        if (i == curr_i) return; // nothing to do
```
brad112358, line 540: Could just remove this line

mLRS/Common/fhss.h
```diff
@@ -742,7 +749,10 @@ class tFhss
 #endif
     }
 
-    // only used by tx cli
+    //-- Tx module only
+
+    uint8_t Cnt(void) { return fhss1stBand.Cnt(); }
+    uint8_t CurrI_4mBridge(void) { return fhss1stBand.CurrI_4mBridge(); }
```
brad112358, line 755: Should mBridge provide a way to specify which band we want stats on?

mLRS/Common/frames.h
```diff
@@ -14,6 +14,7 @@
 #include "frame_types.h"
 
 
+extern tGlobalConfig Config;
```
brad112358, line 17: I'm curious why we didn't need this before.

brad112358:
Other than a couple of trivial nits, this all looks great to me!  I think this is a major improvement and as you say, closes a big hole for dual band when recovering from connection loss at distance.

mLRS/Common/frames.h
olliw42: ups, that's true. THX.

mLRS/CommonRx/mlrs-rx.cpp
olliw42: THX. done.

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -853,7 +886,7 @@ INITCONTROLLER_END
         if (dt < 750) break;
         isInTimeGuard = false;
         rfpower.Update();
-        fhss.HopToNext();
+        //fhss.HopToNext();
```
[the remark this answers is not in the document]
olliw42: yes. Like to keep it until the end to remind me where it was before. Can't see a functional difference, but am not sure.

mLRS/Common/fhss.h
olliw42: true. THX.

mLRS/Common/fhss.h
olliw42: maybe at some point. Not now.

mLRS/Common/frames.h
olliw42: ha. These are not needed at all since these are .h files. I just started to add them even if nfc to make the dependencies more clear.

olliw42:
MANY thx for looking at it. I guess I can assume you didn't had a chance to test, right?

I will leave this PR open for a while before merginging, to give us more time to check and test. It's a bit of a critical one and I want to be sure it's ok.

olliw42:
> I would like to see something like your text above added as a comment somewhere in the code. I sometimes miss such an outline for other potentially non-obvious functionality in mLRS code as well, I tend to prefer more comments explaining the strategy/algorithm used than we usually have.

agree that it would help, disagree as it's painfull ... and not always clear where to put. For the relay feture I addeda theory_of_operation.h, to place some description in there. We could do this here too to add info on that mechanism.

olliw42:
I've added a theory_of_operation.h file with some description of the mechanism.

PR #307, opened 2025-03-14
tx,rx, 19 Hz 7x mode for sx127x

olliw42:
title says it

the sx1276 chip set is not compatible with the sx126x, which can confuse, and also in view of LR1121 support (which can do both) we should distinguish between the 19 Hz mode for sx126x and sx127x.

we had a discussion on this before and the suggestion was to call this new mode "19 Hz 7x"

I've tested it somewhat and am hopeful, but it would be great to see you guys also looking over it and testing it.
It should be possible to falsh teh devices back and forth without issues.

### DIFF
--- a/mLRS/Common/arq.h
+++ b/mLRS/Common/arq.h
@@ -205,6 +205,7 @@ void tTransmitArq::SetRetryCntAuto(int32_t _frame_cnt, uint8_t mode)
     case MODE_50HZ:
     case MODE_31HZ:
     case MODE_19HZ:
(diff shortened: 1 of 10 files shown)


jlpoltrack:
Flashed on top of R9M + R9MX, looks good.
- Connected successfully, Lua shows the new mode
- Binding works as well

olliw42:
it should work also on "mixed" flashes
and all sx128x,sx126x

olliw42:
merged with minor modifications

Issue #308, opened 2025-03-14
MAVLinkX data loss/corruption of GPS_RTCM_DATA messages (?)


festlv:
I noticed weird behaviour of mLRS telemetry link when trying to send RTK GPS corrections from GCS to UAV side.
When the rx link type is set to mavlinx, I could observe some data passed to GPS module, but it refused to apply the corrections and stayed in 3d fix mode.
After setting the link mode to transparent, the GPS accepted the corrections (without changing anything else in the setup) and went into RTK fix modes.

The only tip for reproduction I have is: the link transfers several GPS_RTCM_DATA messages, approx 2KiB/s. Number of messages is around 10/s, but they are sent in quick succession.

I don't think I'll have enough time to do data dumps but I figured out that I should at least let others who are using mlrs for this know.

olliw42:
@festlv 
many thx for the report.

It is a bit weired though since the gps_rtcm_data message is in the lib and should be handled. What you are saying seems to say
1. these messages are transmitted but with some loss of packets
2. they are getting somehow corrupted so the gps can't digest anymore

Point 1 could be possible but I don't see how this should be different to then using transparent mode.
Point 2 I just don't understand.

Could you collect some more data?, 
- did you try mavlink instead of mavlinkX? Would be good to understand if this is a mavlinkX issue or an issue of your mavlink handling per see.
- do you have means to log the output on the mLRS receiver, e.g. with a usb-ttl adapter any pymavlink code, to see in which way the GPS_RTCM_DATA message could be corrupted?

Could you share your mLRS settings band, mode, mLRS hardware, mLRS firmware version, as well as your ArduPilot firmware version?

festlv:
1. I actually realized that by setting the transparent mode, I'm losing telemetry on the radio, so I did try mavlink link mode- it still does not work.
2. I have a test script that transmits the RTCM messages (as I was suspecting missionplanner as well)- I'll probably use that and log the messages that are transmitted and messages that are received.


Test setup is: 915MHz, 31Hz rate, Mateksys hardware, v1.3.03 fw, Ardupilot v4.5

olliw42:
thx for the aditional info
as said it's quite wired it would not work, so info of what arrives on the receiver side would be great
if the test sycript you mention is a py script you could share it, we might try it

festlv:
After a bit of further testing, it appears that there is increased data loss with mode = mavlink (vs transparent).
[...]

PR #310, opened 2025-03-18
TX, MAVLink fix, USB tweak

jlpoltrack:
- Add break to serial in link out when not using router
- Increase rxbufsize on USB for serial, e.g. Matek w/ SiK

To address: https://github.com/olliw42/mLRS/issues/308

### DIFF
--- a/mLRS/modules/stm32-usb-device/usbd_conf.h
+++ b/mLRS/modules/stm32-usb-device/usbd_conf.h
@@ -33,7 +33,7 @@ extern "C" {
 #define USB_RXBUFSIZE       512
 #define USB_TXBUFSIZE       2048 // helps with cli
 #else
-#define USB_RXBUFSIZE       512
+#define USB_RXBUFSIZE       2048  // for serial
 #define USB_TXBUFSIZE       512
 #endif
 #define USBD_IRQ_PRIORITY   0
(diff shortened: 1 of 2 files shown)


mLRS/modules/stm32-usb-device/usbd_conf.h
```diff
@@ -33,7 +33,7 @@ extern "C" {
 #define USB_RXBUFSIZE       512
 #define USB_TXBUFSIZE       2048 // helps with cli
 #else
-#define USB_RXBUFSIZE       512
+#define USB_RXBUFSIZE       2048  // for serial
```
olliw42, line 36: I suspect that's not correct
the only tx with USB which is not G4 is the FRM303, and if I recall correctly it has dammed little RAM. 
Does the PR compile for all targets? I would expect at least some RAM warning from the compiler.

jlpoltrack: - G431 is not covered here: https://github.com/olliw42/mLRS/blob/main/mLRS/modules/stm32-usb-device/usbd_conf.h#L32
- USB library seems to double buffer on the Rx side? On G431 this change results in 3 kB more RAM usage, 23.4 > 26.4.  

With 2048:
image

With 512:
image

olliw42: ah, not in g431 ... so the correct approach shoud be to add this mcu to the list of defines in line 32, i.e. like in line 50

I would have to look into the usb lib to recall what's happening LOL

olliw42:
can you please run run_make_firmware.py -t !rx-
(we should do this sort of thing for each PR, one of the last e.g. had a lttle thing missing so it didn't compile for few targets)

olliw42:
FANTASTIC catch, btw !

jlpoltrack:
> can you please run run_make_firmware.py -t !rx-

Sure, all firmware compile fine.

olliw42:
MANY THX!!

PR #312, opened 2025-03-22
Radiomaster ranger

brad112358:
Add support for "big" Radiomaster Ranger.

@jlpoltrack, I used your fan code.  Thanks!

Also, I guess I still don't have a firm grasp on adapting the info from ELRS targets.  based on their power_values, I calculated that POWER_GAIN_DBM should be 30, but that resulted in measured peak power of over twice what it was set too.  33 seems to work perfectly, but can someone explain why?

### DIFF
--- a/mLRS/Common/hal/esp/esp-device_conf.h
+++ b/mLRS/Common/hal/esp/esp-device_conf.h
@@ -199,6 +199,13 @@
   #define FREQUENCY_BAND_915_MHZ_FCC
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_RANGER_2400_ESP32
+  #define DEVICE_NAME "RM Ranger 2.4G"
+  #define DEVICE_IS_TRANSMITTER
+  #define DEVICE_HAS_SX128x
+  #define FREQUENCY_BAND_2P4_GHZ
+#endif
+
 
 //-- ELRS Internal Tx Modules
 
--- a/mLRS/Common/hal/esp/esp-hal.h
+++ b/mLRS/Common/hal/esp/esp-hal.h
@@ -131,6 +131,10 @@
 #include "tx-hal-radiomaster-bandit-series-900-esp32.h"
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_RANGER_2400_ESP32
+#include "tx-hal-radiomaster-ranger-2400-esp32.h"
+#endif
+
 
 // -- DIY
 
--- a/mLRS/Common/hal/esp/tx-hal-radiomaster-ranger-2400-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-radiomaster-ranger-2400-esp32.h
@@ -1,0 +1,376 @@
+//*******************************************************
+// Copyright (c) MLRS project
+// GPL3
+// https://www.gnu.org/licenses/gpl-3.0.de.html
+//*******************************************************
+// hal
+//*******************************************************
+
+/*
+  Flashing ESP8285:
+  - change ser dest to serial2
+  - change ser baudrate to 115200
+  - put Tx module into FLASH_ESP mode from tools
+  - Board: Generic ESP8266 Module
+  - Upload Speed: 115200
+  - Reset Method: no dtr (aka ck)
+*/
+
+//-------------------------------------------------------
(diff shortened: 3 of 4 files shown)


jlpoltrack:
> POWER_GAIN_DBM should be 30

I also calculate this, one thought is that ELRS is usually generous with their power levels.

brad112358:
I also measured the ELRS power levels and they are only a little high and consistent with what I see with mLRS with POWER_GAIN_DBM set to 33.  When I set it to 30, I saw well over double the power I expected.

olliw42:
I would always go with your measurements :)

lgtm (but I can't test of course) @tmcadam wasn't it that you also have a big Ranger?

this also needs support in mLRSFlasher app, if similar to bandit then like so I guess: https://github.com/olliw42/mLRS-Flasher/blob/master/assets/mLRS_metadata.py#L127-L142

jlpoltrack:
> I also measured the ELRS power levels and they are only a little high and consistent with what I see with mLRS with POWER_GAIN_DBM set to 33. When I set it to 30, I saw well over double the power I expected.

Ah interesting, I kind of doubt it would have this much impact, but I see you don't have the define for the DC-DC converter which Ranger has: https://github.com/ExpressLRS/targets/blob/master/TX/Radiomaster%20Ranger.json#L11

Example: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/stm32/tx-hal-diy-e28dual-board02-f103cb.h#L126

brad112358:
> Example: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/stm32/tx-hal-diy-e28dual-board02-f103cb.h#L126

Thanks!  I'll try this.  Note that the reason I didn't include this is because I referenced mostly Bandit when I wrote the HAL file, and the ELRS json for Bandit also has the "radio_dcdc": true, but our HAL doesn't account for that so I missed it.  Should Bandit also be using the dcdc regulator?

brad112358:
Adding SX_USE_REGULATOR_MODE_DCDC reduces the measured peak output by only about 1-6% (more effect at the low end) which works out well with slightly more accurate output at all settings and slightly lower minimum as long as I also leave POWER_GAIN_DBM set to 33.

We are still missing something.  What does power_min, power_high and power_max do in ELRS?

jlpoltrack:
> We are still missing something. What does power_min, power_high and power_max do in ELRS?

This was the hint - so there are 6 power levels (min = 1, max = 6) but the `power_values` array has 7 entries.  They left the 2 W / 33 dBm entry in there but not accessible unless you changed the max to 7.  So now when you do the calculation you get a gain of 33 which us what you empirically found:

image

brad112358:
I suspected something like that, but I was too lazy to look to see what power_max was used for in the ELRS code. :)

Since we have about half the maximum duty cycle, shouldn't it be safe to expose the highest possible power?  I tried adding a 2000 mw entry to rfpower_list[], But I don't have an attenuator adapter and my RF meter specs are 1W max.  I compared the 1W and 2W settings with a small air gap between the Tx antenna and power meter antenna and it looks like the 2W setting is putting out about 1.5 W which I guess might be a useful gain.

jlpoltrack:
> shouldn't it be safe to expose the highest possible power?

1 Watt is the legal limit for RC in the US (so likely why this is 'hidden'), so IMO best to just leave as-is.

brad112358:
Then, as far as I can see, this should be ready to merge.  @tmcadam, did you happen to find time to test this?

olliw42:
> this also needs support in mLRSFlasher app, if similar to bandit then like so I guess: https://github.com/olliw42/mLRS-Flasher/blob/master/assets/mLRS_metadata.py#L127-L142

:)

tmcadam:
Bench tested this branch on my Ranger. Didn't see any issues, bound okay with solid LQ. Fan, LEDs, OLED, 5pos all looked good.

olliw42:
very cool
so, folks, I will merge this, and update the pre-release firmware folder ... this then needs to be done, https://github.com/olliw42/mLRS/pull/312#issuecomment-2746324738. I can't test, so someone of you needs to add this and test :)

olliw42:
many THX, folks.

brad112358:
Thanks!

PR #316, opened 2025-04-08
subsume mlrs-wireless-bridge-esp8266 into mlrs-wireless-bridge

brad112358:


esp/mlrs-wireless-bridge/mlrs-wireless-bridge-boards.h
```diff
@@ -108,8 +114,28 @@ GPIO15 = RTC_GPIO13
...
+    #undef USE_SERIAL_DBG1
+    #undef USE_SERIAL1_DBG
+    #undef USE_SERIAL2_DBG
+
+    #undef LED_IO
+    #define LED_IO 16
+    #define USE_LED
+
+    #undef GPIO0_IO
+    #define GPIO0_IO  0
```
brad112358, line 132: I suppose I should leave out GPIO0_ID here so it can be defined or not in the .ino file or create a second module without it.  Which do you prefer?

olliw42: I'm sorry but I had no time yet to look at it and without it's not clear what you exactly mean and what the issue at hand is.
If "second module" means yet another .h file, I guess I would less prefer that.

brad112358: No, module as in another esp8266 variant "board/module" section right here.

olliw42: ok, I guess I see your point. If I understand correctly, I would suggest this:
The generic case I think is
- not needed for now
- may work with the existing GENERIC
- and if not may either made to work with the existing GENERIC or one could add a esp8266 generic later
So, I guess my suggestion to simply just rename the current GENERIC_8266 to something like e.g. MODULE_ESP8266_INTERNAL_TX

olliw42: as a general comment I guess I would prefer to have, in e.g. the #if splices, the EP8266 to come as last, so rather somthing like #ifndef ESP8266.
I also don't quite comprehend yet why e.g. the inlcudes had to be moved up, if possible I would prefer them to stay where they are.
I also don't quite comprehend yet why the "#ifdef GPIO0_IO #define USE_AT_MODE #endif" piece did go away.

brad112358: The esp8266 is also configurable on many external ELRS Tx modules, so I think MODULE_ESP8266_ELRS_TX instead of MODULE_ESP8266_INTERNAL_TX

brad112358: When both the #ifdef and #else clauses are present, I really perfer ifdef over ifndef.  Let me know if you feel strongly about the order.

The reason I moved the #includes is because one of them is needed to define the IPAddress type.

The "#ifdef GPIO0_IO #define USE_AT_MODE #endif" piece is still there.

olliw42: MODULE_ESP8266_INTERNAL_TX is good

order, hm. would have liked to have esp32 dominance ... hmhm

> The reason I moved the #includes is because one of them is needed to define the IPAddress type.
which one woud it be which would be need newly now?

brad112358: I'm getting mixed signals from you here.  You gave thumbs up to MODULE_ESP8266_ELRS_TX and then said MODULE_ESP8266_INTERNAL_TX is good.  I'm confused enough on my own!

brad112358: > which one woud it be which would be need newly now?

ESP8266WiFi.h  It also just more common to include standard system headers before jumping into defining variables, etc.

brad112358: > order, hm. would have liked to have esp32 dominance ... hmhm

Try to think of it as esp32 getting the last word.  :)

olliw42: > I'm getting mixed signals from you here. You gave thumbs up to MODULE_ESP8266_ELRS_TX and then said MODULE_ESP8266_INTERNAL_TX is good. I'm confused enough on my own!

ohh ... sorry ... copied the worng ... MODULE_ESP8266_ELRS_TX  is good :)

olliw42: > Try to think of it as esp32 getting the last word. :)

haha hard to beat that argument :D so go ahead :)

olliw42: > ESP8266WiFi.h It also just more common to include standard system headers before jumping into defining variables, etc.

well, yes ... but not with Arduino, here we often can keep things to have it cleaner ...
so, again, is this needed or not?

brad112358: ESP8266WiFi.h needs to be included before we can use IPAddress.  I guess if you want, I can split up the includes and put most of them back where they were, but it still makes more sense to me this way.

olliw42: look at the old code ... it was (seemingly) the same situation !!
so, again, did you had to do it becasue you got a compile error, or did you do it presumptiously

brad112358: Compile error for esp8266.  That's why I said "needs".  I moved the rest of the includes as well to keep them all together.  Do you want them split up?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -257,6 +268,21 @@ void serialFlushRx(void)
 
 void setup_device_name(void)
 {
+#ifdef ESP8266
+    String MAC_buf;
+    MAC_buf = WiFi.macAddress();
+    // No point in any sort of hash here, the last bytes of the MAC address in hex is more useful
```
olliw42, line 274: I don't agree
pl do as for esp32
we want to keep it consistent and transparent

esp/mlrs-wireless-bridge/mlrs-wireless-bridge-boards.h
brad112358: The comment on the include also explained all of this:
#include <ESP8266WiFi.h> // need this before WiFi settings section

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -145,43 +172,27 @@ String bluetooth_device_name = ""; // "mLRS BT"; // Bluetooth device name, "" re
...
-#if (WIRELESS_PROTOCOL != 2) // WiFi TCP, UDP, BT
+#if (WIRELESS_PROTOCOL != 2) // not UDPCl
+
+#if (defined USE_AT_MODE || (WIRELESS_PROTOCOL == 3)) && ! defined ESP8266 // for AT commands we require BT be available except 8266
+  #define USE_WIRELESS_PROTOCOL_BLUETOOTH
+  #include <BluetoothSerial.h>
+#endif
 
 // WiFi TCP, UDP
+#ifdef ESP8266 // could eliminate this and use broadcast for first message
```
olliw42, line 191: pl remove the comment
I think it is not esp8266 specific, and it also is poiting into the future

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -194,7 +205,7 @@ WiFiClient client;
 BluetoothSerial SerialBT;
 #endif
 
-#elif (WIRELESS_PROTOCOL == 2) // WiFi UDPCl
+#else // (WIRELESS_PROTOCOL == 2) WiFi UDPCl
```
olliw42, line 208: why this change?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -395,7 +433,9 @@ void setup()
 
     // Serial 
     size_t rxbufsize = SERIAL.setRxBufferSize(2*1024); // must come before uart started, retuns 0 if it fails
+#ifndef ESP8266 // do we even need this on esp32?
```
olliw42, line 436: why this change?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: I didn't say the comment was esp8266 specific.  The comment is specific to the ifdef itself.  Using broadcast for the first message would remove the need for the ifdef. The last octet would be set to 255 for both esp32 and esp8266.  I intend do to that at some point, but I didn't want to do it in this PR.  The comment is to remind us that it should probably be broadcast.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: The way it was done for the esp32 is far from transparent and not at all ideal, but I'll work on making the esp8266 use the same math since you insist.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: Because the matching #if is "#if (WIRELESS_PROTOCOL != 2) // not UDPCl" and there is no point in #elif with the same condition tested again.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: Because esp8266 does not define serial.setTxBufferSize()

olliw42: so pl change the comment to // not supported by ESP8266

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
olliw42: yes, but it aligns better with all the other ifdefs
I would wish changes would have been made with the philosophie to change as little as possible to the esp32 code :)

esp/mlrs-wireless-bridge/mlrs-wireless-bridge-boards.h
olliw42: that's strange it is needed for the esp8266 and not the esp32 ... I guess I will need to analyse retrospectively what's really going on here
so leave it as it is, I will go over it again after merge

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
olliw42: is understood, but the placement of the comment alludes to it's esp8266 specific
so leave it, I will change or remove after merge

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: I see you have merged,  I'd suggest we seek an answer to the question I asked in this comment as google suggests that it values over the default of 128 may not actually do anything.  If it's not needed, we can remove one ifdef.   I hope you will find time to look at the other two much simpler PRs, when they are merged, I can work on my UDPCl ideas.  Thanks!

olliw42:
@brad112358 
I had to roll back this merge. 
I got an compile error for the ESP8266 saying that "Preferences.h" isn't available. 
Did a quick google search and on this quick I saw comments saying that's not existing for esp8266. I have no idea what's up but my merge was a bit premature.

olliw42:
is that correct that one should install in the Luibrary Manager the "Preferences by Volodymyr Shymanskyy" ?

brad112358:
> is that correct that one should install via the Library Manager the "Preferences by Volodymyr Shymanskyy" ?

Yes,sorry.  I should have mentioned that!

brad112358:
Just to confirm, this is all working for me.

PR #317, opened 2025-04-08
Add DEVICE_BRIDGE_HAS_BT and associated allowed_mask

brad112358:


olliw42:
I think we could/should consider to implement a mechanism which allows the main to get this information from the flashed wireless.
Maybe like additionally asking for a config byte. Or simply adding such a thing to the name request. Or something like this.

Along the line we also could consider adding the mechanism for handing the bindphrase to the esp. To speed up exchange we could consider having bulkier messages which carry more info than just one data item

olliw42:
in line 434 after  `SANITIZE(Tx[config_id].WifiProtocol, WIFI_PROTOCOL_NUM, WIFI_PROTOCOL_UDP, WIFI_PROTOCOL_UDP);`  a `TST_NOTALLOWED(Tx_XXXX_allowed_mask, Tx[config_id].XXXX, XXXX);` is missing (all those with an allowed_mask need that)

mLRS/Common/setup.h
```diff
@@ -153,6 +153,13 @@ void setup_configure_metadata(void)
     SetupMetaData.Tx_Buzzer_allowed_mask = 0; // not available, do not display
 #endif
 
+    // Tx Bridge WiFi Protocol : "TCP,UDP,BT"
+#ifdef DEVICE_BRIDGE_HAS_BT
```
olliw42, line 157: I'd like to suggest to rename this define to `DEVICE_HAS_ESP_WIFI_BRIDGE_ESP8266`
it would then have to be removed from the two stm32 tx hal files where you've added it, and be added to the respective esp tx hal files.
It then also needs to be added to the list of defines in hal.h for documentation.

otherwise, we could use a not "DEVICE_HAS_XXXXX" type define, like we do to specify the esp pins etc pp. Howverm, then it should be named starting with ESP_XXXX. Could be for instance ESP_IS_ESP8266. It when should be added not in the top however, but in the "esp" section down below, where also the ESP_ pin defines are.

brad112358:
I've pushed a new version to make the requested changes.  Thanks!

PR #320, opened 2025-04-11
ESP32 Faster Flash

jlpoltrack:
Use 80 MHz and QIO for flash instead of 40 MHz and DIO - flash should be ~4x faster on ESP32.  This is now the default in the Arduino IDE.

As ESP32 has big (32 kB) cache, this change doesn't have much of an impact on receiver perofrmance but it helps out quite a bit on the Tx side - loop rate is ~5% faster, max loop time and jitter decrease.

@brad112358 @tmcadam perhaps you could check to see if you can correlate these observations.

This will also require a small update to the Desktop App, will do that when/if this is closer to being merged.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -28,8 +28,8 @@ build_flags = -Os
 
 [env_common_esp32]
 platform = espressif32@6.6.0
-;board_build.f_flash = 80000000L
-;board_build.flash_mode = qio
+board_build.f_flash = 80000000L
+board_build.flash_mode = qio
 board_build.f_cpu = 240000000L
 build_type = release
 build_flags =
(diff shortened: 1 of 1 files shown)


olliw42:
I can understand that it makes flashing faster ... I can't understand why it should make the code faster ... but that's why I'm not an esp person LOL

having said this, I'm useless here LOL. @brad112358 @tmcadam test & confrimatiion from you guys would be much appreciated :)

olliw42:
> This will also require a small update to the Desktop App, will do that when/if this is closer to being merged. Corresponding branch for the Desktop App: https://github.com/jlpoltrack/mLRS-Flasher/tree/esp32-faster-flash

does this mean that we would break backwards compatibility ???

jlpoltrack:
> I can't understand why it should make the code faster

The firmware on ESP32 is stored on a SPI flash, this makes that connection faster.

> does this mean that we would break backwards compatibility

The bootloader.bin is what sets these flash modes, so it works with existing firmware binaries.

olliw42:
@jlpoltrack @tmcadam 
so, should this be merged?

brad112358:
I've been including this in some of my testing with no problem observed.  I can't think of any downside, other, than possibly a very tiny increase i power use.  Unless @jlpoltrack has since found a problem, I'd suggest merging this.

brad112358:
It seems like slightly faster is a good thing, especially for the esp32.  @jlpoltrack should this be merged?  Has anyone seen any problems with it?

PR #322, opened 2025-04-19
Extend and improve UDPCl and UDP bridge support

brad112358:
Make UDPCl protocol select-able via AT commands.
Include device_id in the ssid which UDPCl attempts to connect to. Pass bind phrase to bridge for use in UDPCl access point password. 
Start UDP and UDPCl in broadcast mode to remove need for bridge to guess groundstation IP First received message switches to unicast to avoid slow Arduino broadcast receive. Don't require groundstation to send first in UDPCl mode to allow use of default UDP connection on groundstation. 
Attempt to reconnect to AP if WiFi connection is lost.

Tested with RM Pocket internal ELRS Tx, M5 C3U, and M5 Pico.


olliw42:
fantastic
two comments
- you touched a number of the WIRELESSPROTOCOL defines. I' not sure you recal, the whole thing should work also with the AT_MODE not defined, and in that case one should be able to set the wireless protocol via that define in the top (and code pieces should be blocked out). On first glance the changes seem to break this
- it seems to me the PR has a number of what you think clean up pieces. This makes it hard to review. I would prefer you do the additions/changes needed in this PR, with as little changes as possible. Cleanups etc can come in a 2nd PR.

brad112358:
The only WIRELESS_PROTOCOL conditional code left after you added the AT command configuration was UDPCl  The rest was already run-time selection.  This now adds UDPCl to the list of run-time selections which does remove some ifdefs, but this is a trend that started a long time ago.
Other than a couple of trivial comment corrections, I don't believe there are any unnecessary cleanups here.

olliw42:
@brad112358 
I'm working through this PR, and have a fundamenrtal question as regards UDPCl. I know, you most likely have explained this already 100 times to me, yet probably again:

In the old code you have to set the network_ssid and network_password, and in my understanding this are the credentials of your home network (or whatever else).
I can't understand how this would work now since the network_ssid is replaced by the generic device name and also the password isn't that of your existing network. 
How would this ever work at my home ... I mean, it's really the last thing I would want to do to fool around in my home's settings ... ????

olliw42:
@brad112358 
I have merged this now, but have worked out some changes which I will push as follow up. 

All the setting, storing etc. pp should be working. I don't know if this UDPCl thing is working.
Please carefully check, and raise a follow up PR as needed to get it working

I have not inlcuded the "udp improvement" changes, this I suggest to do separate later.

PR #323, opened 2025-04-22
Bridge improvements for UDP and UDPCl

brad112358:
Start UDP and UDPCl in broadcast mode to remove need for bridge to guess ground station IP (Eliminates hard coded IP addresses). 
First received message switches to unicast to avoid slow Arduino broadcast receive. 
Don't require ground station to send first in UDPCl mode to allow use of default UDP connection on ground station. Attempt to reconnect to AP if WiFi connection is lost.

### DIFF
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
@@ -102,25 +102,30 @@ Troubleshooting:
 //**********************//
 //*** WiFi settings ***//
 
+#define USE_BROADCAST_FOR_UDP // Comment out if you want to use only the first WiFi client to connect
+
 // for TCP, UDP (only for these two)
 // ssid = "" results in a default name, like "mLRS-13427 AP UDP"
 // password = "" makes it an open AP
 String ssid = ""; // "mLRS AP"; // Wifi name
 String password = ""; // "thisisgreat"; // WiFi password
 
+// Our IP address in AP mode
 IPAddress ip(192, 168, 4, 55); // connect to this IP // MissionPlanner default is 127.0.0.1, so enter 192.168.4.55 in MP
 
 int port_tcp = 5760; // connect to this port per TCP // MissionPlanner default is 5760
 int port_udp = 14550; // connect to this port per UDP // MissionPlanner default is 14550
 
-// for UDPCl (only for UDPCl)
+// for UDPCl (only for UDPCl) Station mode
 // network_ssid = "" results in
 // - a default name, like "mLRS-13427 STA UDPCl"
 // - and a default password which includes your mLRS bindphrase, like "mLRS-mlrs.0"
 String network_ssid = ""; // name of your WiFi network
 String network_password = "****"; // password to access your WiFi network
 
-IPAddress ip_udpcl(192, 168, 0, 164); // connect to this IP // MissionPlanner default is 192.168.0.164
+// Uncomment and edit the next two lines only if your AP has no DHCP server
+//#define FORCED_CLIENT_ADDRESS
+//IPAddress ip_udpcl(192, 168, 0, 164); // Our static IP in STA mode // Configure MissionPlanner UDPCl with this address
 
 int port_udpcl = 14550; // connect to this port per UDPCL // MissionPlanner default is 14550
 
@@ -213,13 +218,17 @@ String bluetooth_device_name = ""; // "mLRS BT"; // Bluetooth device name
 // Internals
 //-------------------------------------------------------
 
-// WiFi TCP, UDP
+// WiFi UDP: remote host IP
+#ifdef USE_BROADCAST_FOR_UDP
+IPAddress ip_udp(ip[0], ip[1], ip[2], 255); // Use subnet broadcast until we know who we are talking too.
+#else // Assume first address we offer via DHCP is our peer;  Start with unicast
 #ifndef ESP8266
 IPAddress ip_udp(ip[0], ip[1], ip[2], ip[3]+1); // usually the client/MissionPlanner gets assigned +1
 #else
 // the ESP8266 requires different, appears to be a bug in the Arduino lib
 IPAddress ip_udp(ip[0], ip[1], ip[2], ip[3]+99); // the first DHCP client/MissionPlanner gets assigned +99
 #endif
+#endif // USE_BROADCAST_FOR_UDP
 // TCP, UDP, UDPCl 
 IPAddress ip_gateway(0, 0, 0, 0);
 IPAddress netmask(255, 255, 255, 0);
@@ -386,7 +395,9 @@ if (g_protocol == WIRELESS_PROTOCOL_UDPCl) {
     // STA mode
     WiFi.mode(WIFI_STA);
     WiFi.disconnect();
+#ifdef FORCED_CLIENT_ADDRESS // Normally undefined to use DHCP
     WiFi.config(ip_udpcl, ip_gateway, netmask);
+#endif
     WiFi.begin(device_name.c_str(), device_password.c_str());
 
     while (WiFi.status() != WL_CONNECTED) {
@@ -401,6 +412,8 @@ if (g_protocol == WIRELESS_PROTOCOL_UDPCl) {
     DBG_PRINTLN(WiFi.localIP());
 
     setup_wifipower();
+    ip_udp = WiFi.broadcastIP();  // For easy GS configuration:  Enabling UDP should be enough
+    udp.stop();
     udp.begin(port_udpcl);
 
 }
@@ -512,6 +525,10 @@ void loop()
         return;
     }
 
+    if ((g_protocol == WIRELESS_PROTOCOL_UDPCl) && (! is_connected) && (WiFi.status() != WL_CONNECTED)) {
+       setup_wifi(); // attempt to reconnect to AP if we got disconnected
+    }
+
 if (g_protocol == WIRELESS_PROTOCOL_TCP) {
 //-- WiFi TCP
 
@@ -555,8 +572,9 @@ if (g_protocol == WIRELESS_PROTOCOL_TCP) {
     }
 
 } else
-if (g_protocol == WIRELESS_PROTOCOL_UDP) {
-//-- WiFi UDP
+if (g_protocol == WIRELESS_PROTOCOL_UDP || g_protocol == WIRELESS_PROTOCOL_UDPCl) {
+//-- WiFi UDP (AP mode) or UDPCl (Station mode)
+// Note that we don't have to send (broadcast) first in UDPCl mode, but this way it works even with plain UDP selected on the ground station and we also save some code
 
     int packetSize = udp.parsePacket();
     if (packetSize) {
@@ -564,6 +582,8 @@ if (g_protocol == WIRELESS_PROTOCOL_UDP) {
         SERIAL.write(buf, len);
         is_connected = true;
         is_connected_tlast_ms = millis();
+        ip_udp = udp.remoteIP(); // Assume only one peer and switch to Unicast to avoid Aurdino performance issue
+        port_udp = udp.remotePort();
     }
 
     tnow_ms = millis(); // may not be relevant, but just update it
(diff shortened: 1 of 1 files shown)


brad112358:
I'll try here again to explain the use case for UDPCl (UDP Station mode) to connect to a device ID based SSID with a bind phrase based password.   Obviously, this is only a stop-gap solution for users who can't use AP mode UDP and can't or won't compile their own firmware with their network's SSID and password.  Some home routers do allow a second AP to be configured which would be usable with this, but as you pointed out, users are unlikely to find this convienent.  But, I expect the main use case to be a mobile device like my Android phone which refuses to route packets via their cellular radios when connected to the bridge in AP mode.   In the past, I have used Bluetooth with my phone, but, esp8266...  In this case, it is trivial to setup my phone's hotspot to serve this specific SSID and password.  In fact, I find this slightly more robust in use than Bluetooth, since when Bluetooth is disconnected, (for example, because I saved a LUA setting change just before taking off), I need to disconnect and reconnect the bluetooth connection in QGC on my phone to get the connection working again (causing parameter download to be repeated).  With station mode, the GCS connection is now transparently recovered.  Hopefully someday we can expand the mLRS configuration system to allow longer strings and then we can have a much better UDPCl pre-compiled solution.

brad112358:
Even if we end up adding Bluetooth support to the mLRS ESP32 without use of a second esp, I will probably use this pre-configured UDPCl protocol with the 900 ELRS modules.  Though, I might switch to Bluetooth for 2400 to reduce interference.

brad112358:
There is one detail I haven't made my mind up on here:  Should we keep the device id based SSID I have currently implemented or would a non-unique SSID like just "mLRS STA UDPCl" be better.  The later is obviously more convenient since the user doesn't need to determine the ID portion by checking with UDP or TCP mode first and if they have more than one Tx module, they don't need to keep changing their phone hotspot settings to match the Tx module they are currently using.  The obvious advantage of the current (device id based SSID) is no collision with other users, though the bind phrase could, at least prevent other users from connecting my mistake.  What do you guys think?

olliw42:
MANY thx!
Much cleaner ... yet I have to admit I still do have issues understanding several things in the code. Certainly my limitation. I also need to go through your nice descriptions few more times.
I'd like to suggest thus to first have a first round of trivial comments, before I drop my understanding questions. To not have too many loose ends.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -102,25 +102,30 @@ Troubleshooting:
 //**********************//
 //*** WiFi settings ***//
 
+#define USE_BROADCAST_FOR_UDP // Comment out if you want to use only the first WiFi client to connect
```
olliw42, line 105: I understand the intention of the commenst etc. pp in this Wifi section, but I would like to keep the comments in this section more practical, i.e., more for illiterate folks like me. The following comments are along this line.

Pl move this to down below. Could e.g. go like
```
// for UDP (only UDP)
// Comment out if you want to use only the first WiFi client to connect
#define USE_BROADCAST_FOR_UDP
```

I'm not sure I understand the comment. Is it meaning Comment out if you want to connect to the first DHCP client? (here I pick up language used below)

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -102,25 +102,30 @@ Troubleshooting:
...
+
 // for TCP, UDP (only for these two)
 // ssid = "" results in a default name, like "mLRS-13427 AP UDP"
 // password = "" makes it an open AP
 String ssid = ""; // "mLRS AP"; // Wifi name
 String password = ""; // "thisisgreat"; // WiFi password
 
+// Our IP address in AP mode
```
olliw42, line 113: pl drop. I have two reasons, it breaks up the line to the "for TCP, UDP" section and one can piece it together only if one konws that these are AP, and it's not relevant that it's in AP mode

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -102,25 +102,30 @@ Troubleshooting:
...
 // network_ssid = "" results in
 // - a default name, like "mLRS-13427 STA UDPCl"
 // - and a default password which includes your mLRS bindphrase, like "mLRS-mlrs.0"
 String network_ssid = ""; // name of your WiFi network
 String network_password = "****"; // password to access your WiFi network
 
-IPAddress ip_udpcl(192, 168, 0, 164); // connect to this IP // MissionPlanner default is 192.168.0.164
+// Uncomment and edit the next two lines only if your AP has no DHCP server
+//#define FORCED_CLIENT_ADDRESS
+//IPAddress ip_udpcl(192, 168, 0, 164); // Our static IP in STA mode // Configure MissionPlanner UDPCl with this address
```
olliw42, line 128: this doesn't have to be outcommented, right? (or does it give a compile warning?)
I don't like the change of the comment ... STA is somewhat superfluous, "Configure" sounds scary, and I miss that it's MP's default

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -102,25 +102,30 @@ Troubleshooting:
...
-// for UDPCl (only for UDPCl)
+// for UDPCl (only for UDPCl) Station mode
 // network_ssid = "" results in
 // - a default name, like "mLRS-13427 STA UDPCl"
 // - and a default password which includes your mLRS bindphrase, like "mLRS-mlrs.0"
 String network_ssid = ""; // name of your WiFi network
 String network_password = "****"; // password to access your WiFi network
 
-IPAddress ip_udpcl(192, 168, 0, 164); // connect to this IP // MissionPlanner default is 192.168.0.164
+// Uncomment and edit the next two lines only if your AP has no DHCP server
```
olliw42, line 126: maybe `// Uncomment this line if your access point/hot spot has no DHCP server (not very likely)`

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -386,7 +395,9 @@ if (g_protocol == WIRELESS_PROTOCOL_UDPCl) {
     // STA mode
     WiFi.mode(WIFI_STA);
     WiFi.disconnect();
+#ifdef FORCED_CLIENT_ADDRESS // Normally undefined to use DHCP
```
olliw42, line 398: I'd suggest to remove the comment here
and add a comment in the next line like // don't use DHCP but force to the specified IP

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -401,6 +412,8 @@ if (g_protocol == WIRELESS_PROTOCOL_UDPCl) {
     DBG_PRINTLN(WiFi.localIP());
 
     setup_wifipower();
+    ip_udp = WiFi.broadcastIP();  // For easy GS configuration:  Enabling UDP should be enough
```
olliw42, line 415: except that it promises to make something easy, I really don't understand the comment. What does this make easier than before?

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -512,6 +525,10 @@ void loop()
         return;
     }
 
+    if ((g_protocol == WIRELESS_PROTOCOL_UDPCl) && (! is_connected) && (WiFi.status() != WL_CONNECTED)) {
```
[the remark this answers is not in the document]
brad112358: Sure.  FYI, the reason I chose this location is that it is near the other call to setup_wifi() and someday we may find a reason to support TCP with WiFi Station mode.  But, I can move it for now.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -607,38 +627,6 @@ if (g_protocol == WIRELESS_PROTOCOL_BT) {
...
-
-    int packetSize = udp.parsePacket();
-    if (packetSize) {
-        int len = udp.read(buf, sizeof(buf));
-        SERIAL.write(buf, len);
-        is_connected = true;
-        is_connected_tlast_ms = millis();
-    }
-
-    if (!is_connected) {
```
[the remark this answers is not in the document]
brad112358: I improved usability of our UDPCl mode (which really should be called STA mode) by removing the requirement that the "connection" must be initiated by the ground station.  Now, by default, we UDP broadcast MAVLink messages we receive over the serial link even before we receive the first UDP message from the groundstation.  This allows use of this mode without UDPCl on the Mission Planner side.  This way, the user doesn't need to configure the ground station with our address.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
```diff
@@ -102,25 +102,30 @@ Troubleshooting:
...
 String password = ""; // "thisisgreat"; // WiFi password
 
+// Our IP address in AP mode
 IPAddress ip(192, 168, 4, 55); // connect to this IP // MissionPlanner default is 127.0.0.1, so enter 192.168.4.55 in MP
 
 int port_tcp = 5760; // connect to this port per TCP // MissionPlanner default is 5760
 int port_udp = 14550; // connect to this port per UDP // MissionPlanner default is 14550
 
-// for UDPCl (only for UDPCl)
+// for UDPCl (only for UDPCl) Station mode
```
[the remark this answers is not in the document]
brad112358: It's Station mode that's important here.  There is no special support required on the bridge side to use "UDPCl" mode on Mission Planner.  For that matter, UDPCl on Mission Planner is only needed if we don't send (broadcast or Unicast) before we receive the first message from the ground station.  IE, it is never needed on our end and the only useful aspect of our UDPCl "protocol" is the fact that it is the only way we have to specify the bridge should use station mode.  UDPCl in Mission Planner just means "send first to this address"  which is only required if the other end can't send first which would only happen in cases where an IP router is in the path and broadcast won't work.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: But, this is not MP's default!  MP's default is 128.0.0.1 (localhost).  This is some random address that you said you forgot where it came from.  Our UDPCl mode uses WiFi Station mode and we should get our address via DHCP and should NEVER need to use a statically configured address like this.  IE, this should never have been here in the overwhelming common case.

brad112358: I don't want to leave line 128 uncommented because that might lead the user to believe that this address is being used for something when, in fact, it's just some random IP address which serves no purpose in the normal case.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
brad112358: Before, it was required to setup a UDPCl connection in MP with our address specified (which one would have to obtain from or configure in their network DHCP server.  By allowing our end to send first and by using broadcast for the first message, we eliminate the need for the ground station to know our address ahead of time.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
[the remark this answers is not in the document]
olliw42: if ever, this should go then into te tcp section

olliw42:
>There is one detail I haven't made my mind up on here: Should we keep the device id based SSID I have currently implemented or would a non-unique SSID like just "mLRS STA UDPCl" be better. The later is obviously more convenient since the user doesn't need to determine the ID portion by checking with UDP or TCP mode first and if they have more than one Tx module, they don't need to keep changing their phone hotspot settings to match the Tx module they are currently using. The obvious advantage of the current (device id based SSID) is no collision with other users, though the bind phrase could, at least prevent other users from connecting my mistake. I'm leaning towards removing the device ID and using a non-unique SSID to make it slightly easier for users to setup. The feature probably won't be used enough that a collision is likely anyway.
What do you guys think?

my first impulse was, sure, easier is better, but on second thought I think having the same scheme as for tcp/udp is better and we should stick to it. It's just more consistent. I also note that with some effort we can make the number available in the CLI, Lua, and OLed, so the user can easily look up. This would alos benefitial for tcp/udp.

To this end I note that currently we hae 5 or 4 digits araound, and we would want to harmonize.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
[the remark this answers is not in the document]
olliw42: it is superfluous nevertheless, and of no benefit for those who know and confusing for those who don't.

esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
olliw42: > But, this is not MP's default! MP's default is 128.0.0.1 (localhost).

in fact, that's correct ... checked on a pristine system. That's confusing now. Anyway.

olliw42:
I want to move this forward
so I'll do as before, merge this, merge a follow up, you then can complain, and we then can see
the USE_BROADCAST_FOR_UDP is IMHO not correctly handled, so I did also fool around with that as it made sense to me, but it may not yet be the correct thing to do

PR #324, opened 2025-04-27
wirelessbridge, next attempt

olliw42:
the broadcast thing for UDPSTA is not yet clear
hence this PR for working iot out
starts with what would be my best current guess

the discussion points are the 
-  `WiFi.config(ip_udp, ip_gateway, netmask); // no WiFi.config() when we start in broadcast`
- `ip_udp = WiFi.broadcastIP(); // start with broadcast`
 
in the wifi_setup()
haven't touched the first (seems needs to go), did for the 2nd what I think it could be

@brad112358

### DIFF
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
@@ -219,16 +219,12 @@ String bluetooth_device_name = ""; // "mLRS BT"; // Bluetooth device name
 // TCP, UDP, UDPCl 
 IPAddress ip_gateway(0, 0, 0, 0);
 IPAddress netmask(255, 255, 255, 0);
-// UDP (variable is used by UDPSTA, but is set by itself)
-#ifdef WIFI_USE_BROADCAST_FOR_UDP
-IPAddress ip_udp(ip[0], ip[1], ip[2], 255); // use broadcast until we know who we are talking too
-#else // use unicast, assume first address offered by DHCP is our remote
+// UDP, UDPSTA (might be set or overwritten as needed by the user configuration)
 #ifndef ESP8266
 IPAddress ip_udp(ip[0], ip[1], ip[2], ip[3]+1); // usually the client/MissionPlanner gets assigned +1
 #else // the ESP8266 requires different, appears to be a bug in the Arduino lib
 IPAddress ip_udp(ip[0], ip[1], ip[2], ip[3]+99); // the first DHCP client/MissionPlanner gets assigned +99
 #endif
-#endif // WIFI_USE_BROADCAST_FOR_UDP
 WiFiUDP udp;
 // TCP
 WiFiServer server(port_tcp);
@@ -411,8 +407,7 @@ if (g_protocol == WIRELESS_PROTOCOL_UDPSTA) {
 
     setup_wifipower();
 #ifdef WIFI_USE_BROADCAST_FOR_UDP
-    ip_udp = WiFi.broadcastIP(); // ?? a good comment would be nice
-    udp.stop();    
+    ip_udp = WiFi.broadcastIP(); // start with broadcast
 #endif    
     udp.begin(port_udp);
 
@@ -598,6 +593,7 @@ if (g_protocol == WIRELESS_PROTOCOL_UDP || g_protocol == WIRELESS_PROTOCOL_UDPST
 
     if (g_protocol == WIRELESS_PROTOCOL_UDPSTA){
         if (!is_connected && WiFi.status() != WL_CONNECTED) {
+            udp.stop();    
             setup_wifi(); // attempt to reconnect if we got disconnected
         }
     }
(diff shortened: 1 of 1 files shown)


brad112358:
>     * `WiFi.config(ip_udp, ip_gateway, netmask); // no WiFi.config() when we start in broadcast`

WiFi.config() has nothing to do with broadcast/unicast!   WiFi.config() sets our IP address (which must not be a broadcast address), not our peer's address which we send to (which can be broadcast or unicast in the case of UDP).

We must always call WiFi.softAPConfig() when we are a WiFi AP.  In addition to setting our ip address, it also sets the gateway, and netmask for the subnet we are creating.  By default, it also creates a DHCP server which will hand out IP addresses to stations which connect to our AP and inform them of the gateway and netmask.

We should not need or want to call the analogous WiFi.config() when we are a WiFi Station because that information will normally be provided to us by the AP's DHCP server.  We would only call WiFi.config() in this case if we want to ignore the DHCP configuration process because we need a fixed IP address for some reason or if the AP has no DHCP server.  If we call WiFi.config() we need to be sure we select an unused IP on the exact subnet which is being used on the AP and we must also know the correct netmask.  If we get either of these things wrong, we will not be able to communicate with any other nodes, not even with broadcast.  Therefore, we would never use WiFi.config() in STA mode in a generic pre-compiled binary, it only makes sense when a user is doing something quite unusual with custom compiled code.

Have you, perhaps confused the addresses of the two endpoints or perhaps the source and destination addresses?

Every node on an IP (Internet layer) subnet has it's own address.
Every TCP connection and every UDP message (Transport layer) is defined by a "4 tuple" which consists of the source IP address, the source port, the destination IP address, and the destination port.  TCP can't work with broadcast and UDP broadcast only applies to the destination IP address, never a source address.

olliw42:
what is `ip_udp = WiFi.broadcastIP();` actually returning? I.e., where it is deriving from what it is returning?

brad112358:
> what is ip_udp = WiFi.broadcastIP(); actually returning? I.e., where it is deriving from what it is returning?

The answer to this question might be very simple, but it depends on a lot of background understanding of IP networks which you have indicated you don't necessarily have and in that respect this might require a very long answer.  I'm tempted to try to find you a good link or book, but I'll take a crack at summarizing it myself.

I think I have mentioned in a few places in code comments and in github comments that when we connect to an WiFi AP in station mode that normally the DHCP protocol will be used to obtain our IP network configuration.  And, I know I've also mentioned that when we are in AP mode, softAPConfig() sets up a DHCP server to provide IP network configuration to the WiFi stations which connect to us.  And you wrote the code which provides the IP address and netmask to softAPConfig().  So, I'm not sure what you are missing.  Perhaps you want more details about how DHCP works?  Or, perhaps more details about how the subnet broadcast address is determined from an IP address and the netmask is all you need.  Or, perhaps you just don't have a good mental model of IP subnets and addressing.  

I'll start with a simple discussion of the addressing side just in case:

IPv4 addresses are 32 bit values, usually represented as 4 octets in "dot" notation.  The first n bits are the subnet number where the most common value of n is 16 or 24, but n can actually be any value between 1 and 30.  The subnet number or subnet address is usually written with n appended after "/", like 10.33.0.0/16 or 192.168.20.0/24 or 35.77.98.110/27.  Every IP address on a WiF/Ethernet network must have the same subnet number and must start with the subnet number and end with the host number or host part of the IP address.  Note that the boundary between subnet and host parts of an IP address is not always on an octet.  The host number of all 1 bits is reserved for broadcast, thus, the subnet broadcast address is the subnet number followed by all 1s.  The host number of all 0 bits is reserved.

When the destination address of an IP packet is the subnet broadcast Any IP address outside of the subnet is only reachable via a router.  Routers don't forward messages with a broadcast destination IP address.

Another way to specify the boundary n is with a subnet mask or network mask.  For example, a /16 subnet has a netmask of 255.255.0.0 and a /24 subnet has a netmask of 255.255.255.0.  A /27 subnet has a netmask of 255.255.255.112.

Since the broadcast IP address for a subnet is the subnet number followed by all 1s, it can be calculated by inverting the netmask and logical or-ing that with the subnet number (or any host IP address on the subnet).

Obviously, WiFi.broadcastIP() simply calculates and returns the broadcast address based on the subnet number or our IP address and netmask.

Our IP address and netmask are set by WiFi.softAPConfig() in the case where we are an Access Point or learned via DHCP (normal case) or set by WiFi.config() (very unusual with strictly manual configuration) in the case where we are connecting to an WiFi AP in station mode.

In summary, when we are an WiFi AP, we choose our IP address and netmask from which the broadcast address is calculated and provide this info to connecting stations via DHCP and when we are a WiFi station, we must use the subnet number and netmask of the network to which we connect to calculate the broadcast address and, again, we learn this via DHCP.

Do you want a summary of the DHCP protocol?

olliw42:
I understand that you do that with the best intentions and bets goodwill ... but it just triggers a "seriously?". I'm sorry.
maybe the answer is somewhere in this text but I fanly don't have the nerves to really read it
it was a simple question
so let me ask specifically, I would think that in both cases it simply replaces the last byte, but in the first case it would be doing this on localIP() and in the other case remoteIP() ... yes/no?

olliw42:
it seems some presumptions prevent useful progressing
this here will have to wait until I have time to properly resolve things myself, or, when back from your crusade push a despaghettized PR
I stop for now, and will unroll main so there is functioning code

PR #326, opened 2025-05-08
UDP STA mode and broadcast UDP

brad112358:
Third attempt

@olliw42 I've tried to stay as close as possible to what you had in your PR which you abandoned.  This means I have not attempted to improve UDPCL, even though I think at the very least, it should try to re-connect like UDP STA does when WiFi gets disconnected.  Hopefully you will decide to drop UDPCL someday so it won't matter.

### DIFF
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
@@ -94,9 +94,9 @@ Troubleshooting:
 #define WIRELESS_PROTOCOL  1
 
 // GPIO0 usage
-// comment out if your Tx module does not support the RESET and GPIO0 lines on the ESP32/ESP82xx (aka AT mode)
+// uncomment for pre-built binaries or if your Tx module supports the RESET and GPIO0 lines on the ESP32/ESP82xx (aka AT mode)
 // the number determines the IO pin, usally it is 0
-#define GPIO0_IO  0
+//#define GPIO0_IO  0
 
 
 //**********************//
@@ -220,7 +220,7 @@ String bluetooth_device_name = ""; // name of your Bluetooth device as it will b
 // TCP, UDP, UDPCl 
 IPAddress ip_gateway(0, 0, 0, 0);
 IPAddress netmask(255, 255, 255, 0);
-// UDP, UDPSTA (Will be overwritten if WIFI_USE_BRADCAST_FOR_UDP is set)
+// UDP, UDPSTA (Will be overwritten if WIFI_USE_BROADCAST_FOR_UDP is set)
 #ifndef ESP8266
 IPAddress ip_udp(ip[0], ip[1], ip[2], ip[3]+1); // usually the client/MissionPlanner gets assigned +1
 #else // the ESP8266 requires different, appears to be a bug in the Arduino lib
@@ -373,7 +373,9 @@ if (g_protocol == WIRELESS_PROTOCOL_TCP || g_protocol == WIRELESS_PROTOCOL_UDP) 
     } else
     if (g_protocol == WIRELESS_PROTOCOL_UDP) {
 #ifdef WIFI_USE_BROADCAST_FOR_UDP
-        ip_udp = WiFi.broadcastIP(); // Start with broadcast
+        //ip_udp = WiFi.broadcastIP(); // Looks like this is only for station mode
+        // Broadcast until we receive
+        ip_udp[3] = 255; // Our subnetmask is 255.255.255.0 so, just the last octet needs to change.
 #endif    
         udp.begin(port_udp);
     }
--- a/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
+++ b/esp/mlrs-wireless-bridge/mlrs-wireless-bridge.ino
@@ -7,7 +7,7 @@
 // Basic but effective & reliable transparent WiFi or Bluetooth <-> serial bridge.
 // Minimizes wireless traffic while respecting latency by better packeting algorithm.
 //*******************************************************
-// 1. Mai. 2025
+// 8. Mai. 2025
 //*********************************************************/
 // inspired by examples from Arduino
 // NOTES:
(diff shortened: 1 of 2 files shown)


olliw42:
great. thx.
seems to be mostly comments, I will comment on them more sometimes later
three quick things
- you suggest to have GPIO0_IO defined per default. I guess I see the reason but I think we want not. The script as is most likely will be used by DIY folks and they mostly will have just tx,rx, I guess. For our support via e.g. mLRSFlasher we would compile as we need anyway.
- WIFI_USE_BRADCAST_FOR_UDP ... this is a most funny missprint :):):)
- the relevant functional change seems to be in lines 605ff. Would have to look back (which I won't :)) but I think I copied what it was. I can see the reason for the change but makes me wonder why it wasn't before. Is this so to say a bug correction? Not suggesting a change, just want to understand.

I would not be surprised if UDPCl will eventually disappear ... if I understand correctly what you were saying it was a kind of fluke ... I just don't want to touch it, there is plenty of time to decide its fate lateron

olliw42:
I assume you have thoroughly tested the three modes, tcp, udp, udpsta, right

brad112358:
>     * you suggest to have GPIO0_IO defined per default. I guess I see the reason but I think we want not. The script as is most likely will be used by DIY folks and they mostly will have just tx,rx, I guess. For our support via e.g. mLRSFlasher we would compile as we need anyway. So, the idea I guess is to have it per default as simple as possible.

OK.  I'll change it back for now.  I seem to remember reading that you considered making configuration work without GPIO0.  Do you remember why you didn't?  Perhaps we should go ahead and do that.  It should be easy enough to enter AT command mode on the bridge at power on and then time out if we don't get a command in the first few seconds.  Of course, this would require a power cycle if the protocol/mode configuration is changed on the Tx module side after power up.

This way, even the DIY and R9 folks could benefit from a pre-compiled bridge binary for common bridge boards.
> 
>     * WIFI_USE_BRADCAST_FOR_UDP ... this is a most funny missprint :):):)

Indeed, one little "O" missed did make it petty funny :)

>     * the relevant functional change seems to be in lines 605ff. Would have to look back (which I won't :)) but I think I copied what it was. I can see the reason for the change but makes me wonder why it wasn't before. Is this so to say a bug correction? Not suggesting a change, just want to understand.

I assume you are refering to the added " if (is_connected != true) { ".  This only matters in the unusual case of having more than one groundstation on the WiFi network, each sending us UDP packets to the same standard port.  I consider this case a misconfiguration, but the extra conditional ensures that we only send packets to the first GS which sends to us instead of potentially switching back and forth with neither groundstation getting the full stream.  If broadcast worked efficiently on arduino, we wouldn't need to switch to unicast and all groundstations would get all messages.

> I would not be surprised if UDPCl will eventually disappear ... if I understand correctly what you were saying it was a kind of fluke ... I just don't want to touch it, there is plenty of time to decide its fate lateron

OK.
I would suggest removing UDPCl as soon as you are confident in UDPSTA since it does have a couple of issues as is and UDPSTA does cover the only use case I know of for UDPCl.

brad112358:
> I assume you have thoroughly tested the three modes, tcp, udp, udpsta, right

Well, I expected it to work the same as my first two PRs since it's mostly just cosmetic diferences, so I had given all 3 protocols only a brief test.  Your question prompted additional testing and I did find a problem with UDP mode; it wasn't sending to the correct broadcast address, though it worked if the groundstation sends first like in UDPcl mode.  It turns out that WiFi.broadcastIP() only works correctly in station mode.  I consider this a bug in Arduino, but it's trivial to avoid in our case.  Of course my first two PRs did work correctly.

olliw42:
> WiFi.broadcastIP()

I can't help to comment that I was asking about that function specifically ;)

> you considered making configuration work without GPIO0

I'm not sure how had that thought, I think there was a discussion and you made a good suggestion of how to do that ... however, when I added the AT mode I in fact tried to go that way, and I misserably failed. I in fact struggled to get the whole timing right at all. Can't remember the exact issues but I remember that I was pretty pi..ed off by the esp's behaviors. What I did not want is that you power on and when have to wait for a good couple secs until the tx module gets to life.  
So, at this point in time I'm not confident that's realistic and ever will materialize ... but as always, we will see

olliw42:
@brad112358 the PR looks good to me, except I would modify some comments a bit (we seem to follow a bit a different philosophy here). So, if I would merge and would in a follow up PR modify the comments, would this be ok for you?

jlpoltrack:
Just an FYI around the GPIO0 discussion, for ESP32-C3, GPIO9 is used instead (found this out while trying to get the backpack working on the Nomad module):

image

olliw42:
good to know, I guess once we are in a state to add c3 we may call it gpio_boot

brad112358:
> > WiFi.broadcastIP()
> 
> I can't help to comment that I was asking about that function specifically ;)

Yup, and it behaves as I described when WiFi is configured as a station.  What I had missed is that it does nothing useful when WiFi is an AP.

brad112358:
> @brad112358 the PR looks good to me, except I would modify some comments a bit (we seem to follow a bit a different philosophy here). So, if I would merge and would in a follow up PR modify the comments, would this be ok for you?

OK

brad112358:
> I'm not sure who had that thought, I think there was a discussion and you made a good suggestion of how to do that ... however, when I added the AT mode I in fact tried to go that way, and I misserably failed. I in fact struggled to get the whole timing right at all. Can't remember the exact issues but I remember that I was pretty pi..ed off by the esp's behaviors. What I did not want is that you power on and when have to wait for a good couple secs until the tx module gets to life.
So, at this point in time I'm not confident that's realistic and ever will materialize ... but as always, we will see

OK, then, I think the solution might be to ignore the GPIO0 connection and just use serial break signaling to enter AT command mode.  This should work for all esp based bridge setups, is effectively out-of-band and should have about the same timing constraints as GPIO0 signaling.

olliw42:
I'm not sure it's going to be as easy as you make it sound, but we will see
I'm not super eager to do that, I'm pretty happy with how it is currently

PR #329, opened 2025-05-17
bind, just sx127x

jlpoltrack:
Just check for presence of SX127x, not dual?


olliw42:
äääähhhhhmmmmm ,,,, good catch THX :)

PR #332, opened 2025-05-24
Configure wireless bridge for betafpv-micro-1w-2400

brad112358:
This depends on PR #331 

Untested.

### DIFF
--- a/mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
+++ b/mLRS/Common/hal/esp/tx-hal-betafpv-micro-1w-2400-esp32.h
@@ -31,6 +31,9 @@
 #define DEVICE_HAS_SINGLE_LED_RGB
 #define DEVICE_HAS_I2C_DISPLAY_ROT180
 #define DEVICE_HAS_FAN_ONOFF
+#define DEVICE_HAS_ESP_WIFI_BRIDGE_ON_SERIAL
+#define DEVICE_HAS_ESP_WIFI_BRIDGE_CONFIGURE
+
 
 // Note on SERIAL_OR_COM:
 // The com uart is not initialized, the serial uart is, So, buffers are set as by the RX/TXBUFSIZE defines for serial.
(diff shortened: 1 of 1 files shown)


jlpoltrack:
Seem to be missing something:

Flash bridge with #331 branch, uncomment ESP82XX, GPIO0_IO
Flash BetaFPV with with this branch

See settings but they don't apply.  Tried to change the WiFi Channel to 13, the Lua reports 13 after a power cycle but Win says the AP is still on channel 6.

olliw42:
I said before that I had spend quite some time on getting the timing and stuff right ... you made some changes and it might be related to that ... I'll make some comments in the other PR

jlpoltrack:
> Can you confirm that the BetaFPV esp32 has no direct control of the reset or gpio0 lines on the esp8266?

This is correct, only Tx / Rx are hooked up to the ESP32 (there's dipswitch block for those lines).

brad112358:
@jlpoltrack I was able to reproduce the problem you observed on my RM pocket by commenting out the first part of AtMode::Do() which watches for changes on GPIO0_IO.  In this test, the problem was that we don't wait long enough at startup for the first AT command because the esp8266 gets going faster than the esp32 and times out.  The STM in the R9M must startup a little faster than the esp32.  Please try changing the initialization of startup_tmo_ms in AtMode::Init() to "startup_tmo_ms = 1750 + millis();" and let me know if that solves the problem.

Thanks!

jlpoltrack:
> Please try changing the initialization of startup_tmo_ms in AtMode::Init() to "startup_tmo_ms = 1750 + millis();" and let me know if that solves the problem.

Yes, that seems to have done it - now WiFi channel changes are applied.

brad112358:
> Yes, that seems to have done it - now WiFi channel changes are applied.

Thanks!  It would be nice to have some idea where the cutoff is.  In my tests, 950 worked, but 850 didn't, but I expect your casemight require more time than my test.  Can you try a couple more tests, perhaps 950 and 1100 and see about where it starts working?  

Thanks!

PR #336, opened 2025-06-19
B&W Lua: Make reload always jump back to the first page.

### LINKED ISSUE #334: Function of LOAD in mLRS-bw-lua script
Hi gentlemen, 
The LUA script for small screen radios (mLRS-bw-luac) seems to have a minor annoyance. 
The (re)_load_ button does not work on the 2nd and upcoming pages. The button works properly on the first page.

**Hardware**:
**RM Zorro** (OTX 2.3.15) with external modules:
- RM Bandit Nano
- DIY E28 F103CB
- Wio E5 Mini
 
**RM Pocket** internal ELRS (ETX 2.10.6)
- internal module flashed with mLRS v1.3.6 or latest dev build (v1.3.7)

**Steps to reproduce the effect:** 
- Enter script. 
- Go to second page or further. 
- Change a parameter. 
- Press "load". Do **not** press "_save_".

**Expected behavior:**
- The parameter should go back to it's initial value. 

**Discovered behavior**
- The parameter keeps its new value.

**other actions done:**
- checked the CLI for the external modules, "_reload_" works as expected
- run the external small bay modules on **RM-TX16S** via BetaFPV module adapter, in the color version of the lua script "reload" works as expected.

brad112358:
This causes MBRIDGE_CMD_REQUEST_INFO to be sent sent so that setup.h:setup_reload() will be called to reload the parameters from NV.

Fixes #334

### DIFF
--- a/lua/mLRS-bw.lua
+++ b/lua/mLRS-bw.lua
@@ -881,7 +881,9 @@ local function doPage(event)
             elseif page == 0 and cursor_idx == Boot_idx then -- Boot pressed
                 sendBoot()
             elseif cursor_idx == Reload_idx - s then -- Reload pressed
+                page = 0  -- move to page 0 to force MBRIDGE_CMD_REQUEST_INFO
                 clearParms()
+                cursor_idx = 9
             elseif cursor_idx == Prev_idx - s then -- Prev pressed
                 clearParms()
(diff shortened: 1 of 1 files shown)


olliw42:
THX

PR #340, opened 2025-07-03
Do Not Merge - 2.4 LR1121

jlpoltrack:
Not for merging, just to show what is necessary for 2.4 to function on LR1121


olliw42:
so, this is again the either-or approach, right?

jlpoltrack:
> so, this is again the either-or approach, right?

Correct, will add this with future PR.

PR #341, opened 2025-07-14
Override stm32cube IDE dirs via OS env

chexov:
KISS way to override toolchain dirs for your needs.
```
export MLRS_ST_DIR="/Users/username/Applications/STM32CubeIDE.app"
export MLRS_GNU_DIR="Contents/Eclipse/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.12.3.rel1.macos64_1.0.100.202403111906/"
python tools/run_make_firmwares.py
```

### DIFF
--- a/tools/run_make_firmwares.py
+++ b/tools/run_make_firmwares.py
@@ -87,6 +87,10 @@ if __name__ == "__main__":
         st_root = os.path.join("/opt",'st')
 
     ST_DIR,GNU_DIR = findSTM32CubeIDEGnuTools(st_root)
+    if os.getenv("MLRS_ST_DIR"):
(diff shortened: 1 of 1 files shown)


olliw42:
many THX!
I see your intention. I personally am however not so excited about this configure via environment thing, which is likely because of my Win  based nature.

Have you considered alternatives, such as command line parameters for setting these, or an ini file or similar?

If such approaches would be no solution to you, I think at minimum the env variables should be named MLRS_ST_DIR and MLRS_GNU_DIR.

olliw42:
as it doesn't harm to have this option I'll merge, other options can still be considered

MANY THX

PR #347, opened 2025-09-13
readme, range record

jlpoltrack:
Updates the ReadMe to Include link to the Discord post of the current range record.


olliw42:
great
I guess I would keep it a bit more neutral, like "In the 900 MHz band, a range of 51 km at 1 W has been reported".

PR #352, opened 2025-11-02
Nomad Tx Support

jlpoltrack:
- Adds Nomad 900 and 2400 Targets, tested with XR4 on 900 and PA-D on 2400
- Need guidance on where to have the button2 check (currently in mlrs-tx.cpp which is not great)


olliw42:
many THX

yes, button2 can't be in main
it should go in some library, and the functionality then enabled by some DEVICE_HAS_BUTTON_FLASH or so
also the dependency on esp. is not so nice ... hmhm
mybe some lib flash.h/cpp, similar to bind?

the LR can do both 900 and 2.4, right. I think we want to not have two targets for each, but to have one target and have it selectable by the user.
I should get some idea on how to do from my efforts with dualband, I may find time next weekend.

in the hal file, could there be some info in the top on e.g. how to flash etc, see some other esp targets, so we do not have to research each time. We need that info also to add it to mLRS-Flasher properly.

olliw42:
the nomad would also allow diversity, right?

olliw42:
@jlpoltrack 
I've worked yesterday and today on the topic, but I'm afraid I haven't produced anything directly useful.

One issue is that I don't have a nomad but justthe matex dualband hardware (I think you have that too, don't you?), which has an sx126x and sx128x, so the switch between the bands also involves a swicth of sx<->sx2. I haven't gotten this to fully work, on 2.4 GHz it connects very occasional only ... couldn't yet figure out what the issue is (all sx/fhss related settings seem to be ok). I've pushed a branch https://github.com/olliw42/mLRS/tree/dev-matek-mtx-multiband, as it might give you some indications (look for DEVICE_HAS_MULTIBAND).
(maybe you have time to look at it too, and may see the issue).

I first was thinking that this would be for you too, and that you just would have to add an DEVICE_HAS_MULTIBAND_LR1121_LR1121 option, but only late in the work realized that you do have two lr1121 stages, so want to do the 900&2.4GHz on the same sx channel, with the other being just diversity.

This should be in fact much, much simpler. Before I comment on this, let m,e comment on the other changes.

I think it would be useful to start a new branch with jsut the changes needed to get 900&2.4 to work, and without all the other esp stuff.

As regards that:
I think doing the button as you did makes sense. I am just puzzled that your debounce code is so much simpler than for bind, see https://github.com/olliw42/mLRS/blob/main/mLRS/Common/bind.h#L129-L153. I actually think your code is too simple. E.g., EnterFlash() would be repeatedly called. That is, I guess here some more debounce and entering care could be needed. The line with USE_ESP_WIFI_BRIDGE_BUTTON_FLASH can go however, use DEVICE_ESP_WIFI_BRIDGE_BUTTON_FLASH. The USE_xxx flags are only intrdouced then mutliple DEVICE_xxx can lead to the same.

olliw42:
as regards the 2.4 & 900 changes in setup.h
did this:

[setup.zip](https://github.com/user-attachments/files/23441836/setup.zip)

untested of course, just my best guess

you may have to do also few changes in fhss.h, I suspect GetCurrFreq_Hz(), GetFreq_x1000() may need adjustment according to the actual band.

I'm not sure that bind would need changes, maybe not, could be though the case as one may have to narrow down to the scanned frequencies. 

The power setting I have made similar to how it is for the sx126x, that is in the function lr11xx_rfpower_calc() the 4th parameter is now the FrequencyBand, so that you can adjust it accordingly. See https://github.com/olliw42/mLRS/blob/main/mLRS/Common/sx-drivers/lr11xx_driver.h#L401.

This part https://github.com/olliw42/mLRS/blob/main/mLRS/Common/sx-drivers/lr11xx_driver.h#L219-L223 you also can make such that it accounts for the frequency band, should spare you your define, i.e., simplify this section https://github.com/olliw42/mLRS/pull/352/files#diff-9dd9bb2d157872028451d2095a3e81e32e09f600a22fffb926e63e8c18c1a8d6R264-R272.

olliw42:
ah, maybe this is what you meant:
you copy the 3 tSxLoraConfiguration for the 24 modes to lr11xx.h, and you then in e.g. this location https://github.com/olliw42/mLRS/blob/main/mLRS/Common/setup.h#L541 can set either the 24 or 900 index by an if (Config.FrequencyBand == 24GHz)

PR #355, opened 2025-11-11
Nomad Tx 2

jlpoltrack:
Early draft, doesn't compile.

At least one problem is that for the power calculation 'gconfig' is not created as part of SX1 / SX2 driver - should this be done or something else?

https://github.com/olliw42/mLRS/blob/8500c153d69244239cf14e810e11a7ab85dfc25a/mLRS/Common/sx-drivers/lr11xx_driver.h#L401

image

Edit - similar situation for bind.h, where configure_mode now needs the frequency as the second parameter.  What's best way to grab the frequency?

### DIFF
--- a/mLRS/Common/bind.h
+++ b/mLRS/Common/bind.h
@@ -111,11 +111,11 @@ void tBindBase::ConfigForBind(void)
     // switch to 19 Mode, select lowest possible power
     // we technically have to distinguish between MODE_19HZ or MODE_19HZ_7X
     // configure_mode() however does currently do the same for both cases
-#ifdef DEVICE_HAS_SX127x
-    configure_mode(MODE_19HZ_7X, Config.FrequencyBand;
-#else
-    configure_mode(MODE_19HZ, Config.FrequencyBand);
-#endif
+    if (Config.Mode == MODE_19HZ_7X) {
+        configure_mode(MODE_19HZ_7X, Config.FrequencyBand);
+    } else {
+        configure_mode(MODE_19HZ, Config.FrequencyBand);
+    }
 
     sx.SetToIdle();
     sx2.SetToIdle();
--- a/mLRS/Common/setup.h
+++ b/mLRS/Common/setup.h
@@ -95,7 +95,7 @@ void setup_configure_metadata(void)
 #elif defined DEVICE_HAS_LR11xx && defined DEVICE_IS_TRANSMITTER
     // MULTIBAND
     // we cannot work out all cases here, as it depends on actual FrequencyBand selection, so we here just do what we can do
-    SetupMetaData.Mode_allowed_mask = 0b010111; // 50 Hz, 31 Hz, 19 Hz, FSK
+    SetupMetaData.Mode_allowed_mask = 0b110111; // 50 Hz, 31 Hz, 19 Hz, 19 Hz 7x, FSK
 #elif defined DEVICE_HAS_LR11xx
     SetupMetaData.Mode_allowed_mask = 0b010110; // 31 Hz, 19 Hz, FSK
 #else
@@ -395,7 +395,7 @@ void setup_sanitize_config(uint8_t config_id)
         break;
     case SETUP_FREQUENCY_BAND_915_MHZ_FCC:
     case SETUP_FREQUENCY_BAND_868_MHZ:
-        SetupMetaData.Mode_allowed_mask &= 0b010110; // filter down to 31 Hz, 19 Hz, FSK
+        SetupMetaData.Mode_allowed_mask &= 0b110110; // filter down to 31 Hz, 19 Hz, 19 Hz 7x, FSK
         break;
     default:
         while(1){} // must not happen, should have been resolved in setup_sanitize()
--- a/mLRS/Common/sx-drivers/lr11xx_driver.h
+++ b/mLRS/Common/sx-drivers/lr11xx_driver.h
@@ -148,6 +148,8 @@ class Lr11xxDriverCommon : public Lr11xxDriverBase
                             config->CodingRate,
                             LR11XX_LORA_LDR_OFF);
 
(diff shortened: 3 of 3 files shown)


mLRS/Common/setup.h
```diff
@@ -1,1046 +1,1086 @@
...
-
-  adjust allowed_mask where it depends on setting
-
-setup_configure_config(Config.ConfigId):
-
-
-
-*/
+//*******************************************************
+// Copyright (c) MLRS project
```
olliw42: not sure what has happened here, but something went terribly wrong with setup.h., if marks all of the lines as being changed ...

pl repair that :)

mLRS/Common/bind.h
```diff
@@ -53,7 +53,7 @@ typedef enum {
 class tBindBase
 {
   public:
-    void Init(void);
+    void Init(tSxGlobalConfig* const global_config);
```
olliw42: I am not yoet sure this is really needed ... maybe resolöved by next comment ...

olliw42:
> Now compiles and 'works' but bind seems clunky as devices will only bind if they are on the configured band.

so, disregarding bind, it allows the user to swicth to all three bands? I would call that a first achievement. Great.

> bind seems clunky as devices will only bind if they are on the configured band.

first, you used a wrong frequency band, may not be the reason however. 

The general idea is that bind cycles through the available band settings, see `HopToNextBind()`. That's this `FrequencyBand_allowed_mask` thing. It is handed over to fhss class in its `Init()` in the variable `fb_allowed_mask`. It hsould have the three bits set for the three bands, i.e., I would debug that first if that's the case. (note: since it goes into fhsss its bits are of type `SX_FHSS_CONFIG_FREQUENCY_xxx`). Its bits are then copied over into the variable `bind_scan_mask`.

My thought - hope - was that it might be sufficient that the three bits are set, but this may not be the case ... hm, looking at tBindBase::ConfigForBind(void), it actually seems that the Tx is NOT cylcing through ... 
we may have to reinvestigate the bind process

that's btw also a major thing which held me back with the dualband improvements for the matek dualband device, since it looked to me it needs to get a way more complicated scheme in that case :(

olliw42:
ähm, for the nomad I'm actually not sure if it's not the correct bind approach that it doesn't cycle but just tries to find the receiver on the adjusted band ...

hmhm

mLRS/Common/sx-drivers/lr11xx_driver.h
```diff
@@ -324,8 +365,9 @@ class Lr11xxDriverCommon : public Lr11xxDriverBase
         return actual_power_dbm;
     }
 
-  private:
     tSxGlobalConfig* gconfig;
```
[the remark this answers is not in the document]
jlpoltrack: Done.

mLRS/Common/bind.h
jlpoltrack: Back to original.

jlpoltrack:
> ähm, for the nomad I'm actually not sure if it's not the correct bind approach that it doesn't cycle but just tries to find the receiver on the adjusted band ...

One thought is to have different behavior for Tx compared to Rx, could be:
- Tx will cycle through all bands it has
- Rx will only use one band for binding

For completeness, I could bind in the previous code but I had to know that the band in EEPROM was the same on both devices (not great).

mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-esp32.h
```diff
@@ -0,0 +1,407 @@
...
+#endif // DEVICE_HAS_ESP_WIFI_BRIDGE_ON_SERIAL2 
+
+
+//-- POWER
+
+#include "../../setup_types.h" // needed for frequency band condition in rfpower calc
+#define SX_USE_LP_PA  // Nomad uses the low power amplifier for the 900 side
+
+
+void lr11xx_rfpower_calc(const int8_t power_dbm, uint8_t* sx_power, int8_t* actual_power_dbm, uint8_t frequencyBand)
```
olliw42: pl call the last variable frequency_band, and const it
see e.g. here https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/hal-power-pa.h#L180

mLRS/Common/hal/esp/tx-hal-radiomaster-nomad-esp32.h
```diff
@@ -0,0 +1,407 @@
...
+{
+    if (frequencyBand == SX_FHSS_CONFIG_FREQUENCY_BAND_2P4_GHZ) {  
+        if (power_dbm > 28) { // -> 30
```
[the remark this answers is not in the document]
jlpoltrack: Addressed

olliw42:
hahah ... so it's not exactly the complete table ;)
but I guess you are saying it all works just fine

not sure if there is any other side effect we need to consider? 
maybe with the power rc channels switching? this still works?
lua should be good, right?
any further concern?

jlpoltrack:
> is that already acounted for/avilable/working?

Well now it does :)

olliw42:
and it binds correctly?

i.e., when you choose 915 in the tx, and it's a sx1275 receiver, it will bind and jump into 19 Hz 7x mode, while if it is a sx1262 receiver it will bind and jumpt into 19 Hz (or higher) mode?

sorry for being so pesky :)

jlpoltrack:
> i.e., when you choose 915 in the tx, and it's a sx1275 receiver, it will bind and jump into 19 Hz 7x mode, while if it is a sx1262 receiver it will bind and jumpt into 19 Hz (or higher) mode?

Should be all the scenarios for 900 MHz, do you see one that shouldn't work?

- Nomad with 19 7x mode, binds fine with Generic 900 (SX1276)
- Nomad with 31 Hz mode, binds fine with XR1 (LR1121)
- Nomad with 31 Hz mode, binds fine with Matek mR900-22 (SX126x)

PR #358, opened 2025-11-13
ESP Button Flash

jlpoltrack:
Title, used bind style debounce.


olliw42:
MANY THX

btw, I am thinking about creating a button class (since we have it now in two places), so may change pieces here, and maybe will reask you to test :)

PR #360, opened 2025-11-15
LR11xx Receivers Multiband

jlpoltrack:
Title, need to figure out:
- Why Bind doesn't work when Tx is on 2.4 and Rx is on 868 (e.g. after erase).  Looks like a call to ConfigForBind is needed when switching between 868/915 and 2.4 as the modulation parameters will be different?
- Confirm power levels
- Not allow 19 Hz 7x mode?

### DIFF
--- a/mLRS/Common/bind.h
+++ b/mLRS/Common/bind.h
@@ -113,7 +113,7 @@ void tBindBase::Init(void)
 
 void tBindBase::ConfigForBind(void)
 {
-    // switch to 19 Hz Mode, select lowest possible power
+    // used by both the Tx and Rx, switch to 19 Hz mode, select lowest possible power
     // we technically have to distinguish between MODE_19HZ or MODE_19HZ_7X
     // configure_mode() however does currently do the same for both cases
     if (Config.Mode == MODE_19HZ_7X) {
@@ -133,6 +133,7 @@ void tBindBase::ConfigForBind(void)
 void tBindBase::HopToNextBind(uint16_t frequency_band) // SETUP_FREQUENCY_BAND_ENUM
 {
     // not nice
+    // used only by Rx
     // we would need SetupMetaData.Mode_allowed_mask before it is adjusted for the selected frequency band
     // we could keep a copy of the un-adjusted SetupMetaData.Mode_allowed_mask
     // we also could provide a function to do it both in setup.h and here


olliw42:
>     * Why Bind doesn't work when Tx is on 2.4 and Rx is on 868 (e.g. after erase).  Looks like a call to ConfigForBind is needed when switching between 868/915 and 2.4 as the modulation parameters will be different?

that's the part which isn't totally clear to me either. One needs to change the config, but I would think it does already to swicth between 868 and 915 ... but I can't see that immediately in the code LOL
maybe it in fact just uses one of these frequencies ...

>     * Not allow 19 Hz 7x mode?

I guess it would be nice if it could be allowed. This implies one needs also test binding with a sx1276 tx module ...

jlpoltrack:
> One needs to change the config, but I would think it does already to swicth between 868 and 915

868 / 915 would use the same modulation parameters, only need to set a different frequency.  So I think fhss.h is able to do that based on what I see in the code.

> I guess it would be nice if it could be allowed. This implies one needs also test binding with a sx1276 tx module ...

Moreso, this logic needs to be redone then: https://github.com/olliw42/mLRS/blob/9267e5cd1c8b11d47ec957813ec696fbab911469/mLRS/Common/bind.h#L114-L117

The receiver would need to cycle through 2.4, 868, 915, 868 7x, 915 7x (or some variation of these 5)

olliw42:
so, there is no config besides the frequencyband called in fhss, i.e., bind is only called outside of fhss ...
I guess if one doesn't want to poluute fhss with the bind class, a way to go would be a call a new bind. method after a `fhss.HopToNextBind()`, i.e., to calls a fhss method to get the frequencyband and then call that new bind method with the frequencyband as variable
that is, at this line https://github.com/olliw42/mLRS/blob/main/mLRS/CommonRx/mlrs-rx.cpp#L843 one may insert something like this
`bind.HopToNextBind(fhss.GetBindBand())`

jo, that piece of code needs to be changed ... could be similar/same code as in the new `bind.HopToNextBind()`

just a wild guess

mLRS/Common/bind.h
```diff
@@ -83,7 +84,8 @@ class tBindBase
     int8_t pressed_cnt;
 
   private:
-    tSxGlobalConfig* gconfig;  
+    tSxGlobalConfig* gconfig;
+    bool last_was_7x = false;
```
olliw42: hm ... this I do not like, as it brings in stateness ,,,
let's see what we can do about it

mLRS/Common/bind.h
```diff
@@ -127,6 +129,42 @@ void tBindBase::ConfigForBind(void)
     sx2.SetToIdle();
 }
 
+void tBindBase::HopToNextBind(uint8_t frequency_band)
+{
+#ifdef DEVICE_HAS_LR11xx
```
olliw42: for the moment ok, but I think we want to have something more generic/general
we want to try to have as little sx-dependencies as possible
so, we could add a comment for now, `// TODO: replace by some more generic flag`

mLRS/Common/hal/esp/rx-hal-generic-c3-lr1121-esp32c3.h
```diff
@@ -170,19 +170,41 @@ IRAM_ATTR void led_blue_toggle(void)
...
+            *actual_power_dbm = 10;
+        }
+
+    }
+}
 
 #define RFPOWER_DEFAULT           0 // index into rfpower_list array
 
 const rfpower_t rfpower_list[] = {
-    { .dbm = POWER_0_DBM, .mW = 1 },
```
olliw42: this is sad that it isn't supported anymore ;)

mLRS/Common/hal/esp/rx-hal-generic-lr1121-td-esp32.h
```diff
@@ -220,19 +220,41 @@ IRAM_ATTR void led_blue_toggle(void)
...
+            *actual_power_dbm = 10;
+        }
+
+    }
+}
 
 #define RFPOWER_DEFAULT           0 // index into rfpower_list array
 
 const rfpower_t rfpower_list[] = {
-    { .dbm = POWER_0_DBM, .mW = 1 },
```
olliw42: this is sad that it isn't supported anymore ;)

olliw42:
first, congratulations, that's a great step forward
I do have some comments, since I'm not yet totally happy with the architecture
my suggestions may not be the final call, I think we just shoudl want to iterate a bit to converge towards a cleaner architecture

mLRS/Common/bind.h
```diff
@@ -127,6 +129,42 @@ void tBindBase::ConfigForBind(void)
...
+    // LR11xx can connect to SX126x, SX127x and SX128x devices, so special considerations needed for binding
+    // for bands that can support the 19 Hz 7x mode, we alternate between 19Hz and 19Hz 7x
+    bool band_supports_7x =
```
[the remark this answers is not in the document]
jlpoltrack: See note in previous comment about allowed mask.  One thought - would it be preferred to check if 19 Hz 7x mode is in the mask then just not switch to 7x when the frequency is 2.4?  Sort of like the inverse logic of what we have now?

mLRS/Common/hal/esp/rx-hal-generic-lr1121-td-esp32.h
olliw42: see before

mLRS/Common/hal/esp/rx-hal-generic-c3-lr1121-esp32c3.h
olliw42: we have that "min" option, we may use that here

mLRS/CommonRx/mlrs-rx.cpp
```diff
@@ -891,7 +893,8 @@ dbg.puts(s8toBCD_s(stats.last_rssi2));*/
         bind.Do();
         switch (bind.Task()) {
         case BIND_TASK_CHANGED_TO_BIND:
-            bind.ConfigForBind();
+            bind.ConfigModeForBind();
+            bind.ConfigRfForBind();
```
olliw42: see before, pl revert

mLRS/CommonRx/mlrs-rx.cpp
```diff
@@ -840,6 +840,8 @@ dbg.puts(s8toBCD_s(stats.last_rssi2));*/
                 link_state = LINK_STATE_RECEIVE; // switch back to RX
             }
             if (fhss.HopToNextBind()) {
+                bind.HopToNextBind(fhss.GetCurrBindSetupFrequencyBand());
+                bind.ConfigRfForBind();
```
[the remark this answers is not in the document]
olliw42: the name of the local method would then be config_rf_for_bind(), following the current pattern 
(sadly, the pattern is not always perfectlöy consistent, but that's the pattern in bind LOL)

jlpoltrack:
> Point is we have to consider also other scenarios besides the specifric one we currently do for the lr1121. For instance, I would the scheme to also work for the mateak dual band device.

Would it be fair to say then that there is a third rule here :) e.g. 'must work across all hardware' (instead of just LR11xx which is what I've been considering)

mLRS/Common/bind.h
```diff
@@ -57,7 +57,8 @@ class tBindBase
     bool IsInBind(void) { return is_in_binding; }
     void StartBind(void) { binding_requested = true; }
     void StopBind(void) { binding_stop_requested = true; }
-    void ConfigForBind(void);
+    void ConfigModeForBind(void);
```
olliw42: ConfigForBind() :)

olliw42:
> Would it be fair to say then that there is a third rule here :) e.g. 'must work across all hardware' (instead of just LR11xx which is what I've been considering)

well, you've brought up 433 ;)
generally, we do not want any hardware specific things in the higher level codes
I know, we not always achieve or do, but try. This here will quickly get into a nightmare with any little addition. You named 433, I'll name dualband. Someone else may want only 915, yet someone else may want 900 IN. So, it's not about hardware, it's also about what features to use on a hardware.

I think a uint16_t which so to say maps the frequency band allowed_mask should allow us to do the 19Hz/19Hz7x  toggles every call. Let's call it momentarily mode_mask.
On entering the bind it is set to all zero or all 1 (0xFFFF) depending on the Config.mode being 19Hz or 19Hz7x.
In HopToNext, the bit corresponding to the current frequency band is toggled. If it's a band which is "sensitive" to 19Hz or 19Hz7x, this bit value is taken to set 19Hz or 19Hz7x.

Should make us going through the above cycle, with not too much code exercises

If not clear I could attempt a code example, and test it on normal devices.

mLRS/Common/bind.h
jlpoltrack: Done

mLRS/Common/bind.h
```diff
@@ -103,20 +106,65 @@ void tBindBase::Init(void)
...
     auto_bind_tmo_ms = 1000 * RX_BIND_MODE_AFTER_POWERUP_TIME_SEC;
+
+    mode_mask = 0;
 }
 
 
 void tBindBase::ConfigForBind(void)
 {
-    // switch to 19 Mode, select lowest possible power
+    // switch to 19 Hz Mode, select lowest possible power
```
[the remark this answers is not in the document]
jlpoltrack: No problem, think should be good now.

olliw42:
ah. No problem, if that has it made it into here. I can close the other PR.

this here I think shoud however be tested a little bit more thoroghly

PR #361, opened 2025-11-18
common, setup, adjust mode according to hardware capabilities

olliw42:
@jlpoltrack 
see code change
I believe that this line was incorrect, as the allowed modes need to be adjusted according to the capabilities of the hardware.

For instance, for the LR2021, one would allow also FLRC for 2.4GHz, that is the current bit mask would not apply.

I'd like you to double check and confrim that is still doing the right thing for the lr1121


olliw42:
so you've tested it sufficiently extensively?

PR #362, opened 2025-11-19
Color Lua Update for EdgeTx 3.0

jlpoltrack:
EdgeTx 3.0 will drop support for some of the color constants used by the current lua script (https://luadoc.edgetx.org/lua-api-reference/constants/color-constants#deprecated-color-constants).  This means that the existing script will no longer function.  Updating to the new color constants will also break support for OpenTx.

Approach:
- Rename existing lua to `mLRS-OTX.lua`
- Create new lua with updated color constants called `mLRS-ETX.lua`

Tested working on EdgeTx 2.11, untested on EdgeTx 3.0 (waiting for hardware).

CC: @rotorman

### DIFF
--- a/lua/mLRS.lua
+++ b/lua/mLRS.lua
@@ -10,12 +10,7 @@
 -- copy script to SCRIPTS\TOOLS folder on OpenTx SD card
 -- works with mLRS v1.3.03 and later, mOTX v33
 
-local version = '2025-11-19.01'
-
-local g_textColor
-local g_titleBgColor
-local g_menuTitleColor
-local g_textDisableColor
+local version = '2025-11-20.01'
 
 local required_tx_mLRS_version_int = 10303 -- 'v1.3.03'
 local required_rx_mLRS_version_int = 10303 -- 'v1.3.03'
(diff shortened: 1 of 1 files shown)


lua/mLRS.lua
```diff
@@ -10,7 +10,12 @@
 -- copy script to SCRIPTS\TOOLS folder on OpenTx SD card
 -- works with mLRS v1.3.03 and later, mOTX v33
 
-local version = '2025-11-15.01'
+local version = '2025-11-19.01'
+
+local g_textColor
```
olliw42: please move them to before the line local page_N1 = 9 below

lua/mLRS.lua
```diff
@@ -1350,10 +1375,8 @@ end
 ----------------------------------------------------------------------
 
 local function scriptInit()
-    local ver, radio, maj, minor, rev, osname = getVersion()
```
olliw42: pl keep them
see comment in the above

olliw42:
very nice, and thanks @rotorman for the suggestion, I really like it much better to have it in one file :)

I guess you've used find/serach to ensure you've reoplaced all occurences :)

two little nitpicks

PR #366, opened 2025-11-25
TX15 Internal LR1121 RF

jlpoltrack:
Title.  Tested working with XR4, will followup with corresponding update to mLRS Flasher (has ESP32C3 internal bridge, so this needed some additions).


mLRS/Common/hal/esp/esp-hal.h
```diff
@@ -139,3 +139,7 @@
 #include "tx-hal-radiomaster-nomad-esp32.h"
 #endif
 
+#ifdef TX_ELRS_RADIOMASTER_INTERNAL_TX15_ESP32
```
olliw42: I guess this rather may go to line 122, to the "int" section

jlpoltrack: Moved.

olliw42:
very nice
MANY THX
I'm merging this

PR #372, opened 2025-12-02
setup, rc channels check

jlpoltrack:
- Don't enable rcoverride for all ESP targets, only if they don't have out
- This could be simpler without any ESP check - is there any scenario where a device that doesn't have out shouldn't have this enabled?

### DIFF
--- a/mLRS/Common/setup.h
+++ b/mLRS/Common/setup.h
@@ -245,13 +245,10 @@ void setup_configure_metadata(void)
 //-------------------------------------------------------
 
 // override default setup setting from common_conf.h
-// TODO: when AP4.6 is out, the default should become SEND_RC_CHANNELS_RADIORCCHANNELS
-// RC Channels is only included with 2 MB boards, so override makes more sense?
-#if !defined USE_OUT
-#if defined ESP32 || defined ESP8266
+// enable rc channels for single serial receivers
+#if !defined DEVICE_HAS_OUT
   #undef SETUP_RX_SEND_RC_CHANNELS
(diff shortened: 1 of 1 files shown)


olliw42:
not sure how the esp checks came in
in principle I think you are absolutely correct, that we should do that consistently for when there is no OUT, can't see a logical reason to differentiate here between the chipsets
only reason I could see that we somehow somewhere don't do things fully consistently, i.e., I guess we would want to check which esp targets would be affected and see if they should be affected ...
would be great to get it consistent and thus clean here

I think we should grab the opportunity and actually change it to SEND_RC_CHANNELS_RADIORCCHANNELS
4.6 is out for a while, and if it doesn't suit the user theyx always can adjust through the parameters - it's only the default here!
rcoverride really should die out

jlpoltrack:
> not sure how the esp checks came in

Most ESP hardware has single serial port, therefore it was suggested by users to enable to avoid them having to do so in the Lua for a better experience when getting started.

> for the rp4td we do have a sik firmware, based on the argument it has two uarts ... this could suggest we want to do for e.g. the xr4 too ... however, the rp4td sik appears to give the largest issues, by the number of reports ... hmhm

I would prefer not to, there's enough other Tx hardware that offers a much better experience.

I will update the PR to:
- Use DEVICE_HAS_OUT only
- Enable RC Channels

olliw42:
> I would prefer not to, there's enough other Tx hardware that offers a much better experience.

glad you are saying that :D:D:D

Issue #342, opened 2025-08-03
RP4-TD UART-bridge setup enters failsafe (100 ms red blink) and never reconnects


ddosoff:
Good day, OlliW 👋

First of all, thank you for your tremendous work on mLRS!
What I’m doing

    Two RadioMaster RP4-TD 2.4 GHz boards

        TX side – flashed with tx-radiomaster-rp4td-2400-sik-telem

        RX side – flashed with rx-radiomaster-rp4td-2400

    Using them purely as a UART bridge (simple “pinger” app on the serial ports confirms the link).

Issue

    Link works flawlessly at short range.

    When I walk far enough for the RX to lose the signal, it correctly enters failsafe and the red LED begins flashing very rapidly (~100 ms period).

    After that the RX never regains the link, even when I move back into range; only a rx power-cycle helps.

Expected behaviour

RX should automatically exit failsafe and reconnect to the TX once the signal returns.
Actual behaviour

    Remains in failsafe indefinitely.

    LED keeps flashing at ~10 Hz.

    No data are forwarded over UART until the RX is rebooted.

Environment

    mLRS commit: … <– (current master, cloned 2025-08-03)

Could you please point me to the proper fix or advise how to make the RX re-enter LISTEN → SYNC → CONNECTED automatically?

Thank you! 🙏

olliw42:
this does not look like a normal failsafe

it looks to me like it goes into one of the failure modes
you say it blinks at 10 Hz. Can you pleae describe in more detail the blink pattern of the leds. For both rx and tx (they can be different)

@jlpoltrack @brad112358 @tmcadam   I suspect an issue with that one of them goes into a FAIL. Given the past, a luikely reason could be on the tx side with some timing. Since I'm not so much on the esp side of things, your input would be valuable :)

jlpoltrack:
> But RX still reboots some time on weak signals. I see ESP booting message on the UART.

Can you try and capture the crash + boot message from the ESP? The crash message should be right before the boot message.

ddosoff:
I am in the UART bridge mode. No flight controller. Just ping + echo via UART.
To reproduce the weak signal you can do this:

![Image](https://github.com/user-attachments/assets/8e4c51bd-9734-410e-a92a-a66766b859b6)
![Image](https://github.com/user-attachments/assets/88d7fb60-e6dc-43a4-952d-267411871fcc)

No code modifications.

PR #376, opened 2025-12-08
Dual Band for LR1121 Diversity Hardware

jlpoltrack:
General changes:
- Adds two new frequencies, 915+2.4 and 868+2.4, supporting dual-band transmission on 31 Hz and 19 Hz
- For dual band frequencies, sx1 will always be 868/915 and sx2 will always be 2.4
- Made sx2 always transmit first since 2.4 OTA is longer for both modes
- Multi band support remains with usual frequencies
- Helper function added to determine whether or not current freq is dual band
- Fixed bug in reg to kHz calculation for LR11xx devices
- Fixed bug in rx mavlink_interface, now correctly reports freq2 for dual band
- Updated CLI to print out both sx and sx2 freq sets

More dramatic changes:
- Initially got a lot of Tx Done fails as the first Tx Done IRQ advances the link state so that when the second Tx Done IRQ was received it would get caught by the fail logic. Fail logic now doesn't treat this as a fail for dual band frequencies
- Updated the sx IRQs to only set a flag, the staggered transmission times as a result of the OTAs being different (meaning different reception times) caused watchdog resets. I believe these resets were a result of the IRQ trying to use the SPI bus when something in the main loop was already using it. Resets went away when this change was made.

Not clear to me how STM32 dual band avoided the first issue.  I suppose always using two separate SPI would prevent the second one from being a problem.

### DIFF
--- a/mLRS/Common/arq.h
+++ b/mLRS/Common/arq.h
@@ -206,6 +206,7 @@ void tTransmitArq::SetRetryCntAuto(int32_t _frame_cnt, uint8_t mode)
     case MODE_31HZ:
     case MODE_19HZ:
     case MODE_19HZ_7X:
+    case MODE_DUALBAND_FSK_50HZ:
         SetRetryCnt((_frame_cnt >= 800) ? 2 : 1);
     }
 
--- a/mLRS/Common/common.h
+++ b/mLRS/Common/common.h
@@ -241,33 +241,16 @@ void sxReadFrame(uint8_t antenna, void* const data, void* const data2, uint8_t l
     }
 }
 
-
-void sxSendFrame(uint8_t antenna, void* const data, uint8_t len, uint16_t tmo_ms)
+// Helper functions for cleaner sxSendFrame implementation
+static void sxSendOnBoth(void* const data, uint8_t len, uint16_t tmo_ms)
 {
-#if defined DEVICE_HAS_LR11xx && defined DEVICE_HAS_DIVERSITY_SINGLE_SPI
-    // LR11xx with diversity - check if dual-band frequency
-    if (is_dual_band_frequency(Config.FrequencyBand)) {
-        // dual-band frequency: send on both simultaneously
-        // sx2 (2.4 GHz) transmits first since it takes longer, reduce gap between TX done IRQs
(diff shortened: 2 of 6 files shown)


jlpoltrack:
> I frankly also don't understand why it is needed. If it is a single spi issue, why isn't it then also an issue with receive diversity. I.e., I'm worried that it is rather a bandage to an underlying undicsvovered issue. I would rather want the underlying issue to be resolved.

Fair enough - here's my thoughts on what is occurring, I lean more to there being a subtle bug here for single SPI hardware that is exposed by the staggered reception times:

- For reference, the 2.4 reception will come roughly 100 uS later for 31 Hz and 900 uS later for 19 Hz (based on OTA and choosing which to transmit first)
- When the 900 transmission comes, the Rx_done IRQ is serviced and then do_receive is called in the main loop which reads the buffer out from the lr chip - this takes some time.
- If the 2.4 reception came and triggered the Rx_done IRQ while this read is occurring, there is bus contention here and the watchdog reset will get triggered because the IRQ can't access the SPI.

Why is this not a problem for regular diversity? The two Rx Done IRQs will be triggered at roughly the same time since they should receive the packet at the same time since there is only one transmission.  Both of these will get serviced before do_receive has a chance to grab the SPI in the main loop.  I do also wonder if this can be an issue for regular diversity in poor LQ states as this user reported: https://github.com/olliw42/mLRS/issues/342.

olliw42:
good thought
but I'm not sure I'm convinced. I thinki we do set the timers and points such that the received packets are digested only well after a received packet, and isn't it such that it's equal for all lora modes, that is dependent only on the mode. Therefore, I would argue, these spi transfers happen well after the isrs. That's at least how it should be. 
If your argument is correct, I would think you should see a massive difference for 31 Hz and 19 Hz, since the delays are so different. 
Pure speculation, but maybe starting the next spi transfer on the esp takes "ages", so that when the two packets are transmitted whey indeed don't come at nearly the same time but massively set off in time. I guess some time taking shoudl reveal what's up here.

olliw42:
many thx!
the stm32 dualband will be very much along the lines you have pioneered, but it will indeed also have some tidies which would be good to first agree on.

PR #380, opened 2025-12-17
lua: Add Device Info page to mLRS-bw.lua

jlpoltrack:
Title, tested on GX12.  Doesn't show version since that is already on Main screen.

New option on Main screen:
image

New page, Prev / Next returns to Main:
image


olliw42:
@jlpoltrack I guess it's sitting now for long enough ... I should merge, right?

jlpoltrack:
> @jlpoltrack I guess it's sitting now for long enough ... I should merge, right?

Please do.

PR #381, opened 2025-12-17
Support up to 8 power levels, always display in mW, enable 7 for a few RadioMaster devices

jlpoltrack:
Currently only ~6.5 power levels are available.  Use a few more bytes to support 8, most devices will only need 7

Always display power in mW, tested on BW Lua, Color Lua, CLI and Display.

Updated power levels for Nomad, GX12 and TX15.

### DIFF
--- a/mLRS/Common/common_types.cpp
+++ b/mLRS/Common/common_types.cpp
@@ -494,10 +494,10 @@ void power_optstr_from_power_list(char* const Power_optstr, int16_t* const power
 {
     memset(Power_optstr, 0, slen);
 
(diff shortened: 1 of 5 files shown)


jlpoltrack:
Ok - enjoy your vacation :)

I did originally have 67 working without issues which is the max using item4. But thought that (67 - 44) * 2 = 46 extra bytes was a bit wasteful and decided that just 16 (8 * 2) was enough.

olliw42:
you won the W vs mW poll :) so merging

PR #382, opened 2025-12-18
color lua: auto retry, small 320x240 tweaks

jlpoltrack:
Two updates:

1. After modifying parameters on EdgeTx, it is not uncommon to have to do a reload (for me maybe 50% of the time) so automatically retry once. Perhaps related to using Tx with ESP bridges which do some configuration at startup. This change results in no manual reloads for me.
2. Adjust the 320x240 offsets so that the items are better distributed, tested on PA01.


olliw42:
I guess I suggest to also let it sit until I'm back
and can see&test myself :)
great work btw

PR #387, opened 2026-01-07
Add CAN FD

jlpoltrack:
Somewhat academic.

Had to use 160 MHz Sysclock / 80 MHz fdcan peripheral clock to play nicely with ArduPilot.

Tested on Matek H743 WLITE + mR900 TD.

Debug: 
Screenshot 2026-01-05 at 1 48 23 PM

Setup: 
image

### DIFF
--- a/mLRS/CommonRx/dronecan_interface_rx.h
+++ b/mLRS/CommonRx/dronecan_interface_rx.h
@@ -573,7 +573,7 @@ void tRxDroneCan::SendRcData(tRcData* const rc_out, bool failsafe)
         _p.rc_input.rcin.data[i] = (((int32_t)(rc_out->ch[i]) - 1024) * 601) / 1024 + 1500;
     }
 
-    uint16_t len = dronecan_sensors_rc_RCInput_encode(&_p.rc_input, _buf);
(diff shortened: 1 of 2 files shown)


jlpoltrack:
> That is, I guess, my question is, why is "the different sample points caused by using 170 MHz" a true statement, i.e., why do the sampling points have to be different.

The time quanta (tq) is a function of the clock and bitrate, so they can't precisely align.  Some of the calcs are noted in the g4 driver, e.g. 
image

Note that 85 MHz can't support 2 or 4 Mbps due to this limitation as they do not divide cleanly.

jlpoltrack:
> That is, one needs to go through all timers and timings and double check they are not affected.

Timers uses SystemCoreClock, so shouldn't be impacted, right? https://github.com/olliw42/stm32ll-lib/blob/main/src/stdstm32-peripherals.h#L774-L801

The SPI frequency will reduce from 10.625 MHz to 10 MHz: https://github.com/olliw42/stm32ll-lib/blob/main/stdstm32-spi-template.h#L430-L431 and I guess you could enable 40 MHz...

image

PR #391, opened 2026-01-13
Sx remove delays

jlpoltrack:
Remove delays for SX126x and SX128x for functions used in main loop.  Note - LR11xx never had these delays implemented.


jlpoltrack:
One more note to add here - I'm pretty sure mLRS won't be communicating with the sx chips for some time (ms) after these commands.

olliw42:
note to myself: 
olli, look at dev-sx-readframe-simplification

olliw42:
@jlpoltrack @brad112358 @tmcadam 
merging this now
changes in timing can have unexpected consequences, so we need to watch out if that does something bad on essentially all platforms :) esepecially hangs need to be reported

PR #392, opened 2026-01-13
USB Flow Control for STM32

jlpoltrack:
Title. Benefit is that can use reasonable buffers for USB.


olliw42:
did you do that based on the latest main branch? The init stuff seems to be back in usb_init() instead of in MSP Hal init.

PR #394, opened 2026-01-18
MSP Rx Reboot

jlpoltrack:
For future INAV Passthrough functionality, use the MSP Reboot command to put STM32 receivers into system bootloader.

Tested in combination with: https://jlpoltrack.github.io/mLRS-Flasher/inav-passthrough/

Requires: https://github.com/iNavFlight/inav/pull/11256

### DIFF
--- a/mLRS/Common/protocols/msp_protocol.h
+++ b/mLRS/Common/protocols/msp_protocol.h
@@ -153,6 +153,16 @@ typedef struct
 #define MSP_SONAR_ALTITUDE_LEN  4
 
 
+// MSP_REBOOT  68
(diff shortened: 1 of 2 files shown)


b14ckyy:
@olliw42 He pulls the serial config from INAV and lists all ports that have MSP enabled in his web flasher. the User then picks the port where he has the receiver connected to and this port is then enabled for passthrough. So its not fully automatic. 

But I might miss something here. If he found a way to send that command to the Rx then it should work with bind too. 

@jlpoltrack 
Maybe you can explain that for dummies like me :D

PR #396, opened 2026-01-22
Add TX16S MK3

jlpoltrack:
Untested, lua needs update too for 800 x 480 screen.


jlpoltrack:
Tested on 868/915 MHz with DIY LR2021, looks good - so merging.

PR #400, opened 2026-01-24
Run Make ESP Update

jlpoltrack:
Bunch of changes:

- Add support for Linux & Mac
- Add parallel build support, compiles multiple targets about 30% faster
- New list targets and flash arguments
- Documentation for all arguments

Note - first compile can take some time while PIO is getting setup


jlpoltrack:
Fair enough, I will close this one for now.

PR #404, opened 2026-02-03
Linear power levels

jlpoltrack:
Now when using the power switch channel parameters, you get a linear response throughout the PWM range.

Example for Nomad: 

Power = 250 mW, get 10 / 50 / 250
Power = 1000mW, get 10 / 100 / 1000

### DIFF
--- a/mLRS/Common/rf_power.h
+++ b/mLRS/Common/rf_power.h
@@ -76,10 +76,12 @@ void tRfPower::Set(tRcData* const rc, uint8_t power_switch_channel, uint8_t powe
     int16_t rc_val = rc->ch[power_switch_channel + 3]; // ch5 .. ch16 -> 3 .. 15
 
     // linear mapping: divide RC range evenly across available power levels
-    // rc_val: 11 bits, 1 .. 1024 .. 2047 for +-120%
-    // use safe range: ~200 (-100%) to ~1848 (+100%)
+    // rc_val: 11 bits, 1 ... 1024 ... 2047 for +-120%
+    // mLRS uses 172 ... 1024 ... 1876 for +-100%
+    // use safe range: 200 (--98.5%) to 1848 (+98.5%)
     // low stick = low power, high stick = high power
-    // +1 in denominator: range 200..1848 inclusive has 1649 values, ensures max input maps to max index
+    // +1 in denominator: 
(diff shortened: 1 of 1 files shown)


mLRS/Common/rf_power.h
```diff
@@ -73,65 +73,16 @@ void tRfPower::Set(tRcData* const rc, uint8_t power_switch_channel, uint8_t powe
...
-            if (new_idx > (int8_t)power - 3) new_idx = (int8_t)power - 3;
-        } else {                        // slot 4 -> max - 1
-            new_idx = (int8_t)power - 1;
-        }
-    }
+    int16_t rc_val = rc->ch[power_switch_channel + 3]; // ch5 .. ch16 -> 3 .. 15
+
+    // linear mapping: divide RC range evenly across available power levels
+    // rc_val: 11 bits, 1 .. 1024 .. 2047 for +-120%
+    // use safe range: ~200 (-100%) to ~1848 (+100%)
```
olliw42: our range for -100%...+100% is 172 ... 1876  (see https://github.com/olliw42/mLRS/blob/main/mLRS/Common/design_decissions.h#L518-L523)
please chanbge that comment to not misslead
like ~200 (-xx %) to ~1848 (+xx %)

mLRS/Common/rf_power.h
```diff
@@ -73,65 +73,16 @@ void tRfPower::Set(tRcData* const rc, uint8_t power_switch_channel, uint8_t powe
...
+    // low stick = low power, high stick = high power
+    // +1 in denominator: range 200..1848 inclusive has 1649 values, ensures max input maps to max index
```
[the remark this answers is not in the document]
jlpoltrack: On the math, an example with power = 3, so num_levels = 4

Without + 1 and rc_val = 1848

((1848 - 200)) * 4) / (1848 - 200)
(1648 * 4) / (1648) = 4

4 is out of range here, the max power is 3...the constrain will catch anyways

With +1 and rc_val = 1848

((1848 - 200)) * 4) / (1848 - 200 + 1)
(1648 * 4) / (1649) = 3 (truncated)

Correct, as max power is 3

mLRS/Common/rf_power.h
jlpoltrack: Comment updated, yea I chose 200 since that was a nice number on the lower end :)

PR #405, opened 2026-02-10
Minor dissapearing cursor bug fix and some code cleanup for BW Lua

brad112358:
Since the number of cursor index positions is now the same for all pages except the info page, we can eliminate some of the ugly special case code associated with page 0.  Also, reduce the options on the info page to just "Next" which returns to page 0.  Bonus:  This should use just a tad less memory.

Tested and compiled on EdgeTx version 2.11.4.

@jlpoltrack, perhaps you can give this a quick look and test.  Thanks!


jlpoltrack:
@brad112358 this works fine for me on GX12.  One minor suggestion - can you share the EdgeTx version that you used for the compiled version just for reference in the first comment of the PR?

PR #406, opened 2026-02-10
ESP32C3 Bridge Update

jlpoltrack:
Built with latest .ino with 2.0.17 Arduino Core


olliw42:
maybe we uphold a bit
I'm looking at adding a cli option to set ssid/pswd for UDPSTA
unless it doesn't hurt to just merge this when pl LGTM and do (with the README change LOL)

PR #408, opened 2026-02-13
Feat: Add support for BAYCKRC UR500 (ESP32C3)

sunfengxin0:
This PR introduces support for a new hardware target: the BAYCKRC UR500.

Changes:

New Hardware Support: Added rx-hal-bayckrc-ur500-esp32c3.h containing pin definitions and configuration for the BAYCKRC UR500 receiver.
Build System: Updated platformio.ini to include the bayckrc-ur500-esp32c3 environment.

Testing:

Verified compilation for the bayckrc-ur500-esp32c3 target.
Tested on actual hardware to verify startup and basic functionality.
 Successfully tested communication with Nomad on both 2.4GHz and 915MHz bands.

### DIFF
--- a/mLRS/Common/hal/esp/rx-hal-bayckrc-ur500-esp32c3.h
+++ b/mLRS/Common/hal/esp/rx-hal-bayckrc-ur500-esp32c3.h
@@ -28,8 +28,8 @@
 // UR500 overlay: serial1_rx=18, serial1_tx=19
 #define UARTB_USE_SERIAL
 #define UARTB_BAUD                RX_SERIAL_BAUDRATE
-#define UARTB_USE_TX_IO           IO_P19
-#define UARTB_USE_RX_IO           IO_P18
+#define UARTB_USE_TX_IO           IO_P21
+#define UARTB_USE_RX_IO           IO_P20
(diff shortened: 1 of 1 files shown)


sunfengxin0:
The UR500 is Bayck's next-generation receiver with PA, using the LR1121 RF chipset. It has a stable supply in China and is available for purchase on Taobao.Compared with the UR1000, it is lighter, smaller, and of course cheaper.I think this receiver does not cost much more than the Bayck 915M Nano Pro, but has the advantage of supporting 2.4G, so it may become a recommended receiver later on.Of course, we can perhaps wait until Bayck has a stable supply available overseas before considering merging it into the main branch.

olliw42:
> I would like to know whether MLRS is planning to support as many ELRS hardware variants as possible, or only select a subset of reliable ones.

fair question but has been answered many times and should be obvious from the supported devices: 
mLRS does NOT intended to support as many ELRS devices as possible. It (officially) supports only devices which we consider useful to support (what ever that means).
mLRS is flexible enough to allow anyone to adapt it easily to their needs, but we don't consider that our job.

Setting up automated procedures can have their own set of quirks constant issues, and also cost time. We don't do here since we don't feel it compensates our time.

PR #414, opened 2026-02-15
Disp Y Offset

jlpoltrack:
Adds a define for the Y Offset on the OLED Displays.

Does add 1 position tweak for the Rx settings screen when there is no receiver connected - move this down 4 pixels.

Couldn't see any issues with 22 on single color OLEDs but kept at 20, so should be NFC.

### DIFF
--- a/mLRS/CommonTx/disp.h
+++ b/mLRS/CommonTx/disp.h
@@ -54,7 +54,7 @@ void i2c_spin(uint16_t chunksize);
 
 // y-offset for content below header; 20 was original value, use 22 for dual-color OLEDs
 #ifndef DISP_CONTENT_Y_BASE
-  #define DISP_CONTENT_Y_BASE               20
+  #define DISP_CONTENT_Y_BASE   20
 #endif
 
 


mLRS/CommonTx/disp.h
```diff
@@ -956,7 +959,7 @@ void tTxDisp::draw_page_rx(void)
     draw_header("Rx");
 
     if (!connected()) {
-        gdisp_setcurXY(0, 20);
+        gdisp_setcurXY(0, DISP_CONTENT_Y_BASE + 4);
```
olliw42: +4 ?? intentional?

mLRS/CommonTx/disp.h
```diff
@@ -52,6 +52,9 @@ void i2c_spin(uint16_t chunksize);
 #define DISP_START_PAGE_TMO_MS  SYSTICK_DELAY_MS(1500)
 #define KEYS_DEBOUNCE_TMO_MS    SYSTICK_DELAY_MS(40)
 
+// y-offset for content below header; 20 was original value, use 22 for dual-color OLEDs
+#define DISP_CONTENT_Y_BASE     20
```
olliw42: you could wrap it with something like
```
#ifndef DISP_CONTENT_Y_BASE
  #define DISP_CONTENT_Y_BASE               20
#endif
```

jlpoltrack: Sure, done.

PR #418, opened 2026-02-26
ESP UART Better bytesavailable

jlpoltrack:
Small optimization - bytesavailable uses a uint under the hood, so no need to call twice.


jlpoltrack:
> Arduino allows values below 0, and that's why we did that.

Do you have a source for this?  The Arduino core uses uint32_t: https://github.com/espressif/arduino-esp32/blob/release/v2.x/cores/esp32/esp32-hal-uart.h#L39

jlpoltrack:
> If we want to avoid the second function call, I'd suggest just caching the value from the first call in a local variable.

Fair enough, updated.

PR #421, opened 2026-03-05
New Tx Ser Dest - USB

jlpoltrack:
Adds new 'USB' option to Tx Ser Dest for devices with native USB (STM32G4) instead of being locked to CLI.  Only added to mR900-30 at the moment.


jlpoltrack:
Based on discussion, sounds like this will be handled another way, closing.

PR #422, opened 2026-03-06
ESPNOW Sniffer

jlpoltrack:
Listen only sniffer for ESPNOW

CC: @vrquaeler

### DIFF
--- a/esp/mlrs-espnow/espnow_gcs.h
+++ b/esp/mlrs-espnow/espnow_gcs.h
@@ -7,7 +7,7 @@
 // GCS Mode — ESP-NOW API bridge (bidirectional serial <-> ESP-NOW)
 // Supports ESP32, ESP32C3, ESP32S3
 //*******************************************************
-// 7. Mar. 2026
+// 11. Mar. 2026
 //*******************************************************
 #ifndef ESPNOW_GCS_H
 #define ESPNOW_GCS_H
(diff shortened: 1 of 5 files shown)


vrquaeler:
> that's the all-and-only-DIY attitude I think ;) (I was thinking, what if a vendor would offer a epsnow dongle...)

Sorry about that.

> why is it that the gcs and sniffer versions are so massively different?
not understanding espnow so answer might be obvious/trivial, but naively I would have though it would need "just" not sending to become a sniffer?

Technically he ESPNOW is a two-way-protocol with callbacks etc. For the sniffer there have to be ways to circumvent that. 

>  if it couldn't be usefull if the gcs/sniffer could do also other protocols? E.g., I don't have BLE, so with a little dongle could use that. Or, we had that example before, a UDP-sniffer for the AT. Etc. pp.

No first hand experience with  BLE

olliw42:
yo, but as indicated you can keep the list of peers, like we do already for UDP.

did a bit more research. I think we really can combine gcs&sniffer in one code (and make it very similar to txmodules's). The only difference is that for a sniffer we do not connect the esp's rx wire, so that it never receives anything on the serial. So, also no switch needed. In the code all it needs to be done is to regularly send an empty frame if there is no data on the rx serial. In the tx module we just need to do a list of peers, and send to all peers, pretty much as with UDP. Also would nicen up the "latched" thingy. 
Would be way more maintainable, and also more future proof. (I'm convinced one of the first things folks will want is a comnversion from MSP to CRSF or MAVLink)

btw, we need to add some id or bindphrase or such to the payload. To be user firendly we porobably need some bind mode then. Ups, button needed LOL.

PR #425, opened 2026-03-06
ESP PIO Update

jlpoltrack:
Updates the PlatformIO ESP32 framework to the latest.  Updates the underlying Arduino Core from 2.0.14 to 2.0.17 which fixes a bug on dual-core ESP, main loop runs ~60% faster.


jlpoltrack:
IMO no - though should probably be first of the ESP PRs after 1.4 release.

PR #427, opened 2026-03-19
Bridge eFuse Mac

jlpoltrack:
Title, tested on DB30 and TX15 (ESP32C3).

Note - once merged will create a new PR with the ESP32C3 binary.


olliw42:
MANY THX
note, will do a slight rearrange of trhe ifdefs as follow up.

PR #432, opened 2026-03-26
LR2021 DIO RF Config

jlpoltrack:
Was missing the DIO pin, aligned to datasheet:

image


mLRS/Common/sx-drivers/lr20xx.cpp
```diff
@@ -185,11 +185,12 @@ uint8_t buf[2];
 
 void Lr20xxDriverBase::SetDioRfSwitchConfig(uint8_t Dio, uint8_t Config)
 {
-uint8_t buf[1];
+uint8_t buf[2];
 
-    buf[0] = (Config & 0x1F);
+    buf[0] = Dio; // allowed values are 5 - 11
+    buf[1] = (Config);
```
olliw42: pl put back in the & 0x1F

PR #435, opened 2026-03-29
ESPNOW Multipeer

jlpoltrack:
- Adds registration to allow for multiple peers on ESPNOW.
- 'Active' Listener mode via define.
- Timeout logic removed, connection handled via beacons.

@vrquaeler


jlpoltrack:
@vrquaeler Any update?

olliw42:
MANY THX

PR #436, opened 2026-03-30
DroneCAN getbuf

jlpoltrack:
Add getbuf method to DroneCAN.

If merged, will follow-up with update to common and common_types to include getbuf methods.

### DIFF
--- a/mLRS/CommonRx/dronecan_interface_rx.h
+++ b/mLRS/CommonRx/dronecan_interface_rx.h
@@ -473,6 +473,13 @@ uint8_t tRxDroneCan::getc(void)
 }
 
 
+void tRxDroneCan::getbuf(uint8_t* const buf, uint16_t len)
+{
+    tunnel_targetted_stats.fc_to_ser_rate += len;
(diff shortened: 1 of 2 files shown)


jlpoltrack:
> so, what does this really solve

Thought was to read a fixed number of bytes using getbuf instead of reading a byte at a time up to a full message (which improves perf on ESP platform).  If no interest, I will close.

olliw42:
> Thought was to read a fixed number of bytes using getbuf instead of reading a byte at a time up to a full message (which improves perf on ESP platform). If no interest, I will close.

I understand that, but I don't see how it would possibly align with the parsers. which would need then a loop for single byte read themselves, which woul. It's also not clear if the perf improvement is practically relevant or more of theoretical nature here. Since one eventually needs to break the read down to single byte reads anyway, I would think that if there is a perfomance gain it should be at the level of the serial dirver, i.e., e.g., it should fill a buffer with block-wise reads, from which the higher levels than can do the byte-wise reads. The SerialBase class is btw the high-level "driver". I also don't see in which code places one actually would use the block-wise read, can't be many. So, it's not that it'S a matter of interest or not, but I miss some indications of answer to such thoughts :)
 The API here

PR #445, opened 2026-04-08
More LR Driver

jlpoltrack:
- Switches IRQ to be set by define and not hardcoded to  DIO 7
- Error if this define is not set
- Adds method for setting up DIOs
- Power calc to allow for custom levels instead of fixed gain


olliw42:
this I was wondering also about with the previous PR on SX_TCXO_VOLTAGE
this way requires including some sx driver .h files in the hal file, right?
If so, I'm not sure I like that approach. I guess I would prefer providing more detailed defines, such as SX_TCXO_VOLTAGE_3P3V.
Dito e.g. SX_IRQ_DIO_7, plus what you now need

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -8,6 +8,10 @@
 // Configuration defines:
 // #define POWER_USE_DEFAULT_RFPOWER_CALC
 // #define SX_TCXO_VOLTAGE
+// #define SX_DIO_IRQ_NO
```
olliw42: should be SX_USE_IRQ_DIO_NO

<strike>can't add in teh above line, hence here
SX_TXCO_VOLTAGE should become SX_USE_TXCO_VOLTAGE</strike>
I have done that as part of the last tidying PR

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -8,6 +8,10 @@
...
+// #define SX_DIO_RFSW_NUM
```
olliw42: should not be needed

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -8,6 +8,10 @@
...
+// #define SX_DIO_RFSW_DIOS
```
olliw42: should be SX_USE_RFSW_DIO_NOS

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -8,6 +8,10 @@
...
+// #define SX_DIO_RFSW_CONFIGS
```
olliw42: should be SX_USE_RFSW_DIO_CONFIGS

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -330,10 +334,23 @@ class Lr20xxDriverCommon : public Lr20xxDriverBase
...
+        // DIO Setup
+        // IRQ is mandatory, other DIOs depend on the hardware design
+        // DIO 5 requires LR20XX_DIO_SLEEP_PULL_UP, so do for all
+
+        SetDioFunction(SX_DIO_IRQ_NO, LR20XX_DIO_FUNCTION_IRQ, LR20XX_DIO_SLEEP_PULL_UP);
+        SetDioIrqConfig(SX_DIO_IRQ_NO, LR20XX_IRQ_TX_DONE | LR20XX_IRQ_RX_DONE | LR20XX_IRQ_TIMEOUT);
         ClearIrq(LR20XX_IRQ_ALL);
 
+#ifdef SX_DIO_RFSW_NUM
+        const uint8_t lr_dio_rfsw[] = { SX_DIO_RFSW_DIOS };
```
olliw42: in lr11xx driver we use the construct 

const uint8_t lr_dio_rfsw[] = SX_DIO_RFSW_DIOS;

I think I like that better. The SX_DIO_RFSW_NUM should not be needed as you can get if from the size of the array

olliw42:
fyi, have pushed the announced tidying PR, which also changed to SX_USE_TCXO_VOLTAGE
and created the conflicts
sorry for that

jlpoltrack:
> this way requires including some sx driver .h files in the hal file, right?

At least for PlatformIO Targets, no additional includes are needed.  Just use enums from the library files via define, example:

image

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: Done.

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: _NUM removed.

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: Done.

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: Done.

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: Sure, done.

olliw42:
pl try then also for STM32 ... following the include path I'm almost sure it won't work 
(how could it, you need to include the hal before the sx driver as the hal sets defines for the sx driver ... don't underatand how it can work for vsc)

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -330,10 +333,23 @@ class Lr20xxDriverCommon : public Lr20xxDriverBase
...
+        // DIO Setup
+        // IRQ is mandatory, other DIOs depend on the hardware design
+        // DIO 5 requires LR20XX_DIO_SLEEP_PULL_UP, so do for all
+
+        SetDioFunction(SX_USE_DIO_IRQ_NO, LR20XX_DIO_FUNCTION_IRQ, LR20XX_DIO_SLEEP_PULL_UP);
+        SetDioIrqConfig(SX_USE_DIO_IRQ_NO, LR20XX_IRQ_TX_DONE | LR20XX_IRQ_RX_DONE | LR20XX_IRQ_TIMEOUT);
         ClearIrq(LR20XX_IRQ_ALL);
 
+#ifdef SX_USE_DIO_RFSW_DIOS
+        const uint8_t lr_dio_rfsw[] = { SX_USE_DIO_RFSW_DIOS };
```
olliw42: still has the {} ... ;)
compare lr11xx driver: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/sx-drivers/lr11xx_driver.h#L228

mLRS/Common/sx-drivers/lr20xx_driver.h
```diff
@@ -8,6 +8,9 @@
 // Configuration defines:
 // #define POWER_USE_DEFAULT_RFPOWER_CALC
 // #define SX_USE_TCXO_VOLTAGE
+// #define SX_USE_DIO_IRQ_NO
```
olliw42: hm, not exactly as suggested
I'd like the DIO to come after the purpose
SX_USE_IRQ_DIO_NO
SX_USE_RFSW_DIO_NOS
SX_USE_RFSW_DIO_CONFIGS
this aligns with the pattern used elsewhere

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: Done.

mLRS/Common/sx-drivers/lr20xx_driver.h
jlpoltrack: Done.

jlpoltrack:
> pl try then also for STM32 ... following the include path I'm almost sure it won't work

You're probably right, but I don't have any STM32 targets in my branch, so I can't test.

> don't underatand how it can work for vsc

I think since we include the radio submodule like so (line 501):
image

olliw42:
ohhkkaayy ... it only includes the module lib drivers ... 

so, I'm almost sure, won't work on stm32 ...

jlpoltrack:
> so, I'm almost sure, won't work on stm32 ...

Using numbers instead of the enums is an option too - this is done on LR1121 for the rfsw stuff.

olliw42:
> Using numbers instead of the enums is an option too - this is done on LR1121 for the rfsw stuff.

was thinking this too, would work well for the DIO no ... but I guess would be ugly for the configs ...

PR #446, opened 2026-04-08
Esp easy internal tx flash

jlpoltrack:
Can now flash ESP internal modules directly from VS Code without invoking ETX passthrough.py

@brad112358 @tmcadam


olliw42:
this is working also on Win, right?

maybe the .py cpould be placed in a subfolder tools/vscode

jlpoltrack:
> this is working also on Win, right?

Yes, works on Windows.

olliw42:
putting the .py into a subfolder tools/vscode, resonable to you?

jlpoltrack:
> putting the .py into a subfolder tools/vscode, resonable to you?

Sure, I moved to that location.

PR #447, opened 2026-04-08
800x480 Lua

jlpoltrack:
Better use 800x480 space, some photos:

image

image


olliw42:
clearly much better
but I find the layout still somewhat not tidy in several places (like gaps etc)
I got my Mk3 too, I guess I may wnat to play with that before merge

PR #449, opened 2026-04-17
Use sys.executable to invoke dronescan_dsdlc.py

moosss:
On systems where only `python3` is available (not `python`), calling `dronecan_dsdlc.py` directly via `os.system()` relies on the script's shebang (`#!/usr/bin/env python`), which fails with `/usr/bin/env: 'python': No such file or directory`.

**Fix:** Prepend `sys.executable` to the command in `generate_dsdl()` so the same interpreter already running `dronecan_generate_c_library.py` is reused.

**Tested on:** Ubuntu 24.04 (WSL2) where `python3` is installed but `python` is not.


olliw42:
@brad112358 this makes sense, right?

brad112358:
Yes, this will work and should fix the issue.

olliw42:
MANY THX!

PR #453, opened 2026-04-29
CRSF Refresh Adds

jlpoltrack:
Add the new elements to the refresh block.


olliw42:
changed code to array based, became obsolote

PR #456, opened 2026-05-26
Don't require correct serial port selection for wireless bridge flashing

brad112358:
The run time checking here is unnecessary and was problematic when I was trying to flash the wireless bridge on the AX12 before we had a working Lua script.  Removing this removes a step when flashing and should result in fewer user errors.


olliw42:
THX!

PR #455, opened 2026-05-26
Introduce SERIAL_FIXED_BAUD define to lock serial baud

brad112358:
This is needed for the RadioMaster AX12

PR #457, opened 2026-05-26
Add RadioMaster AX12 target

brad112358:
Add support for the RadioMaster AX12 internal Tx module.
~~This depends on #455~~


tmcadam:
I've built and flashed this branch. Not seeing any issues. Able to bind and maintain stable connection (bench), lua works fine and able to connect to QGC app.

mLRS/CommonTx/mlrs-tx.cpp
```diff
@@ -757,7 +761,11 @@ RESTARTCONTROLLER
   #ifdef DEVICE_HAS_ESP_WIFI_BRIDGE_W_PASSTHRU_VIA_JRPIN5
     esp.Init(&jrpin5serial, serial, serial2, Config.SerialBaudrate, &Setup.Tx[Config.ConfigId], &Setup.Common[Config.ConfigId]);
   #else
-    esp.Init(comport, serial, serial2, Config.SerialBaudrate, &Setup.Tx[Config.ConfigId], &Setup.Common[Config.ConfigId]);
+    #ifdef DEVICE_HAS_ESP_WIFI_BRIDGE_W_PASSTHRU_VIA_SERIAL
```
olliw42, line 764: IMHO something went wrong here. There shouldn't be such changes, right?

brad112358: Again, the AX12 is a little different than other handsets.  It has a second serial connection between the MCU and the internal Tx module which is used for telemetry.  Instead of using the CRSF (JRPIN5) serial port for flashing like most internal modules, this second serial port (serial) is used.

As for if this difference should be accounted for here or elsewhere, it seems to me the pattern has already been set.

If this doesn't make sense, you will need to give me more hints about what you think went wrong.

olliw42: IMHO this doesn't make sense. Both comport and serial are handed over in this function, so it can be choosen inside the esp.h library what is wanted for whatever case. So, this here can be as it was.
To me this here also has the smell that it breaks things in pieces. The canonic is that it's on com and either serial or serial2. This choices here don't fit that pattern.
So, pl, keep it as it was, and if the AX12 really needs something special, let's do it in esp.h (if it really needs that),

brad112358: IMHO if this doesn't make sense, then the ifdef that was already here didn't make sense to start with.  The first parameter in tTxEspWifiBridge::Init() is _comport, but if DEVICE_HAS_ESP_WIFI_BRIDGE_W_PASSTHRU_VIA_JRPIN5 is defined, then, &jrpin5serial is passed instead of comport.  This is because this parameter should have been called _passthru, not _comport.  The first parameter is really the port which is used for flashing which is not always comport.  For most internal modules it is JRPIN5 and in the case of the AX12, it is serial.  If you want the logic for which port to use for passthrough to move inside of esp.h, then, the existing ifdef DEVICE_HAS_ESP_WIFI_BRIDGE_W_PASSTHRU_VIA_JRPIN5 should also move inside esp.h and in that case, we need yet another serial parameter for init().

If we need to pass pointers to all the serial ports, why not just directly reference the tSerialBase pointers for serial, comport, serial2, etc directly from common.h instead of passing them all as parameters?

Would you feel better if the second parameter was NULL instead of serial?

This brings up another point which doesn't make sense to me.  Why are we passing both _serialport  and _serial2port to tTxEspWifiBridge::Init() ?  Shouldn't init () just specify which port is the passthrough port and which port the bridge is on?

I think it would have been cleaner if the hal files did:

#define DEVICE_ESP_PORT serial  // serial or serial2
#define DEVICE_ESP_PASSTHROUGH  jrpin5serial // jrpin5serial or comport or serial

instead of:

#define DEVICE_HAS_ESP_WIFI_BRIDGE_ON_SERIAL
#define DEVICE_HAS_ESP_WIFI_BRIDGE_W_PASSTHRU_VIA_JRPIN5

Then, we could just call:
esp.Init(DEVICE_ESP_PASSTHROUGH, DEVICE_ESP_PORT,  Config.SerialBaudrate, &Setup.Tx[Config.ConfigId], &Setup.Common[Config.ConfigId]);

No ifdefs required!

But this PR is not the place for any of this refactoring.

I believe he way I did it here is completely consistent with what has been done before.  I suggest we leave it as I have it and clean this mess up in future PRs.

olliw42:
I don't really understand what your issue is.
I will not say that the current architecture of handling all the serials couldn't be improved (it IMHO won't be so easy though). However, we do have that architecture in many places, and changing it would be a substantial amount of work. We can do that before any AX12 stuff, but my assumption was to get the AX12 in rather sooner than later.
If that assumption should be valid, then:
since all required serials are passed in the Init(), there is no needed to pass it in twice. The job can be done within the esp.h library.

DEVICE_HAS_ESP_WIFI_BRIDGE_W_PASSTHRU_VIA_JRPIN5 is "required" since jrpin5serial is not always available (in contrast to comport). It is correct though that this could be resolved by ensuring that a jrpin5serial is always defined by respective code in other places.

as regards why several serials are passed: Fact is, somewhere in the code the decission needs to be made which serial it should be. So, it's all just about in which place in the code this decission should be made. It looks natural to me to do it in the respective sublibrary.

Your idea for a "better" HAL I don't agree with, IMHO that would make it understandable to just few superheros in this universe (and I wouldn't be among them).

brad112358:
I had given up trying to explain my view point and was about to push the change you suggested but I see you have gotten it done without further delay which is good.  Sorry I missed adding the documentation in hal.h

I guess the next step is new binaries?

Thanks!

olliw42:
@brad112358 I frankly didn't fully understood why that would have needed elaborate discussion at all, so I just did it, to me it looks clean and logical now. 
MANY THX to you and @tmcadam for having spend all that time on the AX12, it's ugly duckness did make it easy.

PR #458, opened 2026-05-26
Lua changes to support Radiomaster AX12

brad112358:
I've added a partial workaround for the missing getRSSI() function.  The workaround won't discover when the receiver connects or disconnects, but it works in the static case and should suffice until Radiomaster implements getRSSI().


tmcadam:
I've given your version a test and it seems to working fine. 

Only issue I can see is the parameter counter is a bit garbled during load/reload. It's pretty minor and not sure if it's worth looking into. 

image

brad112358:
I've noticed that as well.  Unless you see something obvious we are doing to cause this, I'm inclined to wait until firmware is available and then report this to Radiomaster as a bug on their end.

olliw42:
MANY THX.
I will merge and then do few tiny nfc tides.

PR #461, opened 2026-06-04
tools, run_make_firmwares, arm-none-eabi toolchain on PATH (no CubeIDE)

jlpoltrack:
Allows one to use arm toolchain instead of requiring CubeIDE.

Tested on MacOS only.

### DIFF
--- a/tools/run_make_firmwares.py
+++ b/tools/run_make_firmwares.py
@@ -9,7 +9,7 @@
  run_make_firmwares.py
  3rd version, doesn't use make but calls gnu directly
  gave up on cmake, hence naive by hand
- version 7.05.2026
+ version 5.06.2026
 ********************************************************
 '''
 import os
--- a/tools/run_make_firmwares.py
+++ b/tools/run_make_firmwares.py
@@ -78,7 +78,7 @@ def findSTM32CubeIDEGnuTools(search_root):
     return st_dir, gnu_dir
 
 
-ST_DIR,GNU_DIR = '', ''
+ST_DIR,GNU_DIR,GCC_DIR = '', '', ''
 
 # do this only when called from main context
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


tools/run_make_firmwares.py
```diff
@@ -93,17 +94,24 @@ def findSTM32CubeIDEGnuTools(search_root):
...
+    else:
+        print('STM32CubeIDE found in:', ST_DIR)
+        print('gnu-tools found in:', GNU_DIR)
     print('------------------------------------------------------------')
 
 
 #-- GCC preliminaries
 
-GCC_DIR = os.path.join(ST_DIR,GNU_DIR,'tools','bin')
+GCC_DIR = GCC_DIR_OVERRIDE if GCC_DIR_OVERRIDE else os.path.join(ST_DIR,GNU_DIR,'tools','bin')
```
olliw42: I think this line should go, ST_DIR,GNU_DIR could be set in the above as needed

tools/run_make_firmwares.py
```diff
@@ -79,6 +79,7 @@ def findSTM32CubeIDEGnuTools(search_root):
 
 
 ST_DIR,GNU_DIR = '', ''
+GCC_DIR_OVERRIDE = '' # set when a standalone arm-none-eabi toolchain on PATH is used instead of CubeIDE
```
olliw42: I think this can go, could be all done in the __main__ code path, by setting ST_DIR,GNU_DIR as needed

tools/run_make_firmwares.py
```diff
@@ -93,17 +94,24 @@ def findSTM32CubeIDEGnuTools(search_root):
...
+        gcc_path = shutil.which('arm-none-eabi-gcc')
+        if gcc_path:
+            GCC_DIR_OVERRIDE = os.path.dirname(gcc_path)
+            print('STM32CubeIDE not found, using arm-none-eabi toolchain on PATH:')
+            print(GCC_DIR_OVERRIDE)
+        else:
+            print('ERROR: gnu-tools not found! (neither STM32CubeIDE nor arm-none-eabi-gcc on PATH)')
+            exit(1)
+    else:
+        print('STM32CubeIDE found in:', ST_DIR)
```
olliw42: if all done in here and ST_DIR,GNU_DIR set as needed, can go out of the if again as it was before

olliw42:
I think the featuire can be achieved with less changes to the original code.
If I'm mistaken, pl explain :)

tools/run_make_firmwares.py
jlpoltrack: No more GCC_DIR_OVERRIDE

tools/run_make_firmwares.py
jlpoltrack: No more GCC_DIR_OVERRIDE

tools/run_make_firmwares.py
jlpoltrack: Done

tools/run_make_firmwares.py
```diff
@@ -103,7 +110,7 @@ def findSTM32CubeIDEGnuTools(search_root):
 
 #-- GCC preliminaries
 
-GCC_DIR = os.path.join(ST_DIR,GNU_DIR,'tools','bin')
+GCC_DIR = os.path.join(ST_DIR,GNU_DIR,'tools','bin') if GNU_DIR else ST_DIR
```
olliw42: still don't see why that is needed
ST_DIR and GNU_DIR should be setable such that this line as it was gives the desired outcome...
not?

jlpoltrack: Well the toolchain doesn't need the tools/bin as part of the path... This was the reason for the GCC_DIR_OVERRIDE in the initial try.

olliw42: ah ... ok. So what we actually should do is to set GCC_DIR in the __main__ sectionpart
That is, the line GCC_DIR = os.path.join(ST_DIR,GNU_DIR,'tools','bin') shoudl also go into the __main__ part, and the outcome should be  GCC_DIR. 
I guess, befiore that we should have GCC_DIR = "".
Does this make sense?

jlpoltrack: Does this mean that I should get rid of the entire GCC preliminaries section as well? It would no longer be used and only contain comments.

olliw42:
THX!

PR #462, opened 2026-06-05
tools, run_make_firmwares, parallel builds

jlpoltrack:
Compile multiple files using subprocess (uses the number of CPU cores on the machine), results in a 2.5x speedup on my machine.

### DIFF
--- a/tools/run_make_firmwares.py
+++ b/tools/run_make_firmwares.py
@@ -17,6 +17,8 @@ import pathlib
 import shutil
 import re
 import sys
+import subprocess
+from concurrent.futures import ThreadPoolExecutor
 
 
 #-- installation dependent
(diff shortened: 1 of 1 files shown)


olliw42:
interesting, but I'm somewhat disappointed by that it only brings a factor of 2.5x.
Have you tried ProcessPoolExecutor instead of ThreadPoolExecutor? It could be faster in this case.
Also, why not placing all os.systems() file into a list, and doing it for all of them? Would include also linking (and might make the code very slim)

jlpoltrack:
- Process would require some additional code and I don't think it would result in extra performance
- Parallel targets (in addition) could also be explored but this brings more complexity as well

2.5x is just my machine, would be good if you can confirm it is better/worse/similar on yours.

olliw42:
not usre that code argument is correct, but anyway...
tested it and on my machine it's 2.5-3x faster.
do we have some sort of garantee it's producing the same code as the original?
(code can depend on the sequence of compilation, and we had cases there this matters, e.g. old mavX issues)

jlpoltrack:
> do we have some sort of garantee it's producing the same code as the original?

As part of the work I checked that .hex files produced were identical between main and this branch.

PR #463, opened 2026-06-07
tools, new flash_stm32

jlpoltrack:
New tool that allows one to build + flash STM32 targets from command line, so CubeIDE + CubeProgrammer not necessarily needed for development.

- Uses existing make_firmware to build
- Can flash with DFU or SWD with arg

### DIFF
--- a/tools/run_flash_stm32.py
+++ b/tools/run_flash_stm32.py
@@ -1,0 +1,163 @@
+#!/usr/bin/env python
+'''
+*******************************************************
+ Copyright (c) MLRS project
+ GPL3
+ https://www.gnu.org/licenses/gpl-3.0.de.html
+ OlliW @ www.olliw.eu
+*******************************************************
+ run_flash_stm32.py
+ requires dfu-util (DFU) or st-flash (SWD)
+ version 7.06.2026
+********************************************************
+'''
+import os
+import sys
+import shutil
+import glob
+import argparse
+import subprocess
+
+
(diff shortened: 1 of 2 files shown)


tmcadam:
I guess this could open the way to building binaries automatically in github action, on merge to main. 

This is a very outdated PR that was doing something like that for the esp devices 
https://github.com/olliw42/mLRS/pull/155

olliw42:
this would work only on linux, won't work on Win, and I'm very much against suggesting Win users having to install stuff which needs that Zadig nonsense. So, Win users would be back to STM32CubeProgrammer ... and having to install something anyways.
So, either it's extended to use STM32CubeProgrammer or made clear to be lin only by e.g. changing the name to "..._linux.py".

as regards automatic builds, this here won't help as it's mainly about adding flashing. I'm also somehwat hesistant with auto builds. It seems many folks use dev builds, so I guess many would then use nightlies ... which may or may not work. A controlled way for dev builds looks reasonable to me. Just my 2 cents though.

jlpoltrack:
> So, either it's extended to use STM32CubeProgrammer or made clear to be lin only by e.g. changing the name to "..._linux.py".

I chose the rename option, if there's need in the future I can look into extending for STM32CubeProgrammer.  Note - all of my tests have been on MacOS.

olliw42:
will merge, but change the name to _linux. _nix is a bit nerdy IMHO ;)

PR #464, opened 2026-06-08
tools, run_make_firmware, multi target build

jlpoltrack:
More of a proposal:

- Builds 4 targets at once (was the best on 10 core Mac and 8 core Win machine)
- Measured 1.65x speedup
- Limitation - the prints get intertwined and look a bit messy


olliw42:
I strongly think that this is going into a wrong direction. Your 3rd bullet confirms that, but not only that. 
I think if one want's to get further improvement, the principle functions need to be changed substantially. The fact that this here gives some speedup in fact demonstrated that already the previous approach was anything than ideal.
So, closing it. If you don't agree pl feel free to reopen.

PR #471, opened 2026-06-27
Serial routing updates

jlpoltrack:
Enable TxSerDest2 on ESP32
Filter TxSerDest/TxSerDest2 lists based on setting of other

Remove duplicated ChannelOrder SANITIZE line 
Catch a Dest1/Dest2 conflict when one is wbridge and the other is serial2, since both share uartD

Fix TxSerDest2=serial when TxSerDest=mbridge by parking the unused serial on uartD so serial2 owns uartB
Fix bridge being held in reset when TxSerDest2=wbridge and enable autoconfig as well

Various comment updates


olliw42:
many thx, have taken over the essential parts

PR #481, opened 2026-07-23
Run_Make Always Generates Objects

jlpoltrack:
For discussion.  Now that one can compile without CubeIDE, avoids having to deal with stale objects.list


olliw42:
I actually wanted it to do that on purpose (we can discuss elsewhere why). 
If that's a problem for without CubeIDE, it should change behavior only for that case.
Closing, pl fell free to reopen.

PR #484, opened 2026-07-27
ESP JRPIn5 MCU Support

jlpoltrack:
Adds support for JRPin5 to ESP32-S3 and ESP32-C3 MCUs.

ESP32-C3 and ESP32-S3 validated on ELRS receivers.


olliw42:
before merging, could you please indicate what this solves, what it is needed for. Just for my education.

jlpoltrack:
> could you please indicate what this solves, what it is needed for

Prep for any future ESP32-S3 Tx Modules with half-duplex JRPin5.  Today, all ESP32-S3 we support are full-duplex.

PR #486, opened 2026-08-01
MSP: use BOXIDS instead of BOXNAMES for INAV flight mode detection

b14ckyy:
While digging into the reports of missing flight modes on the radio I found that the rx never gets its box table on many INAV setups: INAV refuses to answer MSP_BOXNAMES when the reply does not fit its MSP output buffer, which is easily the case on INAV 8/9 once GPS and nav modes are enabled (49 active boxes = 543 bytes on my bench). Without BOXNAMES the rx never sends MSPX_STATUS and the FM field stays dead. That buffer limit is an INAV bug and I will report it there, but it made me question why we depend on BOXNAMES at all.

This switches the rx-internal box fetching to MSP_BOXIDS. The inavBoxes[] table gets a permanentId column (from fc_msp_box.c) and the rx builds inav_flight_modes_box_mode_flags[] from the BOXIDS reply, position in the reply = bit position in MSP2_INAV_STATUS, same as before with the names.

Why I think BOXIDS is the better fit, independent of the INAV bug:

- BOXNAMES never was the version-independent source of truth it looks like. We match the names against our own baked-in table anyway, unknown names are dropped. The only dynamic information we actually need is which box sits on which bit position, and BOXIDS delivers exactly that.
- permanentIds are frozen by INAV by contract, that is what they exist for. Names have been renamed in the past and will be again.
- The reply is ~50 bytes instead of 340-650, so no compression needed on the internal path and no big buffer on the rx stack.
- The fetch pattern is unchanged: requested once after connect, then disabled, re-armed on disconnect, exactly like BOXNAMES before. The mapping is just cleared before it is rebuilt, so a changed FC config cannot leave stale entries behind after a reconnect.
- This is also what the rest of the ecosystem does: INAV Configurator only requests BOXIDS and resolves names from its own permanentId table (BOXNAMES does not appear in their code at all), Ardudeck does BOXIDS + static lookup for INAV with the explicit comment that the BOXNAMES reply is too large, and mwptools works fine from BOXIDS alone and only logs an error if BOXNAMES fails.

The BOXNAMES passthrough for a GCS on the ground is untouched, requests are still forwarded and the reply still gets compressed over the link, mwptools keeps working. Only the flight mode mapping no longer depends on it.

Bench tested against INAV 9.1.0 (Matek F765, mR900-22 rx, mR900-30 tx module): with GPS enabled BOXNAMES still fails FC-side but flight modes now show correctly via BOXIDS. Draft for now, I am still testing against other INAV versions and in the simulator.

### DIFF
--- a/mLRS/CommonRx/msp_interface_rx.h
+++ b/mLRS/CommonRx/msp_interface_rx.h
@@ -78,7 +78,7 @@ class tRxMsp
     uint32_t tick_tlast_ms;
 
     #define MSP_TELM_COUNT  6
-    #define MSP_TELM_BOXNAMES_ID  5
+    #define MSP_TELM_BOXIDS_ID  5
 
     const uint16_t telm_function[MSP_TELM_COUNT] = {
         MSP2_INAV_STATUS,
@@ -86,7 +86,7 @@ class tRxMsp
         MSP2_INAV_ANALOG,
         MSP_RAW_GPS,
         MSP_ALTITUDE,
-        MSP_BOXNAMES, // MSP_BOXIDS, seems to hold incorrect flags ??
+        MSP_BOXIDS, // permanent box ids, stable across INAV versions, much smaller reply than MSP_BOXNAMES
     };
 
     const uint8_t telm_freq[MSP_TELM_COUNT] = {
@@ -95,7 +95,7 @@ class tRxMsp
         1,  // 1 Hz = 10*100 ms, MSP_INAV_ANALOG
         2,  // 2 Hz = 5*100 ms, MSP_RAW_GPS
         2,  // 2 Hz = 5*100 ms, MSP_ALTITUDE
-        1,  // this is set to zero once MSP_BOXNAMES has been gotten once, disables request
+        1,  // this is set to zero once MSP_BOXIDS has been gotten once, disables request
     };
 
     typedef struct {
@@ -107,7 +107,7 @@ class tRxMsp
 
     void telm_set_default_rate(uint8_t n) { telm[n].rate = (telm_freq[n] > 0) ? 10 / telm_freq[n] : 0; }
 
-    uint8_t inav_flight_modes_box_mode_flags[INAV_FLIGHT_MODES_COUNT]; // store info from MSP_BOXNAMES
+    uint8_t inav_flight_modes_box_mode_flags[INAV_FLIGHT_MODES_COUNT]; // store info from MSP_BOXIDS
 
     // to handle command MSP_REBOOT
     uint32_t reboot_activate_ms;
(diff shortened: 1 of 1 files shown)


olliw42:
many thx!
I did some stuff in the meantime myself, which is why there is the conflict. You'll need to update.

olliw42:
on first view, I think your procedure for reading out BOXIDS is not correct, I would think you need to loop through all since it's a mask, not stop after the first found

b14ckyy:
> on first view, I think your procedure for reading out BOXIDS is not correct, I would think you need to loop through all since it's a mask, not stop after the first found

the BOXIDS reply is not a mask, it's a plain list. INAV's `serializeBoxReply()` in `fc_msp_box.c` writes one byte per active box, each byte being that box's permanentId:

```
void serializeBoxReply(sbuf_t *dst)
{
    for (int i = 0; i < activeBoxIdCount; i++) {
        const box_t *box = findBoxByActiveBoxId(activeBoxIds[i]);
        if (!box) {
            continue;
        }
        sbufWriteU8(dst, box->permanentId);
    }
}
```

So the position of a byte in the reply is the box's bit position in the INAV_STATUS box mode flags, and the byte value is the permanentId. The mask thing you have in mind is probably MSP_ACTIVEBOXES (113), that one is indeed a bitmask.

The `break` in my loop only ends the inner lookup over our inavBoxes[] table once the permanentId for the current reply byte is found. The outer loop still walks over every byte of the reply, so nothing is skipped.

I also sniffed the raw reply on the bench to be sure: with my test config the FC answers with 49 bytes for 49 active boxes, e.g. 00 33 3d 01 02 23 05 08 ... = ARM, PREARM, MULTI FUNCTION, ANGLE, HORIZON, ... which matches the BOXNAMES order exactly. Flight modes come up correctly with it, tested armed/disarmed and mode switching.

Fun fact: the configurator does the same thing, it only requests BOXIDS and resolves the names from its own permanentId table, BOXNAMES isn't used there at all.

olliw42:
thx, didn't realized it's just the inner loop
I guess we want to think about the i and n, there is some sort of loose convention, will check

b14ckyy:
> thx, didn't realized it's just the inner loop I guess we want to think about the i and n, there is some sort of loose convention, will check

checked against the existing code: the new loop follows the same pattern as `mspX_boxnames_payload_compress()` — outer `i` over the payload bytes, inner `n` over the `inavBoxes[]` table, same types. So it should already match the convention, but happy to rename if you prefer something else.

olliw42:
the boxnames part should go away in both the rx and tx code, I think
unless there is a higher-level reason to keep

technically, the BOXID when requested from the receiver should be absorbed by the receiver and not send to the tx, but I guess given it's relatively small and stopped soon anyway, we may accept that
but should be just one additional line of code, so maybe worth it

olliw42:
> checked against the existing code: the new loop follows the same pattern as `mspX_boxnames_payload_compress()` — outer `i` over the payload bytes, inner `n` over the `inavBoxes[]` table, same types. So it should already match the convention, but happy to rename if you prefer something else.

fair enough

olliw42:
ok, mspX_boxnames_payload_compress and decompress need to also be changed such that the box_flags thing goes away.

PR #492, opened 2026-08-06
Passthrough RPM Wind

jlpoltrack:
Add RPM & Wind passthrough for Yaapu.

Very windy where I fly :)

### DIFF
--- a/mLRS/Common/protocols/passthrough_protocol.h
+++ b/mLRS/Common/protocols/passthrough_protocol.h
@@ -25,7 +25,8 @@
   TERRAIN_REPORT        #136  -> 0x500B TERRAIN                           -
   BATTERY_STATUS        #147  -> 0x5003 BATT_1, 0x5008 BATT_2             SRx_EXTRA3
   FENCE_STATUS          #162  -> 0x5001 AP_STATUS                         SRx_EXT_STAT
-  // #173, #181, #226 are ArduPilot dialect specific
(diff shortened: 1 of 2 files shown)


mLRS/Common/protocols/passthrough_protocol.h
```diff
@@ -36,7 +36,7 @@
   EXTENDED_SYS_STATE    #245                                              -
 
   // #168 is ArduPilot dialect specific
-  WIND                  #168                                              SRx_EXTRA3
+  WIND                  #168  -> 0x500C WIND                              SRx_EXTRA3
```
olliw42: could you please move that to line 28, and start it with // #168, #173, ....

olliw42:
just one nitpick :)

olliw42:
maybe worth checking if WIND is still ArduPilot dialect specific, and/or if there is maybe a newer common dialect version

mLRS/Common/protocols/passthrough_protocol.h
jlpoltrack: Done

jlpoltrack:
> maybe worth checking if WIND is still ArduPilot dialect specific, and/or if there is maybe a newer common dialect version

WIND is ArduPilot specifc, there is https://mavlink.io/zh/messages/common.html#WIND_COV but not used by ArduPilot.

Looks like RPM has been moved somewhat recently, so small update there.

PR #493, opened 2026-08-06
Target Insensitive in Tool Scripts

jlpoltrack:
Since the targets seem a bit inconsistent IMO and I always forget, just remove case sensitivity for the target argument in the tool scripts.

Examples:

- R9M (with caps)
- Wio (with caps)
- frsky (without caps)

### DIFF
--- a/tools/run_flash_stm32_linux.py
+++ b/tools/run_flash_stm32_linux.py
@@ -46,10 +46,9 @@ def find_firmware_hex(target):
     # .hex files in tools/build/firmware/ whose name contains the target substring (case insensitive)
     if not os.path.isdir(MLRS_FIRMWARE_DIR):
         return []
-    target = target.lower()
     hexes = []
     for path in sorted(glob.glob(os.path.join(MLRS_FIRMWARE_DIR, '*.hex'))):
-        if target in os.path.basename(path).lower():
+        if target.lower() in os.path.basename(path).lower():
             hexes.append(path)
     return hexes
 
--- a/tools/run_flash_stm32_linux.py
+++ b/tools/run_flash_stm32_linux.py
@@ -43,12 +43,13 @@ def build_target(target, defines):
 
 
 def find_firmware_hex(target):
-    # .hex files in tools/build/firmware/ whose name contains the target substring
+    # .hex files in tools/build/firmware/ whose name contains the target substring (case insensitive)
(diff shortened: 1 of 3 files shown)


tools/run_flash_stm32_linux.py
```diff
@@ -43,12 +43,13 @@ def build_target(target, defines):
...
 def find_firmware_hex(target):
-    # .hex files in tools/build/firmware/ whose name contains the target substring
+    # .hex files in tools/build/firmware/ whose name contains the target substring (case insensitive)
     if not os.path.isdir(MLRS_FIRMWARE_DIR):
         return []
+    target = target.lower()
     hexes = []
     for path in sorted(glob.glob(os.path.join(MLRS_FIRMWARE_DIR, '*.hex'))):
-        if target in os.path.basename(path):
+        if target in os.path.basename(path).lower():
```
olliw42: it's nitpick but I'm not so happy with the different naming here and below (where it's target_name)
I think we can avoid this here actually by removing line 49 and writing here
`if taget.lower() in os.poath.basename.lower():`
I would find that cleaner

tools/run_make_firmwares.py
```diff
@@ -1268,9 +1268,10 @@ def mlrs_copy_all_hex_etc():
 
     target_cnt = 0
     for target in targetlist:
+        target_name = target.target.lower()
         if ((cmdline_target == '') or
-            (cmdline_target[0] != '!' and cmdline_target in target.target) or
-            (cmdline_target[0] == '!' and not cmdline_target[1:] in target.target)):
+            (cmdline_target[0] != '!' and cmdline_target in target_name) or
```
olliw42: it's nitpick, but I'm not so happy with the naming.
I could imagine either (a) removing line 1271 and writing target.target.lower() (yes it's twice, but hey, we don't have to runtime optimize) or (b) maybe calling it taget_lower instead of target_name

olliw42:
I really like that, why didn't we do this since ever LOL
two nasty nitpicks though :)

tools/run_flash_stm32_linux.py
jlpoltrack: Done

tools/run_make_firmwares.py
jlpoltrack: Done

olliw42:
many thx!

PR #495, opened 2026-08-10
Bayck UR1000 Support

jlpoltrack:
Title, as requested by a few users.

Pending feedback from user.


mLRS/Common/hal/esp/esp-hal.h
```diff
@@ -103,6 +103,10 @@
 #include "rx-hal-flysky-pr02-2400-td-esp32s3.h"
 #endif
 
+#ifdef RX_BAYCK_UR1000_LR1121_ESP32C3
```
olliw42: first off, I guess should be `RX_ELRS_BAYCK_UR1000_LR1121_ESP32C3`
(and dito `RX_ELRS_FLYSKY_PR02_2400_TD_ESP32S3`)
second, I think it should not be located under "selected devices"
(dito the FLYSKY)
if I recall correctly "selected devices" meant that they are recommended by us, which i think we can't say for the UR1000 as nobody of us has really tested it well...
i.e., I think we shoudl higher up add a section
`//-- Generic LR1121 ELRS boards with overlays or other variations`
and locate the UR1000 there. 
The FlySky shoudl then located in the section `//-- Generic 2.4 GHz ELRS boards with overlays or other variations` where the iflights are

this also should parallel in platformio

platformio.ini
```diff
@@ -350,6 +350,19 @@ build_flags =
   -D RX_FLYSKY_PR02_2400_TD_ESP32S3
 
 
+[env:rx-bayck-ur1000]
```
olliw42: see above

olliw42:
I can't comment on the esp hal. Not sure we want @tmcadam to also have a look, your call.

mLRS/Common/hal/esp/esp-hal.h
jlpoltrack: Done

platformio.ini
jlpoltrack: Done

jlpoltrack:
I will move this out of Draft status but think we sit on for at least a few days.

PR #497, opened 2026-08-12
Tools Flasher, Use CubeProg if avail

jlpoltrack:
Prefer CubeProgrammer CLI if available, reasons:

- Newer families (e.g. H5 and C5) are not fully supported by st-flash meaning one can't use SWD with those MCUs
- dfu-util is noticeably slower when flashing compared to the CubeProgrammerCLI

Tested on Mac.  @brad112358 you want to confirm on Linux?


brad112358:
Yes, this works as expected for me with and without CubeProg installed in my home directory on Linux.

BTW, I saw that newer GCC versions were enabled in a different commit which didn't change this script.  I noticed this script is still ignoring newer GCC, so I guess this file needs a change for newer GCC as well.

jlpoltrack:
> BTW, I saw that newer GCC versions were enabled in a different commit which didn't change this script. I noticed this script is still ignoring newer GCC, so I guess this file needs a change for newer GCC as well.

Not sure I follow, this script just calls run_make which in turn looks for arm-none-eabi-gcc on path (https://github.com/olliw42/mLRS/blob/main/tools/run_make_firmwares.py#L104).

I've been using with 15.3.1 (default now in Homebrew) without any issues.

brad112358:
Oops.  Yes, sorry, it's tools/run_make_firmwares.py which still needs to change.

olliw42:
pl undo the gcc > =12 change, just did it also

otherwise, this is good to merge? 2x LGTM for you?

brad112358:
LGTM

jlpoltrack:
Did the revert, so good to merge.

PR #498, opened 2026-08-12
Fix Hiding Warning

jlpoltrack:
To address `warning: 'virtual void tSerialBase::Init()' was hidden [-Woverloaded-virtual=]` recent compilers (e.g GCC 15.3.1) show.

### DIFF
--- a/mLRS/CommonTx/mbridge_interface.h
+++ b/mLRS/CommonTx/mbridge_interface.h
@@ -38,6 +38,7 @@ typedef enum {
 class tMBridge : public tPin5BridgeBase, public tSerialBase
 {
   public:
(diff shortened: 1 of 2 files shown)


olliw42:
I don't understand this, and something looks dodgy here, at least with mbridge it's strange it seems to complain only aboiut SreialBase init and not also tPin5BridgeBase. I think this needs a bit more investigation to better understand what's up.

jlpoltrack:
Here's the full output: 
image

So tPin5BridgeBase doesn't have a virtual Init like tSerialBase does: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/common_types.h#L68 which causes the warning. (A bit over my head)

olliw42:
ah, got it. MANY THX, Let's see what we should do about it.

PR #500, opened 2026-08-15
Fix WS2812 Color Flicker

jlpoltrack:
- Defer an update instead of aborting an in-flight frame
- Add 300 uS latch gap

Assumes that power_supply_set_leds will be called every frame.


olliw42:
many thx. I however think that's a typical AI - äh - wanabeintelligent "solution"
besides not stopping a dma transmission before it's time another issue is also that one should not allow filling the buf during a dma transmission !! It also failed to realize that do and trigger are actually not used in the main code LOL
the proper solution would be to double buffer
however, I hope with the latest changes the flickering should be gone. If not, we need to do it properly.
closing

PR #501, opened 2026-08-17
Add BAYCKRC UR500 & UR1000 (ESP32-C3 + LR1121) as Tx module targets

zws1023:
## Summary

Add support for using **BAYCKRC UR500** and **UR1000** dual-band receivers as MLRS **Tx (JR-bay) modules**.

Both boards are ESP32-C3 + LR1121 based receivers sold by BAYCKRC, sharing the same layout file ("Generic C3 LR1121.json"). They differ only in the external PA power table.

### Changes

**UR500** (commit 1):
- New HAL `tx-hal-bayck-ur500-esp32c3.h` — 3..27 dBm (2..500 mW)
- Register `TX_BAYCK_UR500_ESP32C3` in `esp-device_conf.h` and `esp-hal.h`
- Add `env_common_tx_esp32c3` and `tx-bayck-ur500` build target to `platformio.ini`

**UR1000** (commit 2):
- New HAL `tx-hal-bayck-ur1000-esp32c3.h` — 3..30 dBm (2..1000 mW)
- Register `TX_BAYCK_UR1000_ESP32C3` in `esp-device_conf.h` and `esp-hal.h`
- Add `tx-bayck-ur1000` build target to `platformio.ini`

### Pinout (identical for both, from ExpressLRS ur500/ur1000 overlay)
- LR1121 SPI: NSS=7 / MISO=5 / MOSI=4 / SCK=6 / RST=2 / DIO=1 / BUSY=3
- RGB LED=8, button=9
- JR bay pin5 (CRSF) -> GPIO18 (half-duplex, 400k)
- Serial port (MAVLink) -> GPIO20/21 (UARTB, 115200 baud default)
- RFSW_CTRL: {15, 0, 4, 12, 0, 2, 0, 1}

### Testing
- Both compile cleanly with PlatformIO (`tx-bayck-ur500`, `tx-bayck-ur1000`).
- Firmware reports device names `BAYCK UR500` / `BAYCK UR1000` (verified in built ELF).

Pre-built binaries: https://github.com/zws1023/mLRS/releases

/cc @olliw42

### DIFF
--- a/mLRS/Common/hal/esp/esp-device_conf.h
+++ b/mLRS/Common/hal/esp/esp-device_conf.h
@@ -226,6 +226,15 @@
   #define FREQUENCY_BAND_915_MHZ_FCC
 #endif
 
+#ifdef TX_BAYCK_UR500_ESP32C3 // receiver used as Tx module
+  #define DEVICE_NAME "BAYCK UR500"
+  #define DEVICE_IS_TRANSMITTER
+  #define DEVICE_HAS_LR11xx
+  #define FREQUENCY_BAND_2P4_GHZ
+  #define FREQUENCY_BAND_868_MHZ
+  #define FREQUENCY_BAND_915_MHZ_FCC
+#endif
+
 
 //-- ELRS Internal Tx Modules
 
--- a/mLRS/Common/hal/esp/esp-device_conf.h
+++ b/mLRS/Common/hal/esp/esp-device_conf.h
@@ -235,6 +235,15 @@
   #define FREQUENCY_BAND_915_MHZ_FCC
 #endif
 
+#ifdef TX_BAYCK_UR1000_ESP32C3 // receiver used as Tx module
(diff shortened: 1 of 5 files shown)


olliw42:
@zws1023 many thx for your effort and contribution, much appreciated.
However, we unfortunately cannot accept this as such. First, given the UR1000 exists, the UR500 is not worth the consideration. Second, and most importantly, this does not have a COM nor a JRpin5 nor an OLed, so it simply cannot be configured and thus is totally useless. Third, I don't see any evidence for this having been tested extensively in practice. In fact, I'd say, if you would have tried what you've created you immediately would have stumbled over the 2nd point. We do not accept unvalidated code, by principle.
Hence, I reject and close.
The UR1000 does have a USB port which could provide the missing COM functionality, making it potentiaklly intersting as a simple Tx device, but the code needs to be done.

PR #503, opened 2026-08-22
GCC 16 Warning Fix

jlpoltrack:
A heads up - GCC 16 produces the following warning:

image

This PR only touches the structs that produce the warning as there are many others throughout the code that don't produce the warning.  Unfortunately, this results in inconsistent formats.


olliw42:
hmm ... isn't there a way to set a compiler flag to disable that warning?

PR #504, opened 2026-08-31
ESP Bridge Bug

jlpoltrack:
This define went away in the 'hal, wirelessbridge, simplify defines' commit and was preventing the Bridge from being available for use / flashing.

### DIFF
--- a/mLRS/CommonTx/esp-wifi-bridge.h
+++ b/mLRS/CommonTx/esp-wifi-bridge.h
@@ -163,10 +163,17 @@ void tTxEspWifiBridge::Init(void)
 #else
     com = Serials.com;
 #endif
-    ser = Serials.uartd; // uartd is always available, but could be a dummy port
-    ser_baud = Config.SerialBaudrate; // TODO: why not always 115200 ?
+    // this code is compiled only when USE_ESP_WIFI_BRIDGE, which implies USE_WIRELESS_BRIDGE
+    // and hence USE_SERIAL2, so uartd is always a real port here and never nullptr
+    ser = Serials.uartd;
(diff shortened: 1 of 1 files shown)


olliw42:
uuuupppssss ... THX

it should then be however protected by the same #ifdef as uartd is, i.e., by #ifdef USE_SERIAL2 // DEVICE_HAS_SERIAL2 || USE_WIRELESS_BRIDGE, since otherwise it would always be != nullptr. Shouldn't it?

jlpoltrack:
The Init will only run based on this in the Hal, so seemed repetitive: https://github.com/olliw42/mLRS/blob/main/mLRS/Common/hal/hal.h#L296-L303

olliw42:
ohhhkkkaayyyy ... this looks true
in that case one however could also remove the ser nullptr test, right
(I guess a comment which reminds that this code is called anyway only when serial2 is available might be helpfull)

olliw42:
also, isn't ` ser_baud = Config.SerialBaudrate; // TODO: why not always 115200 ? ` then a bug too, souldn't it then be Config.SerialBauxdrate2 ??? 

something is odd here ... 

EDIT: shouldn't it be something like

```
if (Setup.Tx[Config.ConfigId].SerialPort2 == TX_SERIAL_PORT2_WIRELESS_BRIDGE) 
  ser_baud = Config.SerialBaudrate2
else 
  ser_baud = Config.SerialBaudrate  
```

jlpoltrack:
- Comment added
- Test removed
- Updated baud logic

olliw42:
THANKS

PR #506, opened 2026-09-07
Add ESP Powerup Counter

jlpoltrack:
Title

### DIFF
--- a/mLRS/Common/hal/esp-powerup.h
+++ b/mLRS/Common/hal/esp-powerup.h
@@ -22,6 +22,103 @@ typedef enum {
 } POWERUPCNT_TASK_ENUM;
 
 
+#ifdef ESP32
+// the count is kept in its own nvs namespace, and not in the emulated eeprom, so that
+// the setup data is not rewritten on each power up.
+
+#include <Preferences.h>
+
+
+extern volatile uint32_t millis32(void);
+
+
+#define POWERUPCNT_NVS_NAMESPACE  "mlrs-powerup" // max 15 chars
+#define POWERUPCNT_NVS_KEY        "cnt"
+
+#define POWERUPCNT_BIND_CNT       4 // number of rapid power ups to enter bind mode
+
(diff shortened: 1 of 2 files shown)


olliw42:
so, this isn't using EPPROM anymore, but a 'option byte' which is persistent over power ups? (I think EPS has such a thing)

jlpoltrack:
> so, this isn't using EPPROM anymore

Yes, uses Preferences which is also used by the Wireless Bridge.

olliw42:
> > so, this isn't using EPPROM anymore
> 
> Yes, uses Preferences which is also used by the Wireless Bridge.

hm, I'm not sure I understand what that means. I would have thought that Preferences is also kind of a flash-based simulated EPPROM, and thus prone to wear out. I believe to have read that ESP does have a persistent option byte section, which thsu should be much less prone to wear out. What is ELRS actually using?

jlpoltrack:
> I would have thought that Preferences is also kind of a flash-based simulated EPPROM, and thus prone to wear out.

The libraries handle wear-leveling by appending updates and then only doing an erase when space runs out.

> ELRS actually using

ELRS uses the EEPROM library for their powerup counter as best as I can tell:

https://github.com/ExpressLRS/ExpressLRS/blob/9ff3fa4cdb184c92eea95e9db5cfe231da3714da/src/lib/CONFIG/config.cpp#L1193-L1213

> believe to have read that ESP does have a persistent option byte section

I see that there is RTC Memory - but this won't survive a power cycle.  Do you have some Docs for this?

olliw42:
> I see that there is RTC Memory - but this won't survive a power cycle. Do you have some Docs for this?

no, that's a long term memory, it's just when I did that first I plugged into may brain 'oh, ESP can do that better' ... but it's years back ...

hm, I guess I should leave it to you ESP guys then to decide how to do that best
sorry if I'm creating confusion

brad112358:
This looks good to me.

olliw42:
MANY THX!

